PR #492, opened 2026-04-13
fix: align static external_pads with computed defaults

srperens:
## Summary
- **MPEG-TS/SRT Input**: static pads showed only video, but default config includes 1 audio track — audio port appeared after save
- **MPEG-TS/SRT Output**: same issue, audio input port appeared after save
- **WHEP Output**: static pads showed audio+video, but default mode is video-only — audio port disappeared after save
- **Vision Mixer**: static pads showed MAX inputs (10 video + 4 DSK), but defaults are 4 video + 0 DSK — ports collapsed after save

Static `external_pads` in each block definition now match what `get_external_pads()` produces with default property values, so blocks look correct immediately when dragged into the editor.

## Test plan
- [ ] Drag each of the 4 blocks into the editor, verify ports match expectations
- [ ] Save the flow, verify no ports appear or disappear
- [ ] Change properties (e.g. add audio tracks), save, verify ports update correctly
- [ ] `cargo test --test openapi_test` passes

### DIFF
--- a/backend/src/blocks/builder.rs
+++ b/backend/src/blocks/builder.rs
@@ -60,9 +60,9 @@ pub enum WhepStreamMode {
     /// Audio only
     Audio,
     /// Video only
-    #[default]
     Video,
     /// Both audio and video
+    #[default]
     AudioVideo,
 }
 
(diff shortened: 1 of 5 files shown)


srperens:
wait a bit with this. it changes default behavior for whep output ports; it adds a default audio port, which is more correct, but it's a change without proper migration for now.

PR #603, opened 2026-05-29
Vision mixer UX: pixel zone editor, snap feedback, honest bus tally

srperens:
## Summary

Four commits sharpening the vision mixer control page and multiview overlay:

- **Pixel zone editor** (61cefd3): X/Y/W/H inputs in PGM pixels alongside the visual canvas. Snap-to-grid toggle (quarters + rule-of-thirds), optional grid overlay, and an active-snap glow so the operator sees which guide line a drag, resize, or typed value locks onto.
- **Multiview via-PiP rings** (753de45): input thumbnails get a dark-red / dark-green status ring when the input is visible indirectly via the PiP on PGM / PVW. Concentric stacking when multiple states apply (PGM-direct always wins visually).
- **Honest bus UI** (213d032): PGM input + PiP buttons are now `<div>` indicators instead of `<button>` — they were never clickable (PGM only changes via Take/Cut/Auto), but their button shape advertised affordance they didn't have. PVW stays a real button. Inputs reachable via the bus PiP get the same dark-red / dark-green tally on the bus button as the multiview ring.
- **Follow-up fixes** (36845e7): drops misleading "on PGM" tooltips from every PGM indicator (color already conveys the state) and reorders the via-PiP / direct CSS blocks so the direct tally wins on equal specificity.

Only one backend change: `vision_mixer_page.rs` now exposes `pgm_w` / `pgm_h` in the page config so the pixel inputs can round-trip through the same dimensions the backend's `NormRect::to_pixels` uses. All other rect math stays in the page. The thumbnail ring work in 753de45 lives in `overlay.rs` and tracks a new dirty-hash over PiP composition so zone edits trigger re-render.

Data model and wire format are unchanged.

## Test plan

- [ ] Open vision mixer page on a flow with ≥1 PiP. Verify version string reads `v2.16-bus-cleanup`.
- [ ] Open a PiP layout editor. Confirm X/Y/W/H inputs appear under the cap field and reflect the current zone rect in PGM pixels.
- [ ] Toggle Snap off, drag a zone — verify free-drag without anchor pull.
- [ ] Toggle Grid on, drag a zone toward 1/3 / 1/2 / 2/3 — verify guide line glows cyan when snapped.
- [ ] Type `640` into the X field on a 1920-wide PGM — verify it snaps to 1/3 with a brief cyan flash.
- [ ] Put a PiP on PGM whose bg = Input 2 and one zone source = Input 5. Verify PGM bus indicators for Input 2 and Input 5 are dark red, and the PVW bus stays unaffected.
- [ ] Take that PiP to PVW instead — verify the dark-green tally appears on the PVW bus for the same inputs and clears from PGM.
- [ ] Hover a non-active PGM indicator — tooltip should read just `Input N`, not `Input N — on PGM`.
- [ ] Confirm multiview thumbnails show concentric rings matching bus state (dark red / dark green) when the input is in a bus PiP.

### DIFF
--- a/backend/static/vision-mixer.html
+++ b/backend/static/vision-mixer.html
@@ -953,7 +953,7 @@
             <span class="conn-dot" id="wsDot" title="WebSocket control">WS</span>
         </div>
         <span class="status-text" id="statusText">Connecting...</span>
-        <span class="shortcut-hint">v2.16-bus-cleanup | 1-9,0 = PVW | Space = AUTO | X = CUT | F = FTB | M = Hide</span>
+        <span class="shortcut-hint">v2.17-bus-cleanup | 1-9,0 = PVW | Space = AUTO | X = CUT | F = FTB | M = Hide</span>
     </div>
 
     <!-- Server-injected configuration (safe JSON encoding) -->
@@ -1251,7 +1251,13 @@
             if (!pipCfg) return set;
             if (pipCfg.bg != null) set.add(pipCfg.bg);
             for (const zone of pipCfg.zones) {
-                for (const src of zone.sources) set.add(src);
+                // Mirror Rust `Zone::effective_sources`: cap to `capacity`,
+                // keeping the newest entries, so the bus tally and the
+                // multiview ring agree even if sources outgrow capacity.
+                const srcs = (zone.capacity != null && zone.capacity < zone.sources.length)
+                    ? zone.sources.slice(zone.sources.length - zone.capacity)
+                    : zone.sources;
+                for (const src of srcs) set.add(src);
             }
             return set;
         }
@@ -1905,10 +1911,6 @@
             }
             return { value: bestPos, anchor: bestAnchor };
         }
-
-        // Thin wrappers preserved for any call sites that only need the value.
-        function snapEdge(v) { return snapEdgeDetect(v).value; }
-        function snapMoveAxis(pos, size) { return snapMoveAxisDetect(pos, size).value; }
 
         // Highlight the grid line for a given (axis, anchor). axis is 'x' for
         // a vertical line, 'y' for a horizontal one. Looks up the DOM element
--- a/backend/static/vision-mixer.html
+++ b/backend/static/vision-mixer.html
@@ -125,6 +125,25 @@
             color: #999;
         }
 
+        /* "Via PiP" tally: an input is visible on this bus indirectly because
+           the PiP currently on the bus uses it as bg or as a zone source.
+           Dimmer hue than the direct active tally so the eye still ranks
+           "direct > via-pip" at a glance. Matches the multiview thumbnail
+           rings in overlay.rs (PGM_VIA_PIP / PVW_VIA_PIP).
+           Declared BEFORE the direct .pgm-active / .pvw-active blocks so the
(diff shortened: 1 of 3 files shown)


backend/static/vision-mixer.html
```diff
@@ -1554,34 +1866,70 @@
...
-                if (d3 < bestDelta) { bestDelta = d3; bestPos = a - size / 2; }
+                if (d3 < bestDelta) { bestDelta = d3; bestPos = a - size / 2; bestAnchor = a; }
             }
-            return bestPos;
+            return { value: bestPos, anchor: bestAnchor };
+        }
+
+        // Thin wrappers preserved for any call sites that only need the value.
+        function snapEdge(v) { return snapEdgeDetect(v).value; }
+        function snapMoveAxis(pos, size) { return snapMoveAxisDetect(pos, size).value; }
```
srperens: **Point 1 — dead code.** These wrappers have no remaining callers in this PR: every `snapEdge` / `snapMoveAxis` call site was converted to the `*Detect` variant (drag/resize handlers and `setZoneRectPx`). Safe to remove.

backend/static/vision-mixer.html
```diff
@@ -1115,12 +1243,37 @@
...
+        // Inputs that participate in a PiP composition (bg + every zone source).
+        // Mirrors `collect_pip_inputs` in overlay.rs so the bus buttons and the
+        // multiview status rings agree on what "visible via this PiP" means.
+        function collectPipInputs(pipCfg) {
+            const set = new Set();
+            if (!pipCfg) return set;
+            if (pipCfg.bg != null) set.add(pipCfg.bg);
+            for (const zone of pipCfg.zones) {
+                for (const src of zone.sources) set.add(src);
+            }
```
srperens, line 1261: **Point 2 — parity with backend `collect_pip_inputs`.** Rust uses `zone.effective_sources()`, which caps to `capacity` (keeping the newest entries). This JS iterates all of `zone.sources`. They agree today only because the push path trims to capacity client-side — but if state ever carried `sources.length > capacity`, the multiview ring (backend) and the bus tally (JS) would show different input sets. Mirror `effective_sources()` here.

backend/static/vision-mixer.html
srperens: Done in e783ab0.

Verified by loading the shipped page in a real browser (Playwright + Chromium): the `snapEdge` / `snapMoveAxis` wrappers are gone (`typeof === "undefined"`), the `*Detect` variants remain functions, and snapping still works (0.34 → 1/3, 0.51 → 1/2, a far value like 0.87 does not snap, move detents to center). No `ReferenceError` on page load.

backend/static/vision-mixer.html
srperens: Done in e783ab0.

Verified in a real browser against the shipped page: with a PiP on PGM whose zone is `capacity=2, sources=[1,2,3,4]`, the via-PiP tally now resolves to `{bg, newest 2}` = inputs `[0,3,4]` and excludes the evicted `[1,2]`. Matches Rust `Zone::effective_sources` across capacity `null` / `0` / `≥len` / `<len`. Pre-fix, the same setup tallied all sources (`[0,1,2,3,4]`), confirming the fix changes behavior.

PR #663, opened 2026-08-10
fix(whip,whep): expose do-retransmission as a block property

wagenet:
## Description

RTX was not configurable on either WebRTC block: `whipserversrc` (WHIP Input) relied on the element default and `whepserversink` (WHEP Output) hardcoded `do-retransmission` to `true`. This PR exposes `do_retransmission` as a block property on both, defaulting to `true` so existing behaviour is unchanged.

**On WHIP Input** this is the receive leg. RTX feeds retransmitted and reordered packets into the `rtph264depay` that Strom's `decodebin` autoplugs downstream, and that depayloader is where the process aborts on the hdrext assertion in `gst_rtp_base_depayload_handle_buffer` (#685). The property gives operators a switch to try when they hit it.

To be clear about what this does and does not claim: **turning RTX off has not been observed to stop the assertion.** The crash is intermittent and a clean baseline run stopped reproducing it, so the causal link is unproven. The default therefore stays `true`, and this ships as an escape hatch, not a verified mitigation. Defaulting it off would impose a certain quality cost (every lost packet becomes a full keyframe request instead of a cheap resend) for an unconfirmed benefit.

**On WHEP Output** the property cannot affect that crash at all: `whepserversink` is send-only and instantiates no depayloader. It is exposed there purely as a bandwidth/quality tunable, and the PR should not be read as claiming otherwise.

## Related Issue

#685 — Process aborts on `gst_rtp_base_depayload` hdrext assertion when video enters WHIP Input.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tests were added that cover the plumbing, not just the parser:

- `do_retransmission_reaches_whepserversink` — builds the block and reads `do-retransmission` back off the `whepserversink` element itself.
- `do_retransmission_reaches_whip_endpoint_config` — reads the value back off the `WhipEndpointConfig` handed to the session manager.

Both were confirmed to **fail** when `do-retransmission` is hardcoded back to a literal on either block. Under that same revert all six of the original parser tests still passed, so those six are unit coverage of the parse helper rather than regression guards, and are retained on that basis. Neither new test needs a skip guard: `gst-plugin-webrtc` is a Cargo dependency registered via `plugin_register_static()`, so both elements exist in CI without any additional system package.

One hop is deliberately not covered: `WhipEndpointConfig` -> `set_property` inside `create_whipserversrc_for_session`. Exercising it requires calling a function that sets a session pipeline to PLAYING and blocks up to five seconds on the state change, which is unsuitable for a unit test. The WHIP test stops at the config boundary.

**What was actually run:** `cargo test --lib do_retransmission` (8 passed), `cargo fmt --check` (clean), `cargo clippy --workspace --all-targets` (clean, zero lints). **Not run:** the full `cargo test --workspace` suite. The change is additive and workspace-wide clippy is clean, but no broader regression check than the above is being claimed.

## Checklist

- [x] I have run `cargo clippy --workspace --all-targets` — clean, zero lints (the previously noted `frontend/src/themes.rs` failure no longer reproduces)
- [ ] I have run `cargo test --workspace` — **not run**; see Testing above for what was
- [ ] I have updated documentation as needed (not applicable — internal implementation detail, no navigational docs affected)

### DIFF
--- a/backend/src/blocks/builtin/whep.rs
+++ b/backend/src/blocks/builtin/whep.rs
@@ -2359,7 +2359,7 @@ fn whep_output_definition() -> BlockDefinition {
             ExposedProperty {
                 name: "do_retransmission".to_string(),
                 label: "Retransmission (RTX)".to_string(),
-                description: "Resend lost packets to viewers on request (NACK-based). Disable only to cap the outgoing bitrate — without it, packet loss forces a full keyframe request instead of a cheap resend.".to_string(),
+                description: "Resend lost packets to viewers on request (NACK-based). Without it, packet loss forces a full keyframe request instead of a cheap resend.".to_string(),
                 property_type: PropertyType::Bool,
                 default_value: Some(PropertyValue::Bool(true)),
                 mapping: PropertyMapping {
--- a/backend/src/blocks/builtin/whep.rs
+++ b/backend/src/blocks/builtin/whep.rs
@@ -167,6 +167,19 @@ fn explicit_track_count(properties: &HashMap<String, PropertyValue>, name: &str)
     })
 }
 
+/// Parse do_retransmission from properties (default: true).
+///
+/// `whepserversink` is send-only, so this is a bandwidth/quality tunable only.
+fn parse_do_retransmission(properties: &HashMap<String, PropertyValue>) -> bool {
+    properties
+        .get("do_retransmission")
+        .and_then(|v| match v {
+            PropertyValue::Bool(b) => Some(*b),
+            _ => None,
+        })
+        .unwrap_or(true)
+}
+
 /// Migrate a legacy `mode` property on a WHEP Output block to explicit
 /// `num_audio_tracks` / `num_video_tracks` counts, and drop `mode`.
 ///
@@ -951,6 +964,8 @@ fn build_whepserversink(
         num_audio_tracks, num_video_tracks
     );
 
+    let do_retransmission = parse_do_retransmission(properties);
+
     // Timestamp offset in milliseconds. A negative value shifts playout earlier,
     // reducing end-to-end latency for this output while maintaining A/V sync.
     // Applied as ts-offset on clocksync and appsink inside whepserversink.
@@ -1021,7 +1036,7 @@ fn build_whepserversink(
         whepserversink.set_property("turn-servers", turn_servers);
     }
 
-    // Disable FEC but keep RTX (retransmission) enabled.
+    // Disable FEC; RTX (retransmission) is configurable, default on.
     // - FEC adds proactive redundancy packets on every stream (~50% constant
     //   overhead, near-double bandwidth for pre-encoded high-bitrate video),
     //   so it stays off.
@@ -1030,7 +1045,7 @@ fn build_whepserversink(
     //   escalates to PLI -> forced keyframe, which is far more expensive and
     //   leaves the picture broken until the keyframe arrives.
     whepserversink.set_property("do-fec", false);
-    whepserversink.set_property("do-retransmission", true);
+    whepserversink.set_property("do-retransmission", do_retransmission);
 
     // Access the signaller child and set its properties
     // Bind to localhost only - axum will proxy external requests
@@ -2341,6 +2356,20 @@ fn whep_output_definition() -> BlockDefinition {
                 live: false,
                 persist: None,
             },
+            ExposedProperty {
+                name: "do_retransmission".to_string(),
+                label: "Retransmission (RTX)".to_string(),
+                description: "Resend lost packets to viewers on request (NACK-based). Disable only to cap the outgoing bitrate — without it, packet loss forces a full keyframe request instead of a cheap resend.".to_string(),
+                property_type: PropertyType::Bool,
+                default_value: Some(PropertyValue::Bool(true)),
+                mapping: PropertyMapping {
+                    element_id: "_block".to_string(),
+                    property_name: "do_retransmission".to_string(),
+                    transform: None,
+                },
+                live: false,
+                persist: None,
+            },
         ],
         // Note: external_pads here are the static defaults (1 video + 1 audio).
         // The actual pads are determined dynamically by WHEPOutputBuilder::get_external_pads()
@@ -2378,6 +2407,13 @@ fn whep_output_definition() -> BlockDefinition {
 mod tests {
     use super::*;
 
+    /// `whepserversink` comes from the `gst-plugin-webrtc` crate, which is only
+    /// registered by the binary. Tests must register it themselves.
+    fn init_gst() {
+        let _ = gst::init();
+        let _ = gstrswebrtc::plugin_register_static();
+    }
+
     /// Build a property map. `legacy_mode` populates the old "mode" enum
     /// (`Some("audio")` etc.) to exercise the migration path; pass `None` for
     /// new flows. Count values pass through `num_audio_tracks` /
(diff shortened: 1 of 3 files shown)


srperens (dismissed):
**Retracted — this review was misattributed.**

The original text of this review described the changes in #665 (`jitterbuffer_latency_ms` on WHIP Input), not this PR. It was posted here by an automated reviewer that crossed two PRs in the same run, so it never verified this PR's diff at all. The text has been replaced to stop it being read as a review of `do-retransmission`.

This PR has not yet had a valid review. It will be re-reviewed under a stricter protocol that requires an inline citation — code path, commit, doc section or test — for every claim, and that verifies the PR's own description claim by claim. Nothing here should be taken as a signal about this change's correctness either way.

The original text is retained outside GitHub for the record.

srperens:
## Review: the WHEP half of this change cannot work around the crash it cites

Verified against the PR diff (commit range of `wagenet/pr-a-do-retransmission`), `origin/main` at `4a36815`, and `gst-inspect-1.0` on GStreamer 1.28.6. The WHIP half looks correct; the WHEP half does not.

### 1. `whepserversink` never depayloads, so `do-retransmission` there cannot prevent a depayloader assertion

The stated motivation is an assertion in **`gst_rtp_base_depayload_handle_buffer`** (`priv->hdrext_buffers` non-empty while `hdrext_delayed` is set) — i.e. inside a *depayloader*. But `whepserversink` has no RTP input and contains no depayloader:

```
$ gst-inspect-1.0 whepserversink
SINK template: 'video_%u'
  Capabilities: video/x-raw, video/x-raw(memory:CUDAMemory|GLMemory|NVMM|D3D11Memory|VAMemory),
                video/x-vp8, video/x-h264, video/x-vp9, video/x-h265, video/x-av1
SINK template: 'audio_%u'
  Capabilities: audio/x-raw, audio/x-opus
```

It takes raw or encoded media and *payloads* it. The property semantics confirm the direction:

```
$ gst-inspect-1.0 whipserversrc
  do-retransmission : Send retransmission events upstream when a packet is late

$ gst-inspect-1.0 whepserversink
  do-retransmission : Whether the element should offer to honor retransmission requests
```

- On `whipserversrc` it is the **receive** side: the jitterbuffer requests RTX, and the recovered packets pass through the depayloader that webrtcsrc instantiates internally. That is where the cited assertion can bite, so the WHIP property is a plausible workaround.
- On `whepserversink` it is the **send** side: honor NACKs from the viewer's browser. No depayloader exists on that path inside the Strom process. Neither `whep.rs` nor `whip.rs` adds one (`grep -n depay backend/src/blocks/builtin/whep.rs` → no matches).

So `do_retransmission=false` on WHEP Output cannot stop that crash.

### 2. The incorrect causal claim ships in operator-facing text

`whep.rs`, the new `ExposedProperty`:

> "Disable only for diagnostics — a known GStreamer bug can crash the process when RTX is combined with RTP header-extension aggregation; without it, packet loss forces a full keyframe request instead of a cheap resend."

The same claim is added to the code comment replacing `whep.rs:1033` (`whepserversink.set_property("do-retransmission", true);` on main). This tells an operator to disable RTX for every viewer — making each packet loss escalate to PLI plus a full keyframe — in pursuit of a crash the switch does not affect.

### 3. The new tests do not guard the change

All six tests exercise only `parse_do_retransmission` against itself:

```rust
assert!(parse_do_retransmission(&raw_props(&[])));
```

None asserts that the parsed value reaches `set_property("do-retransmission", ...)` on the element, or that it survives the `WhipEndpointConfig` -> `create_whipserversrc_for_session` hop (the only non-trivial plumbing in the diff). Revert the fix and the helper and its tests disappear together, so nothing fails. Per the repo's test rules that is a demonstration, not a regression guard.

### 4. The workaround itself is not stated as verified

The description says the crash "was reproduced twice independently on two different code paths hitting the identical assertion", but does not state that setting `do-retransmission=false` was observed to stop it. Worth confirming before the property is justified as a workaround at all.

### Suggested resolution

- **Keep** `do_retransmission` on **WHIP Input**: right side of the crash, correct semantics, default `true` matches the upstream default.
- **Drop it from WHEP Output**, or keep it purely as a tuning knob with all crash-workaround language removed from both the property description and the code comment.
- Add one test that shows the value actually propagates to the element, so the change is guarded.

For the record, the duplicated `parse_do_retransmission` helper in both files is *not* an objection — the inline `PropertyValue::Bool(b) => Some(*b)` pattern is the established idiom in this codebase (`aes67.rs`, `mpegtssrt_input.rs`, `compositor.rs`, `efpsrt_input.rs`, and others).

srperens:
Tested this live on macOS today. The crash is now tracked in #685 with the log that locates it: the asserting depayloader is `rtph264depay0` in the flow's **main** pipeline, autoplugged by Strom's own `decodebin_video_0`. `whipserversrc` hands out `application/x-rtp` and Strom does the depayloading.

That confirms the review above from measurement rather than reasoning: the WHIP half of this PR is on the right side of the bug, the WHEP half cannot affect it.

One thing worth trying before trading RTX away: the header extension in play is TWCC, and Strom turns it up itself (`Set twcc-feedback-interval=200ms on rtpsession1`, `whip.rs:626-635`). Dropping that is a cheaper mitigation if it is sufficient. Both that and `do-retransmission=false` were scaffolded today but never exercised — the assertion is intermittent and a clean baseline run stopped reproducing it. See #685 for the reproduction status.

wagenet:
@srperens I reworked this a bit. I think it is probably mergeable now, but since I couldn't actually reproduce the issue I had originally been addressing I also understand if you don't want to merge it.

srperens (dismissed):
Reviewing PR #663: fix(whip,whep): expose do-retransmission as a block property
Verified against 3493202 (head) vs main 81cdc83.

No prior valid review exists for this PR — the only earlier review (id 4991338543) was retracted as misattributed and never assessed this diff. This is a first-pass v3 review.

**Summary:** Exposes `do_retransmission` as a configurable block property on WHIP Input (`whipserversrc`, previously left at element default) and WHEP Output (`whepserversink`, previously hardcoded `true`). Both default to `true`, preserving existing behavior for flows that don't set it.

**Claim verification**
|---|---|---|
| `whipserversrc` actually has a `do-retransmission` GObject property reaching the RTP transceiver | CONFIRMED (external, version-pinned) | `Cargo.lock` pins `gst-plugin-webrtc` `0.15.2`; fetched `webrtcsrc/imp.rs` at tag `0.15.2` from `GStreamer/gst-plugins-rs`: `BaseWebRTCSrc` registers `ParamSpecBoolean::builder("do-retransmission")` (line 152), applied via `transceiver.set_property("do_nack", ...)`. `webrtcsrc/mod.rs:98,54`: `whipserversrc`/`WhipServerSrc` `@extends BaseWebRTCSrc` — inherited, not invented |
| New tests actually exercise the plumbing and would fail on revert | CONFIRMED | Both `do_retransmission_reaches_whip_endpoint_config` (`whip.rs`) and `do_retransmission_reaches_whepserversink` (`whep.rs`) parametrize over `[true, false]` and assert the read-back value equals the input — a literal-hardcoded property fails one branch regardless of which value is hardcoded |
| New tests ran for real in this run's CI (not skipped) | CONFIRMED | CI log, run 32863323341: all 8 `do_retransmission_*` tests listed individually as `... ok`, matching the PR body's "8 passed" |
| WHEP's RTX property is unrelated to the #685 depayloader crash (send-only, no depayloader) | CONFIRMED | `whepserversink` element itself only payloads outbound RTP; the `decodebin` present elsewhere in `whep.rs` (`setup_audio_decode_chain`, ~line 1968) mixes the block's own internal audio input tracks pre-encode and is unrelated to `#685`, which is scoped to RTP ingest via WHIP Input |
| Turning off RTX has not been confirmed to fix #685 | CONFIRMED, consistent | Matches my own standing #685 investigation (2026-08-25T07:28:45Z comment): "mitigation on hold," causal link unproven |

**Diagnosis:** correctly scoped as an operator switch, not a claimed fix for #685. BOUNDED: existing persisted flows with no `do_retransmission` key parse to `true` via `unwrap_or(true)`, so no silent behavior change on upgrade.

**Blast radius:** LOCAL/SHARED. Follows the exact established `jitterbuffer_latency_ms` pattern (parse fn → `WhipEndpointConfig` field → applied in `create_whipserversrc_for_session`) already used in this file, and `ExposedProperty`/`PropertyType::Bool` are pre-existing `strom-types` types — no new API surface, consistent with `API Contract Check` passing.

**Tests & CI:** `Check (Linux)`, `Build (Linux ARM64/x86_64)`, `API Contract Check`, `Check & Build (WASM)`, `sccache preflight` all pass. Windows/macOS skip by design (`workflow_dispatch`-gated), no platform code in this diff. One gap, disclosed by the author: the `WhipEndpointConfig → whipserversrc.set_property` hop in `create_whipserversrc_for_session` has no test (needs a live PLAYING session). My upstream-source check above confirms the property is real, which covers "does this do something," not a regression guard for that line.

No CLAUDE.md violations found.

**Confidence: HIGH** — the one claim I could not settle from this repo alone (property existence) is now CONFIRMED against the exact pinned upstream version, and every other claim is either CONFIRMED in-repo or an honestly-labeled unconfirmed causal claim the author does not overstate.

Approve.

srperens (approved):
Approving. Verified locally on macOS (GStreamer 1.28.6, `gst-plugin-webrtc` 0.15.2) rather than on the PR body alone.

**Guard strength confirmed.** Hardcoding `do-retransmission` back to a literal on either block fails exactly the two plumbing tests, while the six parser tests still pass — matching the claim in the body. Those six are unit coverage of the parse helper, not regression guards, which the body already says.

**The uncovered hop is safe.** The body flags that `WhipEndpointConfig` -> `set_property` inside `create_whipserversrc_for_session` is not exercised. I probed the crate's element directly: `whipserversrc` exposes `do-retransmission` as `gboolean` and set/get round-trips, so there is no panic risk on that path. Checked against the crate, not the system plugin, since those can diverge.

**No default behaviour change on any existing path.** The element default is `true` on both `whipserversrc` and `whepserversink`, so WHEP (previously hardcoded `true`) and WHIP (previously inheriting the element default) both keep their current behaviour, and existing saved flows without the key parse to `true`. Worth noting WHIP moves from inheriting to pinning the value — same result today, and more predictable if the element default ever changes.

**`live: false` is correct**, not an oversight: `gst-inspect` reports the property as changeable only in NULL or READY, and it is applied before the session pipeline goes to PLAYING.

**Closed the testing gap in the body.** `cargo test --workspace` was listed as not run; it is green here (607 tests, 0 failures, one pre-existing version-gated skip unrelated to this change). `cargo fmt --check` and `cargo clippy --workspace --all-targets` are both clean.

**No frontend change needed.** `PropertyValue::Bool` already renders as a checkbox, and `default_value` is consulted before the `Bool(false)` fallback for unset properties, so the control shows RTX enabled rather than contradicting the backend.

One change pushed to the branch (69392f6): the property descriptions now state only what the property does. The WHIP description previously told operators that disabling RTX was an unconfirmed workaround for the #685 depayloader crash, which invites paying a real quality cost chasing a fix the body itself says has never been observed to work; that context belongs in the PR and the issue, where it already is. Also dropped "Disable only to cap the outgoing bitrate" from WHEP by the same rule.

PR #665, opened 2026-08-10
fix(whip): add jitterbuffer_latency_ms property to prevent stall on session start

wagenet:
## Summary
- WHIP Input's per-session `webrtcbin` left jitterbuffer latency unset, defaulting to 200ms. Combined with `drop-on-latency=true` on the internal `rtpbin`, this was too tight for the initial video keyframe's packet burst on a freshly-created pipeline, causing the burst to be dropped locally before ever reaching `decodebin` — video silently never starts, even though packets arrived fine over the network.
- Adds a configurable `jitterbuffer_latency_ms` property (default 400ms) applied to the webrtcbin's `latency` property at session creation, giving the initial keyframe burst enough room to survive.

## Test plan
- [x] `cargo fmt --all`
- [x] `cargo build`
- [x] `cargo test --workspace` (includes 3 new unit tests for `parse_jitterbuffer_latency_ms`: default, explicit value, negative clamp)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` on the changed files (no new warnings introduced)

### DIFF
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -139,6 +139,17 @@ impl BlockBuilder for WHIPInputBuilder {
 // WHIP Input (whipserversrc - hosts WHIP server)
 // ============================================================================
 
+/// Parse jitterbuffer_latency_ms from properties (default: 400, negative clamps to 0).
+fn parse_jitterbuffer_latency_ms(properties: &HashMap<String, PropertyValue>) -> u32 {
+    properties
+        .get("jitterbuffer_latency_ms")
+        .and_then(|v| match v {
+            PropertyValue::Int(i) => Some((*i).max(0) as u32),
+            _ => None,
+        })
+        .unwrap_or(400)
+}
+
 /// Build WHIP Input per-slot output chains.
 ///
 /// At build time, per-slot chains are created in the main pipeline:
@@ -180,6 +191,14 @@ fn build_whipserversrc(
             _ => None,
         })
         .unwrap_or(true);
+
+    // Jitterbuffer latency: how long to buffer before dropping/releasing packets.
+    // Left unset, webrtcbin defaults to 200ms, which combined with
+    // drop-on-latency=true (below) can be too tight for an initial video
+    // keyframe's packet burst on a freshly-created per-session pipeline,
+    // causing the whole video stream to stall (never reaching decodebin)
+    // even though the packets arrived fine over the network.
+    let jitterbuffer_latency_ms = parse_jitterbuffer_latency_ms(properties);
 
     let max_video_bitrate_kbps = properties
         .get("max_video_bitrate")
@@ -436,6 +455,7 @@ fn build_whipserversrc(
             ice_transport_policy: ctx.ice_transport_policy().to_string(),
             pipeline_weak: gst::glib::WeakRef::new(),
             decode,
+            jitterbuffer_latency_ms,
             dynamic_webrtcbin_store: ctx.dynamic_webrtcbin_store(),
             max_video_bitrate_kbps,
             max_sessions,
@@ -531,6 +551,7 @@ pub fn create_whipserversrc_for_session(
     let dynamic_webrtcbin_store = config.dynamic_webrtcbin_store.clone();
     let block_id_for_callback = config.instance_id.clone();
     let ice_transport_policy = config.ice_transport_policy.clone();
+    let jitterbuffer_latency_ms = config.jitterbuffer_latency_ms;
     // Flag to ensure only one cleanup request per session (shared across ICE callback,
     // inactivity watchdog, etc.)
     let cleanup_sent = Arc::new(AtomicBool::new(false));
@@ -559,6 +580,14 @@ pub fn create_whipserversrc_for_session(
                     info!(
                         "WHIP Input: Set ice-transport-policy={} on webrtcbin {}",
                         ice_transport_policy, element_name
+                    );
+                }
+
+                if element.has_property("latency") {
+                    element.set_property("latency", jitterbuffer_latency_ms);
+                    info!(
+                        "WHIP Input: Set jitterbuffer latency={}ms on webrtcbin {}",
+                        jitterbuffer_latency_ms, element_name
                     );
                 }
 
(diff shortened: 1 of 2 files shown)


srperens (dismissed):
**Retracted — this review was misattributed, and its approval is dismissed.**

The original text of this review described the changes in #663 (making `do-retransmission` configurable), not this PR. It was posted here by an automated reviewer that crossed two PRs in the same run, so it never verified this PR's diff — yet it carried an approval. That approval has been dismissed.

This PR has not yet had a valid review. It will be re-reviewed under a stricter protocol that requires an inline citation — code path, commit, doc section or test — for every claim, and that verifies the PR's own description claim by claim. Nothing here should be taken as a signal about this change's correctness either way.

The original text is retained outside GitHub for the record.

srperens (commented):
# Reviewing PR #665

head=`b36cc4b4` (b36cc4b416d8e4a11bdfe4b39d1225fd10ffca58) vs main=`317e80b1` (317e80b1993e4cab38d52dfed418bdf37f569f51)

## Summary

Adds a new `jitterbuffer_latency_ms` exposed property (default 400ms) to the WHIP input block, and sets it on the internal `webrtcbin`'s `rtpjitterbuffer` via the existing `deep-element-added` signal handler. Intent per PR description: prevent a stall on session start caused by GStreamer's `webrtcbin` default jitterbuffer `latency` (200ms) being too tight for some WHIP publishers.

## Claim verification

|---|---|---|---|
| 1 | New code sets `latency` property on `webrtcbin` inside the pre-existing `deep-element-added` handler, alongside the existing `ice-transport-policy` block | CONFIRMED | CODE: `git show pr665:backend/src/blocks/builtin/whip.rs` lines 555-605 — new `if element.has_property("latency") { element.set_property("latency", jitterbuffer_latency_ms); ... }` block sits directly after the existing `ice-transport-policy` block inside `bin.connect("deep-element-added", false, move \|values\| {...})` |
| 2 | `rtpbin`/jitterbuffer already has `drop-on-latency=true` set in this same handler, so the new `latency` value interacts with existing drop-on-latency behavior rather than introducing a new mechanism | CONFIRMED | CODE: same closure in `whip.rs` on `main` (pre-existing, unmodified by this diff) sets `drop-on-latency=true` on `rtpbin` elements |
| 3 | Default value 400ms is a deliberate increase over GStreamer's `webrtcbin.latency` default | CONFIRMED (default) / EXTERNAL (upstream default value) | CODE: `parse_jitterbuffer_latency_ms` defaults to `400` when property absent. EXTERNAL: WebSearch confirms `GstWebRTCBin.latency` property default is 200ms (RidgeRun GStreamer WebRTC wiki) — corroborates the stated rationale, not independently re-derivable in this sandbox from GStreamer source |
| 4 | Negative/invalid input is clamped rather than causing a panic or invalid property set | CONFIRMED | CODE: `(*i).max(0) as u32` in `parse_jitterbuffer_latency_ms`; TEST: `jitterbuffer_latency_ms_clamps_negative_to_zero` exercises this path |
| 5 | New tests cover default, explicit-value, and negative-clamp cases | CONFIRMED | TEST: `#[cfg(test)] mod tests` in `whip.rs` — `jitterbuffer_latency_ms_defaults_to_400`, `jitterbuffer_latency_ms_respects_explicit_value`, `jitterbuffer_latency_ms_clamps_negative_to_zero`, all present in diff |
| 6 | New `ExposedProperty` follows the existing pattern (`live: false`, `Int` type) consistent with other WHIP block properties | CONFIRMED | CODE: new entry in the block's `ExposedProperty` list matches the shape/format of neighboring entries (e.g. the pre-existing latency-adjacent properties in the same definition) |
| 7 | No `WeakRef`/pipeline-reference violation introduced (per CLAUDE.md GStreamer closure rule) | CONFIRMED | CODE: only a `u32` (`Copy`) value is moved into the closure; no `clone()` of `gst::Pipeline`/`Element`/`Bin` added |
| 8 | CI ran and passed on this PR's head commit | CONTRADICTED | CI: `gh api repos/eyevinn/strom/commits/b36cc4b416d8e4a11bdfe4b39d1225fd10ffca58/check-runs` → `total_count: 0`. The associated workflow run `31396753934` has `status: completed`, `conclusion: action_required` — this is GitHub's fork-PR approval gate; **no checks executed at all**, not a pass or a fail |

## Diagnosis

The change is small, self-contained, and well-targeted: it reuses an existing signal-handler hook point, follows the same conditional-property-set pattern already used for `ice-transport-policy` in the same closure, and its stated interaction with the pre-existing `drop-on-latency=true` setting is real and traceable in the current codebase (not asserted by the PR author without basis).

History check (protocol step 2f): `git log` on `backend/src/blocks/builtin/whip.rs` shows two prior related-but-distinct fixes — `1e65fad` ("set drop-on-latency=true on WHIP and AES67 input blocks") and `0a5786a` ("remove unnecessary queue property overrides, use GStreamer defaults"). Neither addresses the jitterbuffer `latency` property itself; they touch adjacent jitterbuffer/queue tuning in the same file. This is not a case of the same bug being patched repeatedly — it reads as incremental hardening of WHIP jitterbuffer/queue behavior over several PRs, which is consistent with a legitimately evolving problem area rather than a symptom of a misdiagnosed root cause.

## Blast radius

LOCAL — confined to the WHIP input block (`backend/src/blocks/builtin/whip.rs`) and its config struct (`backend/src/whip_session_manager.rs`, `+2` lines for the new `jitterbuffer_latency_ms: u32` field). No changes to `strom-types`, no `StromEvent` variant touched, no API endpoint touched. Default value preserves prior effective behavior only if 400ms happens to match whatever the pipeline previously ran at (webrtcbin default is 200ms per external doc, so this is actually a behavior change for all existing WHIP inputs, increasing default latency — expected and presumably desired per the PR's stated goal, but worth flagging explicitly as a default-behavior change, not purely additive).

## Tests & CI

Three new unit tests cover the default/explicit/clamp cases for the pure parsing function (`parse_jitterbuffer_latency_ms`) — these are real, executable, and would fail if the fix were reverted (they test the parsing logic itself, not just presence of a constant).

However: **CI never executed on this commit.** `check-runs` returns 0, and the single associated Actions run (`31396753934`) is stuck at `conclusion: action_required` — GitHub's approval gate for workflow runs triggered by non-collaborator fork PRs. This means the new tests above have not actually been run by CI on this exact head SHA; their correctness is confirmed by direct source reading in this review, not by an executed CI job. This is the same systemic issue observed on other `wagenet/*` PRs in this batch (see run summary).

Unblock command for a maintainer with repo admin access:
```
gh api repos/eyevinn/strom/actions/runs/31396753934/approve -X POST
```

## Risks / edge cases

- Raising default jitterbuffer latency from webrtcbin's 200ms to 400ms is a real behavior change for all existing WHIP sessions (not just new ones setting the property explicitly) — end-to-end latency for WHIP ingest increases by design. This appears intentional per the PR's stated goal (prevent stall), but should be called out in the PR description if not already (not verified either way — the PR description text itself was not re-quoted here beyond the stated intent).
- No upper bound is enforced on `jitterbuffer_latency_ms` (only clamped at 0 on the low end) — an operator could set an extremely large value; this is consistent with how other `Int`-typed WHIP properties are handled in this codebase (no existing upper-bound pattern to deviate from), so not a regression, just worth noting.

## Requested changes

None from a correctness standpoint — the fix is small, well-tested at the unit level, and consistent with existing patterns in the file. Approval is withheld only because CI has not executed on this head commit (see below).

## Confidence: MEDIUM — code and tests are sound by direct reading, but zero CI execution on this head SHA means the covering-tests-must-run gate is not satisfied. Verdict: Comment (not Approve) until CI is unblocked and passes.

PR #666, opened 2026-08-10
fix(whip): proactively retry keyframe (PLI) request on video pad connect

wagenet:
## Summary
- WHIP Input's existing keyframe-recovery mechanism (a DISCONT-triggered PLI request) only installs once a decoder element exists downstream of `decodebin`. If the publisher's very first keyframe is incomplete or lost before that decoder exists, `decodebin` waits forever for a keyframe that never arrives — a chicken-and-egg stall that's otherwise invisible, since packets keep flowing normally, just never a usable one.
- Fixes this by proactively sending an `UpstreamForceKeyUnitEvent` (PLI) on the video pad as soon as it connects, retried 5 times spaced 800ms apart on a dedicated background thread. A single request isn't reliable either — if retransmission is disabled or unavailable, the keyframe response itself can be lost with no recovery — so retrying gives decode a fair chance to succeed.

## Test plan
- [x] `cargo fmt --all`
- [x] `cargo build`
- [x] `cargo test --workspace`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` on the changed file (no new warnings introduced)
- [ ] No unit test is feasible for this change — it's runtime GStreamer pad/thread behavior, and no pipeline-construction test harness exists in this codebase for WHIP/WHEP. Verified via manual `GST_DEBUG`/pad-caps inspection during live diagnosis of a WHIP Input -> WHEP Output flow.

### DIFF
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -831,6 +831,41 @@ pub fn create_whipserversrc_for_session(
                     pad_name, stream_num, slot, media_type
                 );
 
+                if media_type == "video" {
+                    // Proactively request a keyframe as soon as the video pad
+                    // connects, and keep retrying for a few seconds. Without
+                    // this, if the publisher's initial keyframe is
+                    // incomplete/lost before decodebin has a decoder element
+                    // to install the existing DISCONT-triggered recovery
+                    // probe on, decodebin waits forever for a keyframe that
+                    // never arrives — a chicken-and-egg stall that's
+                    // otherwise invisible (packets keep flowing, just never a
+                    // usable one). If retransmission is disabled/unavailable,
+                    // the keyframe response itself can also be lost with no
+                    // recovery, so a single request isn't reliable — retry a
+                    // few times, spaced out, until decode has had a fair
+                    // chance to succeed.
+                    let retry_pad = pad.clone();
+                    let retry_slot = slot;
+                    std::thread::Builder::new()
+                        .name(format!("whip-keyframe-retry-{}", retry_slot))
+                        .spawn(move || {
+                            for attempt in 0..5 {
+                                let fku = gst_video::UpstreamForceKeyUnitEvent::builder()
+                                    .all_headers(true)
+                                    .build();
+                                retry_pad.send_event(fku);
+                                info!(
+                                    "WHIP Input: Requested keyframe (PLI) on video pad for slot {} (attempt {})",
+                                    retry_slot,
+                                    attempt + 1
+                                );
+                                std::thread::sleep(std::time::Duration::from_millis(800));
+                            }
+                        })
+                        .ok();
+                }
+
                 let ts_offset = shared_ts_offset.clone();
                 let main_pipeline_for_ts = main_pipeline_weak.clone();
                 let media_for_log = media_type.to_string();


srperens (dismissed):
**Intent:** Proactively request a keyframe (PLI) as soon as a video pad connects on a WHIP session, retrying every 800 ms for up to 5 attempts. This fixes a chicken-and-egg stall where the publisher's initial keyframe is lost before `decodebin` has a decoder element in place to handle the DISCONT-triggered recovery path — leaving the pipeline receiving packets but never decoding any of them.

**Risk:** Low-medium. The fix spawns a short-lived background thread per WHIP video session. The retry window (5 × 800 ms = 4 s) is bounded, and `UpstreamForceKeyUnitEvent` is a GStreamer standard event with no side effects beyond asking the encoder to produce a keyframe.

---

**What looks good:**

- Using `std::thread::Builder::new().name(...)` for the retry thread is correct — it gives the thread a meaningful name in debugging tools.
- `.ok()` on the spawn result silently ignores thread-creation failures, which is acceptable here (the worst outcome is no retry).
- The `UpstreamForceKeyUnitEvent::builder().all_headers(true)` is correct for requesting a full keyframe with SPS/PPS/VPS headers.
- `retry_pad` is a clone, not a strong element reference, and `pad.clone()` in GStreamer Rust produces a reference-counted clone (not a cycle risk for a short-lived thread).

**Two things to consider:**

1. **Thread lifetime vs session lifetime:** If the WHIP session is torn down during the 4-second retry window, `retry_pad.send_event(fku)` on a detached pad is safe (it returns false, the thread just logs and sleeps). No crash risk, but the log will show PLI attempts on a dead session. A cancellation token (e.g. an `Arc<AtomicBool>`) could suppress the noisy logs, though this is purely cosmetic.

2. **Interaction with PR #663 (jitterbuffer latency):** These two fixes address different aspects of the same "video never starts" symptom. It would be useful to document in either PR description which problem each solves, so future debuggers know which knob to reach for (stall at jitterbuffer vs stall at decodebin keyframe wait). Both fixes are independently correct.

Neither point is blocking.

srperens (dismissed):
Reviewing PR #666: fix(whip): proactively retry keyframe (PLI) request on video pad connect
Verified against ead7b90b vs main 317e80b1.

**Summary**
- On WHIP video pad connect, spawns a short-lived background thread that sends `UpstreamForceKeyUnitEvent` (PLI) 5 times at 800ms intervals (total ~4s).
- Fixes a chicken-and-egg stall where the initial keyframe is lost before `decodebin` has a decoder element to install the existing DISCONT-triggered recovery probe on.

**Claim verification**

|---|---|---|
| `retry_pad` is a pad clone, not a pipeline/element/bin — no reference cycle | CONFIRMED | `backend/src/blocks/builtin/whip.rs:834` `retry_pad = pad.clone()`. `pad` is a `gst::Pad`, not a `gst::Pipeline`/`Element`/`Bin`. Thread closure captures `retry_pad` (Pad) and `retry_slot` (integer). No GStreamer cycle risk. |
| Thread is bounded — retries stop after 5 attempts | CONFIRMED | `whip.rs:838` `for attempt in 0..5 { … std::thread::sleep(800ms); }` — thread exits after 4s regardless. |
| Thread creation failure is silently ignored via `.ok()` | CONFIRMED | `whip.rs:852` `.ok()` — acceptable for a best-effort retry; worst case is no retry, which is the pre-fix behaviour. |
| `UpstreamForceKeyUnitEvent::builder().all_headers(true)` is the correct PLI mechanism | UNVERIFIED (external) | GStreamer upstream force-key-unit event specification; not tested in this repo. The event type is the standard GStreamer PLI/FIR equivalent. |
| Sending an event on a detached pad after session teardown is safe | UNVERIFIED (external) | GStreamer Pad event API behaviour when pad is not in a pipeline. Common expectation is `send_event` returns `false` without crashing, consistent with GStreamer's safe degradation model. |

**Test coverage** — no test claimed. Runtime GStreamer pad/thread behaviour; no pipeline-construction test harness for WHIP. Verified by manual GST_DEBUG inspection per PR description. Acceptable.

**Risks / edge cases**
- If the session is torn down during the 4s retry window, `send_event` on the detached pad produces log noise but no crash. Adding an `Arc<AtomicBool>` cancellation token would suppress the noise; the current behaviour is acceptable given the bounded window.
- Interaction with PR #665 (jitterbuffer latency): both address different aspects of the "video never starts" symptom. These fixes are independent and both correct.

**Requested changes** — none.

**Confidence: HIGH**

srperens (commented):
# Reviewing PR #666

head=`ead7b90b` (ead7b90b19b638e172ec63d4ef91cff9ebe7a365) vs main=`317e80b1` (317e80b1993e4cab38d52dfed418bdf37f569f51)

## Summary

WHIP Input's existing DISCONT-triggered PLI (keyframe request) recovery only installs once a decoder element exists downstream — if the very first keyframe is lost before that decoder exists, `decodebin` stalls forever waiting for a keyframe that never arrives, while packets keep flowing normally (an invisible stall). This PR proactively sends an `UpstreamForceKeyUnitEvent` on the video pad as soon as it connects, retried 5 times at 800ms intervals on a dedicated background thread, to give decode a better chance of recovering even if a single PLI response is lost.

## Claim verification

|---|---|---|---|
| 1 | The retry logic only runs for video pads, not audio | CONFIRMED | CODE: `if media_type == "video" { ... }` gate wraps the entire new block |
| 2 | 5 retries, spaced 800ms apart, sent as `UpstreamForceKeyUnitEvent` with `all_headers(true)` | CONFIRMED | CODE: `for attempt in 0..5 { let fku = gst_video::UpstreamForceKeyUnitEvent::builder().all_headers(true).build(); retry_pad.send_event(fku); ... std::thread::sleep(std::time::Duration::from_millis(800)); }` |
| 3 | Retry runs on a dedicated background thread, not blocking the signal handler | CONFIRMED | CODE: `std::thread::Builder::new().name(format!("whip-keyframe-retry-{}", retry_slot)).spawn(move \|\| { ... }).ok();` — spawned thread, handler returns immediately after spawn |
| 4 | Captures the pad via `.clone()` inside a signal-handler closure — potential CLAUDE.md WeakRef violation? | **NOT a violation, but a real caveat** | CODE: `let retry_pad = pad.clone();` clones a `gst::Pad`, not a `Pipeline`/`Element`/`Bin` — the CLAUDE.md rule is scoped specifically to those three types, precisely because `GstObject` parent→child links are not strong-ref-counted back up the hierarchy (a Pad holding a ref does not itself keep its owning Element/Pipeline artificially alive). This does **not** violate the letter of the rule. However: the spawned thread's `JoinHandle` is discarded (`.ok()`), so there is no mechanism to cancel or join this thread if the flow is stopped mid-retry — worst case, a `gst::Pad` reference (and the background thread itself) can outlive a `stop_flow()` call by up to ~4 seconds (5 × 800ms). Whether `send_event` on a pad whose pipeline has been torn down mid-flight is safe (no-op/false return) or could interact with `stop_flow()`'s "pipeline survived after drop" P0 log check could not be verified in this sandbox (no live GStreamer runtime) |

## Diagnosis

The core mechanism (proactive + retried PLI) is a reasonable, well-motivated fix for the described chicken-and-egg stall, and is scoped correctly to video-only. The one real risk — an untracked background thread holding a `Pad` reference for up to 4 seconds past a possible flow-stop — is not a clear-cut bug (Pad refs don't keep Pipeline/Bin/Element artificially alive per GStreamer's object-parenting model, so this doesn't create the kind of permanent leak CLAUDE.md's pipeline_lifecycle_test guards against), but it is untested interaction between this new thread and `stop_flow()`'s teardown timing that I could not verify either way.

## Blast radius

LOCAL — confined to the `pad-added` handling path (`create_whipserversrc_for_session`) for video pads in `backend/src/blocks/builtin/whip.rs`. Sends extra PLI events during the first ~4 seconds of every WHIP Input video session (not just failure cases) — a minor behavior change (more frequent keyframe requests than before, always) worth the maintainer's awareness even though it's the intended mechanism.

## Tests & CI

No unit test added — PR description states "no pipeline-construction test harness exists in this codebase for WHIP/WHEP," and describes manual verification via `GST_DEBUG`/pad-caps inspection during live diagnosis — consistent reasoning with #667/#668 in this same batch, not independently reproducible here.

CI: `check-runs` for `ead7b90b` shows the same two-job failure (`Check (Linux)`, `Check & Build (WASM)`) with identical `frontend/src/themes.rs` `f32: From<f64>` fallback errors seen across this whole batch (run `31398152616`, `created_at: 2026-08-10`, `run_started_at: 2026-08-21T11:22` — same ~11-day approval-gate delay). This diff touches only `backend/src/blocks/builtin/whip.rs`, zero overlap with the failing frontend file. Root-caused to the same repo-wide toolchain-drift issue as the rest of this batch (see run summary) — mechanically blocks approval per protocol regardless.

## Risks / edge cases

- Untracked background thread + discarded `JoinHandle`: if a flow is created and immediately stopped, up to 5 threads-worth of `send_event` calls (spread across ~4s) can execute after `stop_flow()` returns. Recommend the maintainer confirm this is benign (pads are safe to send events to post-teardown) or consider tying the retry loop to a cancellation flag/timeout tied to the flow's lifecycle.
- Sends extra PLI traffic on every session start (not just failure cases) — likely negligible for PLI/RTCP overhead, but worth noting as an always-on behavior change rather than purely defensive.

## Requested changes

Suggest (non-blocking): consider capturing a cancellation signal (e.g. an `Arc<AtomicBool>` set by the flow's stop path, or a bounded timeout tied to session teardown) so the retry thread can exit early instead of always running its full ~4-second course after a flow stop. Not required for correctness as verified, but would tighten the interaction with `stop_flow()`.

## Confidence: MEDIUM-HIGH — core logic directly confirmed against the diff; the one open question (thread/pad lifetime vs. flow-stop timing) is honestly flagged as unverified rather than asserted either way. CI failure conclusively unrelated to this diff. Verdict: Comment (mechanical CI gate), not Approve.

srperens (commented):
Reviewing PR #666: fix(whip): proactively retry keyframe (PLI) request on video pad connect
Verified against 48a8c56c (48a8c56ca9be11a0900021e77014feaa200eb05f) vs main 4a36815f (4a36815fed45fc495d548a7e7d99c4dffe13b992).

This supersedes my earlier `protocol=v2` review on this PR (id 4992799604), which never checked CI or traced the new thread's lifecycle against pipeline teardown. See Diagnosis and Blast radius below for what that pass missed.

**Summary**
- Adds a proactive, retried `UpstreamForceKeyUnitEvent` (PLI) request on a WHIP Input video pad as soon as it connects, on a dedicated background thread: 5 attempts, 800ms apart.
- This is a second, independent keyframe-recovery path alongside the existing DISCONT-triggered one, which only activates once decodebin has a decoder element in place.

**Claim verification**

|---|---|---|
| The existing keyframe-recovery mechanism only installs once a decoder element exists downstream of `decodebin` | CONFIRMED | `backend/src/blocks/builtin/whip.rs:642-677`: the `BUFFER`-probe-based DISCONT recovery is installed only inside `if element_klass.contains("Decoder") && element_klass.contains("Video")`, i.e. only after decodebin has already instantiated a video decoder element. |
| If the publisher's first keyframe is lost/incomplete before that decoder exists, decodebin waits forever for a keyframe with no active recovery, even though packets keep flowing | EXTERNAL | About `decodebin`'s internal autoplugging/keyframe-wait behavior when it cannot fully negotiate a decoder without an initial keyframe — not settleable from this repo. Plausible and consistent with well-documented decodebin/h264 decoder behavior; assumption stated. |
| New code proactively sends `UpstreamForceKeyUnitEvent` on the video pad on connect, retried 5×/800ms on a dedicated thread | CONFIRMED | `backend/src/blocks/builtin/whip.rs:846-865`: `std::thread::Builder::new().name(format!("whip-keyframe-retry-{}", retry_slot))...spawn(move \|\| { for attempt in 0..5 { ...retry_pad.send_event(fku); ...sleep(Duration::from_millis(800)); } })`. |
| `cargo test --workspace` passes; clippy clean on the changed file | CONFIRMED | CI job "Check (Linux)" (run 32479383689, job 96762297005) log: multiple `test result: ok. N passed; 0 failed` blocks including `509 passed; 0 failed`; zero failures across every reported binary. |
| "Verified via manual `GST_DEBUG`/pad-caps inspection during live diagnosis" | UNVERIFIED | Self-reported manual testing outside CI/this repo — per protocol the PR's own Verification claim is not itself evidence, and there is no artifact (log, test) in the repo corroborating it. Not contradicted, just not independently checkable. |

**Diagnosis**
Right root cause and right layer: the fix closes exactly the gap the diagnosis identifies — a proactive request that doesn't depend on a decoder already existing, complementing (not replacing) the existing reactive DISCONT probe. It is a REQUEST/RETRY mechanism, so per protocol the cost of the *response* matters, not the request:
- The loop is **OPEN**: it fires exactly 5 times over ~4 seconds on every single WHIP video pad connect, regardless of whether the publisher's first keyframe actually arrived cleanly. There is no check for "decode already succeeded, stop retrying" — `backend/src/blocks/builtin/whip.rs:853` (`for attempt in 0..5`) runs unconditionally once spawned.
- Cost on the healthy path (the common case, where the first keyframe already arrived fine): every video WHIP session now receives up to 5 extra forced-keyframe (PLI) requests within the first ~4 seconds, on top of whatever keyframe the publisher already sent. A remote encoder receiving a PLI typically responds with an out-of-schedule IDR frame, which is materially larger than a P-frame. I have not measured this repo's actual bitrate impact (that depends on the remote publisher's encoder/config, which is EXTERNAL), but the shape of the change is: every session, not just the failing minority, now asks for up to 5 unscheduled keyframes.
- This is a BOUNDED fix in one sense (5 attempts, 4-second window) — past that window, a publisher whose keyframe is still not decoding has no further proactive recovery from this new path (the existing DISCONT probe remains as a fallback once/if a decoder eventually appears).
- History check: this is the third recent PR touching WHIP Input session-start reliability from a different angle — `ce8bd63` (#665, jitterbuffer latency to prevent stall-on-start) landed already, and PR #663 (make RTX/do-retransmission configurable, draft) is open concurrently. Each targets a distinct mechanism as far as I can tell from the diffs, but three concurrent fixes in the same "WHIP session won't start cleanly" area is worth the maintainer keeping in view as a cluster, not three unrelated incidents.

**Blast radius**
LOCAL. The new code is entirely inline inside the `whipserversrc.connect_pad_added` closure (`backend/src/blocks/builtin/whip.rs:748` for the closure, new code at `:846-866`) — it is not a new shared function or type, so there are no other call sites to check.

Lifecycle/teardown trace for the additive thread (per protocol 3b): the closure captures `retry_pad = pad.clone()` (a `gst::Pad`), not a `gst::Element`/`Bin`/`Pipeline`, so CLAUDE.md's WeakRef-in-closure rule doesn't literally apply — but I traced the underlying question anyway. `GstObject`'s parent pointer is documented as a **weak** reference (child does not ref its parent; gstreamer.freedesktop.org `GstObject` docs: parent field described as "this object's parent, weak ref"), so holding a `gst::Pad` clone for the retry thread's ~4-second lifetime does **not** transitively keep the owning element/bin/pipeline alive, and will not trip the `pipeline_weak.upgrade().is_some()` P0 leak check in `stop_flow()` (`backend/src/state.rs:1398-1408`). If a flow is stopped mid-retry, `retry_pad.send_event(fku)` on an already-unparented pad is a documented-safe no-op in GStreamer (no peer/parent to deliver to), not a crash risk. The only residual cost is a harmless background thread that keeps running (and logging) for up to ~4s after an early flow stop — not a resource leak, just wasted work.

**Tests & CI**
- All non-platform-gated checks green: `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check (Linux)`, `Check & Build (WASM)`, `API Contract Check`. `Build (macOS)`/`Build (Windows)` skipped as usual (workflow_dispatch-only), but this diff is not platform-gated code, so that skip doesn't withhold compile coverage the way it does for PRs #668/#669.
- I confirmed via the raw CI log (not just the checkmark) that `cargo test` actually executed — see table above.
- No new automated test. The PR states none is feasible because it's runtime GStreamer pad/thread behavior with no WHIP/WHEP pipeline-construction test harness in this codebase — confirmed: `grep -rl "whip" backend/tests/*.rs` finds no WHIP-specific test file. I agree a meaningful regression test isn't practically achievable without such a harness; building one is a bigger undertaking than this PR's scope.
- Canaries: no pipeline reference-cycle risk introduced (see lifecycle trace above), so `pipeline_lifecycle_test.rs` isn't implicated; no API/WebSocket type changed, so the OpenAPI snapshot test isn't relevant.

**Risks / edge cases**
- Quantify-if-possible: the unconditional 5×/800ms PLI burst on every session's healthy path (see Diagnosis) is the main behavioral cost of this change and is not called out in the PR description, which frames it purely as a fix for the stall case.
- SPECULATIVE (not verified): interaction with PR #663 (making RTX/retransmission configurable) — if retransmission ends up disabled by default in some configurations, this retry mechanism becomes the primary keyframe-loss recovery path rather than a secondary one; not traced against #663's actual diff since it's a separate draft PR.

**Requested changes**
None blocking. Suggested, non-blocking: consider making the retry loop CLOSED (stop early once the video appsrc has actually received a valid frame/keyframe — `video_connected`/pad-caps state already tracked nearby in this function) so the healthy-path cost in Diagnosis drops from "always 5" to "0 in the common case." If the maintainer is fine with the current healthy-path cost, no code change is needed — this is a design tradeoff to confirm, not a defect.

**Confidence: MED** — the fix mechanism, closure captures, and pad-lifetime safety are all directly confirmed against the code and GStreamer's documented object-ownership model; the remaining open item is a design tradeoff (unconditional retry cost on the healthy path) rather than a correctness gap.

srperens:
Live finding from today that changes what this PR is, tracked in #685.

The existing DISCONT-triggered PLI probe (`whip.rs:642-678`) is installed from a `deep-element-added` handler on **`whipserversrc`**, but the video decoder lives in the **main** pipeline — `whipserversrc` outputs `application/x-rtp` and Strom's `decodebin_video_0` autoplugs `rtph264depay` → `h264parse` → `vtdec_hw` there.

`grep -c "Installed keyframe recovery probe"` over a full WHIP session log is **0**. The probe never installs, so there is no keyframe recovery on the WHIP path at all today.

So this is not "add a retry to a mechanism that installs too late" — the mechanism is dead code, and that is worth fixing directly (hook the probe where the decoder actually is) rather than adding a second, open-loop PLI source beside it. The review points above about the unconditional 5×800ms thread still stand.

srperens (commented):
Reviewing PR #666: fix(whip): proactively retry keyframe (PLI) request on video pad connect
Verified against 48a8c56c (48a8c56ca9be11a0900021e77014feaa200eb05f) vs main 4a36815f (4a36815fed45fc495d548a7e7d99c4dffe13b992).

This supersedes my earlier `protocol=v3` review on this PR (id 4995274424, same head). CI and the diff are unchanged, but issue #685 was filed afterward with a finding I independently re-verified in this repo's code, and it corrects a load-bearing claim in my prior review's Diagnosis section.

**Summary**
- Adds a proactive, retried `UpstreamForceKeyUnitEvent` (PLI) request on a WHIP Input video pad as soon as it connects: 5 attempts, 800ms apart, on a dedicated thread.
- My prior review framed this as "a second, independent keyframe-recovery path alongside the existing DISCONT-triggered one." That framing is wrong: the existing path never installs at all (see below), so this PR is not adding redundancy, it's adding the *only* working keyframe recovery on the WHIP Input path.

**Claim verification**

|---|---|---|
| An existing DISCONT-triggered PLI probe exists and is installed once a decoder element is added downstream | Technically true but misleading — see next row | `backend/src/blocks/builtin/whip.rs:642` (`if element_klass.contains("Decoder") && element_klass.contains("Video")`), inside the `deep-element-added` handler at `whip.rs:541` (`bin.connect("deep-element-added", false, move |values| {`), where `bin` at `whip.rs:540` is `whipserversrc.clone().downcast::<gst::Bin>()`. |
| (New, from issue #685) That handler is attached to `whipserversrc`'s own internal `Bin`, but the video decoder lives in a completely separate pipeline object, so the probe can never see it | CONFIRMED, independently traced | `whip.rs:952-955`: `session_pipeline.add(&whipserversrc)` — `whipserversrc` runs in its own per-session `gst::Pipeline` (built at `whip.rs:490`, `let session_pipeline = gst::Pipeline::builder()...`), created inside `create_whipserversrc_for_session` (`whip.rs:466`). The actual video decoder is instantiated by a *different*, statically-built `decodebin` (`whip.rs:348-398`, `let decodebin = gst::ElementFactory::make("decodebin")...name(&decodebin_id)`) inside `build_whipserversrc` (`whip.rs:152`), which feeds the codebase's separate per-slot appsrc→decodebin→videoconvert→tee chain in the main flow pipeline, not inside `whipserversrc`'s own `Bin`. `whipserversrc`'s internal `deep-element-added` (line 541) only ever sees elements GStreamer adds *inside* `whipserversrc` itself (its doc comment at `whip.rs:530` lists what it's for: "ICE policy, TWCC, keyframe recovery, auto-cleanup on ICE failure") — i.e. `rtpbin`/`webrtcbin`/`rtpsession` internals, none of which carry a `Decoder`+`Video` factory klass. The `if element_klass.contains("Decoder")` branch at line 642 is therefore unreachable code: it can never fire, because no element matching that klass is ever added inside the bin this handler is attached to. |
| New code proactively sends `UpstreamForceKeyUnitEvent` on the video pad on connect, retried 5×/800ms on a dedicated thread | CONFIRMED (unchanged from prior pass) | `whip.rs:846-865`, `std::thread::Builder::new().name(format!("whip-keyframe-retry-{}", retry_slot))...spawn(move \|\| { for attempt in 0..5 { ...retry_pad.send_event(fku); ...sleep(Duration::from_millis(800)); } })`. |
| CI green, tests pass (unchanged) | CONFIRMED | Same run as before (32479378678): `Check (Linux)` green, `509 passed; 0 failed`. |

**Diagnosis**
This is a bigger correction than a nitpick: my prior pass verified the *scoping condition* of the existing probe (only installs once a decoder klass element appears) but never asked *where* that decoder klass element would have to appear relative to the bin the handler is attached to. Having now traced it, the existing "existing DISCONT-triggered recovery" this PR's own code comment (`whip.rs:839-841`, "before decodebin has a decoder element to install the existing DISCONT-triggered recovery probe on") describes as a fallback is dead code — it has presumably never fired in production. That means:
- This PR is not adding a second, complementary recovery path. It is the *only* functioning keyframe-recovery mechanism on WHIP Input today.
- The PR description undersells its own importance and doesn't mention the dead-probe finding, likely because the author didn't know either — this isn't a criticism of the PR, it's new information the maintainer should have before merging.
- Per protocol 2g (request/retry mechanisms — cost the response, not the request): the loop is still **OPEN** (`whip.rs:853`, `for attempt in 0..5` runs unconditionally, no "decode already succeeded" check) and still costs every WHIP video session up to 5 forced-keyframe requests over ~4 seconds on the healthy path, exactly as before. But now that this is the *only* recovery mechanism rather than a supplement, an OPEN loop bounded at "5 attempts / 4 seconds" is a more consequential bound than my last review credited: past that window, with the DISCONT probe confirmed dead, there is now *zero* further keyframe recovery on WHIP Input for any session whose first keyframe is still missing after 4 seconds.
- History check (unchanged): third recent PR touching WHIP Input session-start reliability (`ce8bd63` / #665 jitterbuffer latency; #663 draft, RTX configurability). The dead-probe finding in #685 suggests this cluster of fixes has been operating without realizing one of the mechanisms they were building around never worked, which is itself worth the maintainer's attention as a signal that this area needs a harder look, not just another patch.

**Blast radius**
Still LOCAL for this diff's own code (`whip.rs:846-866`, inline in the `pad_added` closure, no new shared function/type). But the *finding* (dead DISCONT probe) is SHARED in effect: it means `whip.rs:642-678`'s ~35 lines are currently unreachable code sitting in the same file, a separate issue from this PR's diff but adjacent to it.

Lifecycle/teardown trace for the additive thread — unchanged from my prior pass and still holds: the closure captures a `gst::Pad` clone, not an `Element`/`Bin`/`Pipeline`; `GstObject`'s parent pointer is a documented weak reference, so this does not keep the pipeline alive past `stop_flow()` (`backend/src/state.rs:1398-1408`), and `send_event` on an unparented pad is a documented-safe no-op.

**Tests & CI**
Unchanged from my last pass: all non-platform-gated checks green (same run 32479378678, confirmed via raw log: `cargo test` executed, `509 passed; 0 failed`). No WHIP-specific test file exists (`grep -rl "whip" backend/tests/*.rs` finds none), and I still agree a pipeline-runtime regression test isn't practically achievable without a WHIP/WHEP test harness that doesn't exist in this repo. Canaries (`pipeline_lifecycle_test.rs`, OpenAPI snapshot) not implicated — no pipeline reference cycles or API/WebSocket type changes here.

**Risks / edge cases**
- The dead DISCONT probe (`whip.rs:530-681`) is effectively unreachable code shipping in the same file this PR touches. Not a defect introduced by this PR, but worth a maintainer decision: either fix its bin scoping (attach `deep-element-added` where the decoder actually lives) or remove it as dead weight, since it currently does nothing but occupies ~150 lines and the misleading code comment this PR itself references at `whip.rs:839-841`.
- Unconditional 5×/800ms PLI burst on every session's healthy path (unchanged finding from my prior pass) — now more load-bearing than previously assessed, since it's the only recovery path, not a backup to one.

**Requested changes**
1. Fix or remove the dead DISCONT-triggered probe (`whip.rs:540-681`): either move the `deep-element-added` connection to the pipeline that actually owns the decoder (the main/session flow pipeline built in `build_whipserversrc`, not `whipserversrc`'s own internal `Bin`), or delete the ~150 lines as unreachable code and correct this PR's own comment at `whip.rs:839-841` which currently describes it as a real fallback. This doesn't have to block *this* PR's merge, but it should not ship silently alongside it without the maintainer knowing.
2. Non-blocking (repeated from my last pass): consider making the retry loop CLOSED — stop early once a valid frame is observed — so the healthy-path cost drops from "always 5" to "0 in the common case," which now matters more given this is the sole recovery mechanism rather than a backup.

**Confidence: MED** — the retry mechanism itself, its closure captures, and pad-lifetime safety are all directly confirmed against the code and unchanged from my last pass. The new element — that the "existing" recovery mechanism this PR was meant to supplement is dead code — is independently re-verified here against `whip.rs`, not just taken from the issue report, but I have not confirmed whether the dead-probe finding is itself already common knowledge to the maintainer (it may be a known, accepted gap).

srperens:
To be clear up front: nothing we have merged or opened supersedes this PR. #687 fixes a different cause of the same "audio works, video does not" symptom (GL memory reaching `whepserversink`, #684), so it does not answer the stall this PR targets. Flagging that so this does not get closed as covered when it is not.

Two measurements from the same investigation that seem relevant here.

**The mechanism this PR extends never runs.** The description says the DISCONT-triggered recovery "only installs once a decoder element exists downstream of `decodebin`". It is stronger than that — it never installs at all. The probe at `whip.rs:642-678` is hooked on `whipserversrc`, while the decoder lives in the *main* pipeline (autoplugged by Strom's `decodebin_video_0`). Over a full session log:

```
$ grep -c "Installed keyframe recovery probe" session.log
0
```

So the chicken-and-egg framing is right, but the fix for it may be the probe's install site rather than a second mechanism alongside it. That looked like the more valuable change to us, and we have not touched it — it is your find.

**The symptom we could reproduce had a different cause.** Every "audio but no video" case we could reproduce deterministically (both from a browser and from `gst-launch`) was the GL-memory one, plus two others: `video-codecs` hard-coded to H264 in `whip.rs`, and a FEC-enabled publisher sending RED-encapsulated video that `decodebin` never depayloads. We never managed to reproduce a lost-first-keyframe stall, which does not mean it does not exist — only that we could not confirm it, so we cannot judge whether the retry helps.

Two things worth a look while it is open, offered as review rather than as a blocker:

- The retries are unconditional: five forced keyframes 800 ms apart on every video pad of every session, with no cancellation once decode succeeds. On a healthy session that is a bitrate spike at every connect.
- The thread captures a strong `pad.clone()` and cannot be stopped, so it keeps sending events for about four seconds after the session is torn down. `CLAUDE.md` has a rule about strong references to GStreamer objects in callbacks for the same class of reason.

A reproduction of the stall would settle it. If you have a capture or a publisher configuration that triggers it, that would be genuinely useful — we would like to be able to test it too.

srperens (commented):
Reviewing PR #666: fix(whip): proactively retry keyframe (PLI) request on video pad connect
Verified against 48a8c56c (48a8c56ca9be11a0900021e77014feaa200eb05f) vs main 7059442 (7059442d4fb65cd622b969cb603b9ccdc8b1c098).

This updates my earlier `protocol=v3` review on this PR (id 4997461057, same head 48a8c56c). Head SHA and CI are unchanged since that pass. A new comment (2026-08-24T08:44:11Z) from a repo member reports independent measurements from the same investigation that produced #684/#687, and corrects/reopens two things my prior review got wrong or overclaimed. I re-verified against the code directly rather than taking the comment's account on faith.

**Summary**
- Adds a proactive, retried `UpstreamForceKeyUnitEvent` (PLI) on a WHIP Input video pad as soon as it connects: 5 attempts, 800ms apart, on a dedicated unjoined thread.
- My prior review's central finding — that the pre-existing DISCONT-triggered recovery probe is dead code and never installs — is independently reconfirmed by the new comment via a runtime log grep, on top of my own static trace. That part stands.
- My prior review's dismissal of the retry thread's pad-lifetime risk as safe was asserted without a citable source. Re-examining it this pass, I can't actually settle it and am walking that part back to UNVERIFIED rather than repeating it as confirmed-safe.

**Claim verification**

|---|---|---|
| The DISCONT-triggered recovery probe (`whip.rs:642`, `if element_klass.contains("Decoder") && element_klass.contains("Video")`) is attached to `whipserversrc`'s own internal `Bin` via `deep-element-added` | CONFIRMED | `whip.rs:540-541`: `if let Ok(bin) = whipserversrc.clone().downcast::<gst::Bin>() { bin.connect("deep-element-added", false, ...` |
| That `Bin` only ever contains elements GStreamer adds inside `whipserversrc` itself (ICE/RTP internals), never the video decoder | CONFIRMED | `whip.rs:952-955`: `session_pipeline.add(&whipserversrc)` — `whipserversrc` lives in its own per-session `gst::Pipeline` (`whip.rs:490`, built inside `create_whipserversrc_for_session`, `whip.rs:466`). The video decoder is a separate `decodebin` (`whip.rs:351-356`, `let decodebin = gst::ElementFactory::make("decodebin").name(&decodebin_id)...` named `decodebin_video_{slot}`) built inside the *different* function `build_whipserversrc` (`whip.rs:152`), feeding the main/flow pipeline via the slot appsrc bridge (`whip.rs:807-818`, `target_appsrc` selection). No path connects that decoder into `whipserversrc`'s own `Bin`. |
| (New, from this comment) grepping a real session log for the probe's install message returns 0 matches | EXTERNAL | Reported measurement, not something I can reproduce without running a live WHIP session in this sandbox. Consistent with, and now doubly corroborating, the static trace above — I'm accepting it as consistent evidence, not as the sole basis for the CONFIRMED verdict on the row above. |
| The new retry loop is unconditional — 5 keyframe requests, 800ms apart, on every video pad of every session, with no check for "decode already succeeded" | CONFIRMED (unchanged from my prior pass) | `whip.rs:848-865`: `let retry_pad = pad.clone(); ... .spawn(move \|\| { for attempt in 0..5 { ... retry_pad.send_event(fku); ... std::thread::sleep(std::time::Duration::from_millis(800)); } })` — no early-exit condition in the loop body. |
| The spawned thread cannot be cancelled once started | CONFIRMED | `whip.rs:852`, `.spawn(move \|\| {...}).ok();` — the `JoinHandle` is discarded via `.ok()`; nothing stores it, so nothing can `join()`, abort, or signal it. |
| (My prior review, id 4997461057) "the closure captures a `gst::Pad` clone, not an `Element`/`Bin`/`Pipeline`... this does not keep the pipeline alive past `stop_flow()`... `GstObject`'s parent pointer is a documented weak reference" | UNVERIFIED — retracting my own prior claim | I checked the two functions that own WHIP session teardown, `WhipSessionManager::teardown_session_pipeline` (`backend/src/whip_session_manager.rs:410-423`) and `AppState::stop_flow` (`backend/src/state.rs:1289` onward, session teardown at `state.rs:1340-1358`). Neither calls `Bin::remove()` or any unparenting function on `whipserversrc` or its pads — teardown is `session_pipeline.set_state(gst::State::Null)` plus dropping Rust-side `HashMap`/`Vec` entries. Whether a live `gst::Pad` clone therefore keeps the parent `Element`/`Pipeline` referenced (via `GstObject`'s parent field) for the thread's ~4s lifetime is a question about `GstObject`'s ref semantics, which this repository does not vendor or document and which I have no way to check in this sandbox (no `pkg-config`, no GStreamer headers/source present). My prior review asserted "weak reference" as fact with no citation; that did not meet this protocol's evidence bar and I should not have written it that way. I'm not asserting the opposite either — this is genuinely open. |
| PR's motivating stall (lost/incomplete first keyframe before a decoder exists) | UNVERIFIED — unreproduced by anyone but the PR author | The new comment states the investigators could reproduce "audio but no video" from three *other* causes (`#684`'s GL-memory case, hardcoded `video-codecs = ["H264"]` at `whip.rs:523`, and FEC/RED-encapsulated video) but never the specific lost-first-keyframe stall this PR targets, and says so explicitly is not proof it doesn't exist. I independently confirm the H264 hardcoding they cite: `whip.rs:523`, `let video_codecs = gst::Array::new(["H264"]);`. The RED/FEC claim is about publisher/`decodebin` behavior — EXTERNAL, not checkable in this repo. |

**Diagnosis**
- The dead-probe finding from my prior pass still holds and is now corroborated by an independent runtime check, not just my static trace. That remains the most important fact for the maintainer regardless of what happens to this PR.
- Right fix for the stated problem, but the stated problem's own existence is now less certain than the PR description implies: nobody besides the original reporter has reproduced the lost-keyframe stall, despite reproducing three *other* "audio but no video" causes on the same investigation. This doesn't mean the fix is wrong, but it does mean approval would currently rest on an untested-by-anyone-else symptom.
- BOUNDED and OPEN (unchanged from my prior pass, protocol 2g): the retry loop fires 5 times / ~4 seconds on every session's video pad regardless of outcome, forcing up to 5 unnecessary IDR-driven bitrate spikes on the healthy path (the common case, since the failure this targets appears rare enough that even a dedicated investigation couldn't trigger it). Past 4 seconds, with the DISCONT probe confirmed dead, there is zero further recovery if the first keyframe is still missing.
- Other trigger of the same "no video, audio fine" symptom class that this PR does not touch: the hardcoded H264 codec list (`whip.rs:523`) and RED/FEC handling flagged in the new comment — both out of scope for this diff but worth the maintainer knowing they're open on the same code path.

**Blast radius**
LOCAL for the diff itself (`whip.rs:834-866`, inline in the existing `connect_pad_added` closure, no new shared function/type — unchanged from my prior pass). The dead-probe finding remains a SHARED-in-effect issue adjacent to this diff (`whip.rs:530-681`, ~150 lines of unreachable code in the same file) but is not part of this PR's changes.
Lifecycle/teardown for the additive thread: see the UNVERIFIED row above — I traced both teardown entry points in this repo and found no explicit pad/element unparenting, but could not settle the consequence without upstream GStreamer source unavailable in this sandbox. The concrete, repo-provable fact is the overlap window: the thread runs for up to ~4 seconds from the moment the video pad connects, with no cancellation, so any session torn down within that ~4s window (e.g. a publisher that connects and immediately fails) has this thread's `send_event` calls and pad reference outliving the teardown call by design, not by race — there's no code path that stops it early.

**Tests & CI**
Unchanged from my last pass (same head 48a8c56c): `Check (Linux)` green, `Build (Linux x86_64/ARM64)` green, WASM/API-contract/sccache green, `Build (macOS)`/`Build (Windows)` skipped (workflow_dispatch-only). No WHIP-specific test file exists (`backend/tests/*.rs` has none matching "whip"); I agree with my prior pass that a pipeline-runtime regression test isn't practically achievable without a WHIP/WHEP test harness this repo doesn't have — stated here rather than silently assumed. Canaries not implicated: no reference-cycle-relevant `Pipeline`/`Element`/`Bin` closure capture (`pipeline_lifecycle_test.rs` doesn't apply, though see the UNVERIFIED pad-lifetime row — that test's scope is specifically the WeakRef pattern for `Pipeline`/`Element`/`Bin`, not arbitrary `Pad` clones, so it would not catch this even if the underlying concern were real), no API/WebSocket type changes (openapi snapshot not implicated).

**Risks / edge cases**
- The dead DISCONT probe (`whip.rs:530-681`) remains unreachable code shipping in the same file this PR touches, now with two independent confirmations (my static trace, the new comment's log grep). Repeated from my prior review since it's still unaddressed.
- Unconditional 5×/800ms PLI burst on every session's healthy path (repeated from my prior pass, protocol 2g cost-of-response).
- The thread's pad-lifetime interaction with teardown is UNVERIFIED, not confirmed-safe as I previously stated — see above.

**Requested changes**
1. Fix or remove the dead DISCONT-triggered probe (`whip.rs:540-681`): move `deep-element-added` to the pipeline that actually owns the decoder, or delete the ~150 unreachable lines and correct this PR's own comment at `whip.rs:838-841` which currently describes it as a real fallback. Doesn't have to block this PR, but the maintainer should decide knowingly.
2. Make the retry loop closed rather than open: track whether a valid frame has been observed (e.g. via the existing `decoder_weak`/DISCONT-probe machinery once fixed, or a shared flag set from the appsrc/decodebin side) and stop retrying once decode succeeds, so the healthy-path cost drops from "always 5" to "0 in the common case."
3. Regardless of how the GstObject parent-ref question resolves, downgrade the retry thread's `gst::Pad` capture to a `WeakRef` (`ObjectExt::downgrade()`) and `upgrade()` inside the loop, returning early on `None`. This is a small, mechanical change (`whip.rs:848` and the loop body at `853-864`) that makes the pad-lifetime question moot rather than needing to be resolved, and matches the spirit of `CLAUDE.md`'s WeakRef guidance even though the rule's literal text only names `Pipeline`/`Element`/`Bin`.
4. Given nobody besides the original reporter has reproduced the specific stall this PR targets, ask for (or attach) a capture/repro before merge, per the new comment's own closing ask.

**Confidence: MED** — the dead-probe finding and the retry mechanism's own mechanics are directly confirmed against the code and doubly corroborated this pass. Confidence is capped at MED rather than raised because I had to retract part of my own prior review (the pad-lifetime safety claim) rather than resolve it, and because the motivating stall remains unreproduced by anyone but the PR author.

srperens:
Correction to my earlier comment: I said we could not reproduce a lost-first-keyframe stall. That is no longer true — we have it, and **your diagnosis was right**.

It reproduces from a browser publisher. `GST_DEBUG=rtph264depay:5` across one stalled session and four working ones in the same run:

| depayloader | Type 1 | Type 24 (STAP-A) | Type 28 (FU-A) | completed access units |
|---|---|---|---|---|
| depay0 | 7 | 1 | 15862 | 1370 |
| depay1 | 3 | 1 | 1824 | 165 |
| depay2 | 3 | 1 | 909 | 102 |
| depay3 | 8 | 2 | 17926 | 1434 |
| **depay4 (stalled)** | 4 | **0** | 14830 | **0** |

Every session that worked received at least one STAP-A — the packet that carries SPS/PPS. The stalled one received none, while 1987 complete fragmented NAL units arrived, so this is not packet loss: it is a stream with no parameter sets in it. Without them `rtph264depay` cannot set caps, so it never outputs an access unit, `decodebin` never exposes a pad, the flow's pipeline never leaves PAUSED, `whepserversink` never reaches PLAYING, and WHEP viewers get `502 Bad Gateway` while audio plays fine. That is the chicken-and-egg stall you described, with the parameter sets as the specific thing that goes missing.

Two things you found that I can also confirm:

- **The existing DISCONT recovery is worse than "only installs once a decoder exists".** It never installs at all — the decoder lives in the main pipeline, not the session pipeline (`grep -c "Installed keyframe recovery probe"` over a full session log returns 0). And it could not have worked if it had: an upstream force-key-unit sent from the main pipeline dies at that pipeline's `appsrc` and never crosses into the WebRTC source. #693 deletes it.
- **Your placement is the correct one.** Sending the request on the WHIP source's pad is the only side of the appsink/appsrc boundary from which it can reach the publisher. That is what #693 does too.

#693 differs from this PR in one respect: cancellation. It waits before asking rather than after, and stops the moment `decodebin` exposes a video pad for that slot, so a session that starts normally is never asked at all (verified: zero requests on a healthy publish). This PR asks five times 800ms apart on every session unconditionally, with no way to stop once video starts, from a thread that keeps firing for about four seconds after teardown. That was the substance of my earlier objection — the implementation, not the problem, and I should have separated those two more clearly the first time.

One thing #693 does *not* settle, and where your input would be valuable: in the instrumented run, four sessions that lived 20–30 seconds were asked five times and video still never started. `pad.send_event()`'s return value is not logged, so we do not know whether the force-key-unit reached something that turns it into a PLI at all. Relevant: `webrtcsrc`'s own `CustomUpstream` handling forwards such events over a gst-plugins-rs control data channel, which a browser does not speak — so the PLI, if there is one, comes from the event continuing upstream into `webrtcbin`'s `rtpsession`, which we have not confirmed at the protocol level. If you have a capture that shows the PLI on the wire, that would close it.

Whatever you decide to do with this PR is yours to call — the finding stands on its own, and it is credited in #693.

srperens:
## Closing this — #693 is merged, and this would now double the PLI source

Correcting my comment from 24 Aug: it was written at 10:24, one minute before #693 merged at 10:25, so it describes #693 as pending. And an earlier comment that day said "nothing we have merged or opened supersedes this PR" — that is no longer true, and I do not want that sentence left standing where someone reads the thread later.

**Your diagnosis was right and it is what shipped.** The chicken-and-egg stall is real, and your placement — sending the request on the WHIP source's pad, the only side of the appsink/appsrc boundary from which it can reach the publisher — is the placement #693 uses. That is credited in #693.

**Why this closes rather than merges.** Two things changed underneath it:

1. **The mechanism this PR extends no longer exists.** #693 deleted the dead DISCONT-triggered probe. On main, `whip.rs:689-696` is now a note explaining why it could never have worked (a force-key-unit sent from the main pipeline dies at that pipeline's `appsrc`).
2. **Applied to current main, this diff adds a second PLI source beside the one already there.** Main has `keyframe_request::request_until_decoding` (`whip.rs:876-900`), which stops the moment `decodebin` exposes a video pad for that slot — a healthy publish is never asked at all. This PR's `whip-keyframe-retry-{slot}` thread asks five times 800 ms apart unconditionally, with no cancellation. Together that is duplicate keyframe requests on every session, including healthy ones. The thread also holds a strong `pad.clone()` and keeps firing for about four seconds after teardown, which is the pattern CLAUDE.md rules out for GStreamer objects in callbacks.

So this is no longer "possibly redundant" — merging it on top of main would regress behaviour that #693 established.

**What would genuinely help, if you have the time.** You have the one thing we never had: a reproduction. Please test main against it — a browser publisher whose stream arrives with no STAP-A, so `rtph264depay` never gets parameter sets. If #693 does not cover your case, reopen this or file a new issue and I will treat it as a live bug rather than a closed one.

**And one open question where your input is worth more than ours.** #693 does not settle whether the force-key-unit becomes a PLI on the wire at all. `pad.send_event()`'s return value is not logged, and `webrtcsrc`'s own `CustomUpstream` handling forwards such events over a gst-plugins-rs control data channel that a browser does not speak — so the PLI, if there is one, has to come from the event continuing upstream into `webrtcbin`'s `rtpsession`. We have not confirmed that at the protocol level. A capture showing the PLI on the wire would close it.

Thanks for this one — the finding stands on its own, and it is in main because of you.

PR #667, opened 2026-08-10
fix(whip): only insert gldownload when decodebin autoplugs a GL-memory decoder

wagenet:
## Description

WHIP Input's video decode chain can autoplug a GL-accelerated decoder (e.g. `vtdec_hw`), producing `video/x-raw(memory:GLMemory)` instead of plain system-memory `video/x-raw`. `videoconvert` passes GL memory through unchanged, so downstream consumers that only accept system-memory video (e.g. WHEP Output's encoders) silently fail while audio keeps working.

`gldownload` fixes this, but it's a `GstGLBaseFilter` that unconditionally requires a negotiated GL context to reach `PLAYING`. Inserting it unconditionally would impose a hard GL dependency on every session, breaking decode on GL-less hosts (headless servers, WSL, CPU-only fallback).

This PR checks the decodebin video pad's negotiated caps at `pad-added` (same idiom as `thumbnail_tap.rs`) and only creates/links `gldownload` when the caps carry `memory:GLMemory`. Otherwise (or if caps are unavailable), it links directly to `videoconvert` as before.

## Type of Change

- [x] Bug fix

## Checklist

- [x] I have run `cargo clippy --workspace --all-targets -- -D warnings` (no new warnings vs. base)
- [ ] I have updated documentation as needed
- [ ] I have added tests for new functionality

No unit test is feasible: this is pipeline-construction/runtime GStreamer code, and no pipeline-construction test harness exists for WHIP/WHEP in this codebase. Verified manually via browser WHIP publish → WHEP view sessions on both the GL/hardware-decode path and a forced software-decode path (`GST_PLUGIN_FEATURE_RANK=vtdec_hw:0,vtdec:0`).

### DIFF
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -368,24 +368,100 @@ fn build_whipserversrc(
                     ElementPadRef::pad(&decodebin_id, "sink"),
                 ));
 
-                // decodebin has dynamic pads — connect pad-added to link to videoconvert
+                // decodebin can autoplug a GL-accelerated decoder (e.g. vtdec_hw),
+                // producing video/x-raw(memory:GLMemory) frames instead of plain
+                // system-memory video/x-raw. videoconvert passes GL memory through
+                // unchanged rather than downloading it, and downstream consumers
+                // (encoders for WHEP Output, etc.) generally only accept
+                // system-memory video/x-raw — without a gldownload, encoding
+                // silently fails ("no codec present that can handle the stream's
+                // type"). A gldownload is only inserted when the negotiated pad
+                // caps actually carry GL memory: gldownload is a GstGLBaseFilter,
+                // which unconditionally requires a negotiated GL context to reach
+                // PLAYING, so inserting it unconditionally would impose a hard GL
+                // dependency on every session and break decode on GL-less hosts
+                // (headless servers, WSL, CPU-only fallback — see gpu.rs). If
+                // caps are ever unavailable, default to the non-GL path: skipping
+                // gldownload when it was needed reproduces the already-diagnosed
+                // GL-memory bug, but inserting it when not needed reproduces a
+                // newer, harder-to-diagnose GL-context failure — the former is
+                // the safer default.
                 let videoconvert_weak = videoconvert.downgrade();
+                let decodebin_weak = decodebin.downgrade();
+                let instance_id_for_pad = instance_id.to_string();
                 decodebin.connect_pad_added(move |_dec, src_pad| {
                     if src_pad.direction() != gst::PadDirection::Src {
                         return;
                     }
-                    if let Some(vc) = videoconvert_weak.upgrade() {
-                        let sink = vc.static_pad("sink").unwrap();
-                        if !sink.is_linked() {
-                            if let Err(e) = src_pad.link(&sink) {
-                                warn!("Failed to link decodebin video pad to videoconvert: {:?}", e);
-                            } else {
-                                info!(
-                                    "WHIP Input: decodebin video pad linked to videoconvert for slot {}",
-                                    slot
-                                );
-                            }
+                    let Some(vc) = videoconvert_weak.upgrade() else {
+                        return;
+                    };
+                    let sink = vc.static_pad("sink").unwrap();
+                    if sink.is_linked() {
+                        return;
+                    }
+
+                    let is_gl_memory = src_pad
+                        .current_caps()
+                        .map(|caps| caps.to_string().contains("memory:GLMemory"))
+                        .unwrap_or(false);
+
+                    if !is_gl_memory {
+                        if let Err(e) = src_pad.link(&sink) {
+                            warn!("Failed to link decodebin video pad to videoconvert: {:?}", e);
+                        } else {
+                            info!(
+                                "WHIP Input: decodebin video pad linked to videoconvert for slot {}",
+                                slot
+                            );
                         }
+                        return;
+                    }
+
+                    let Some(bin) = decodebin_weak
+                        .upgrade()
+                        .and_then(|dec| dec.parent())
+                        .and_then(|p| p.downcast::<gst::Bin>().ok())
+                    else {
+                        warn!("WHIP Input: decodebin has no parent bin, cannot insert gldownload for slot {}", slot);
+                        return;
+                    };
+
+                    let gldownload_id = format!("{}:gldownload_{}", instance_id_for_pad, slot);
+                    let gldownload = match gst::ElementFactory::make("gldownload")
+                        .name(&gldownload_id)
+                        .build()
+                    {
+                        Ok(e) => e,
+                        Err(e) => {
+                            warn!("WHIP Input: failed to create gldownload for slot {}: {:?}", slot, e);
+                            return;
+                        }
+                    };
+
+                    if let Err(e) = bin.add(&gldownload) {
+                        warn!("WHIP Input: failed to add gldownload to bin for slot {}: {:?}", slot, e);
+                        return;
+                    }
+
+                    let gldownload_src = gldownload.static_pad("src").unwrap();
+                    if let Err(e) = gldownload_src.link(&sink) {
+                        warn!("Failed to link gldownload -> videoconvert: {:?}", e);
+                        return;
+                    }
+                    if let Err(e) = gldownload.sync_state_with_parent() {
+                        warn!("WHIP Input: failed to sync gldownload state for slot {}: {:?}", slot, e);
+                        return;
+                    }
+
+                    let gldownload_sink = gldownload.static_pad("sink").unwrap();
+                    if let Err(e) = src_pad.link(&gldownload_sink) {
+                        warn!("Failed to link decodebin video pad to gldownload: {:?}", e);
+                    } else {
+                        info!(
+                            "WHIP Input: decodebin video pad carries GL memory, linked through gldownload for slot {}",
+                            slot
+                        );
                     }
                 });
 


srperens (dismissed):
**Intent:** Fix silent encoding failures that occur when `decodebin` autoplugs a hardware decoder (e.g. `vtdec_hw` on macOS) that outputs `video/x-raw(memory:GLMemory)` frames. Plain `videoconvert` cannot consume GL memory, so downstream encoders never receive usable frames. The fix conditionally inserts a `gldownload` element between `decodebin` and `videoconvert` only when the negotiated caps contain `memory:GLMemory`.

**Risk:** Medium. Dynamic element insertion mid-pipeline inside a `pad-added` callback requires correct ordering (add → link → sync-state). The justification for the conditional approach (avoid unconditional GL context dependency on GL-less hosts) is sound and well-commented.

---

**What looks good:**

- All references captured in the `connect_pad_added` closure use `downgrade()` / `upgrade()` — correct per the project guidelines on GStreamer reference cycles.
- The add → link (`gldownload` src → `videoconvert` sink) → `sync_state_with_parent` → link (`decodebin` src → `gldownload` sink) sequence is the correct order for live element insertion.
- Falling back to the direct link when `gldownload` creation or bin insertion fails is the right defensive choice (reproduces the existing GL-memory bug rather than introducing a new GL-context failure).
- The detailed comment explaining why unconditional `gldownload` insertion would break GL-less hosts is helpful and should be kept.

**One note on the caps check:**

```rust
let is_gl_memory = src_pad
    .current_caps()
    .map(|caps| caps.to_string().contains("memory:GLMemory"))
    .unwrap_or(false);
```

String matching on serialised caps is fragile — a future GStreamer version could change the caps serialisation format. The more robust check would be:

```rust
let gl_ref_caps = gst::Caps::from_str("video/x-raw(memory:GLMemory)").unwrap();
let is_gl_memory = src_pad
    .current_caps()
    .map(|caps| caps.can_intersect(&gl_ref_caps))
    .unwrap_or(false);
```

That said, `memory:GLMemory` is part of the GStreamer GL memory API and has been stable for years, so the string check is unlikely to break in practice. Non-blocking.

srperens (dismissed):
Reviewing PR #667: fix(whip): only insert gldownload when decodebin autoplugs a GL-memory decoder
Verified against 37fce587 vs main 317e80b1.

This supersedes the earlier unverified approval on this PR, which lacked code citations. Verdict unchanged: approve, with one finding below.

**Summary**
- In the `decodebin` `connect_pad_added` callback in `build_whipserversrc`, conditionally inserts `gldownload` between `decodebin` and `videoconvert` only when the negotiated caps carry `memory:GLMemory`.
- Avoids an unconditional GL context dependency on GL-less hosts.

**Claim verification**

|---|---|---|
| Closure captures use WeakRef — no reference cycles | CONFIRMED | `backend/src/blocks/builtin/whip.rs:369` `videoconvert_weak = videoconvert.downgrade()`, `:370` `decodebin_weak = decodebin.downgrade()`. `bin` derived inside callback via `decodebin_weak.upgrade()…parent()` — no strong Bin/Element/Pipeline captured. |
| Element linking order: add → link downstream side → sync_state → link upstream side | CONFIRMED | `whip.rs` (PR head): `bin.add(&gldownload)` → `gldownload_src.link(&sink)` → `gldownload.sync_state_with_parent()` → `src_pad.link(&gldownload_sink)`. Correct order for live element insertion. |
| Falls back to direct link when gldownload creation fails | CONTRADICTED | When `ElementFactory::make("gldownload")` fails, the callback returns without making any link (`whip.rs`: the `Err` arm does `warn!` and `return`). The PR description says "Falling back to the direct link when gldownload creation or bin insertion fails" — this is inaccurate. The video pad is left unlinked if gldownload is unavailable while GL-memory caps are present. In practice, GL-memory caps imply GL plugins are present (so `gldownload` would be available), making this an unlikely real-world failure, but the claim is not what the code does. |
| Caps check `contains("memory:GLMemory")` is stable | UNVERIFIED (external) | GStreamer caps serialisation format; no test in this repo pins it. The legacy review noted this is a long-standing string that is unlikely to change. |

**Test coverage** — no test claimed. Verified manually via browser WHIP publish sessions (per PR description). No pipeline-construction test harness exists for WHIP/WHEP in this codebase — acceptable for this class of change.

**Risks / edge cases**
- If GL-memory caps appear but `gldownload` is somehow unavailable (contradictory in practice), the video pad is left unlinked rather than falling back to a direct `videoconvert` link. The PR description's claim about fallback behaviour is incorrect per the code as written. This is low-risk because the conditions that would trigger it are contradictory, but the description is misleading.
- SPECULATIVE: string-matching serialised caps (`caps.to_string().contains("memory:GLMemory")`) is less robust than `Caps::can_intersect`. A `gst::Caps::from_str("video/x-raw(memory:GLMemory)")` intersection check would be more idiomatic, though the string has been stable across GStreamer versions.

**Requested changes** — none that block merge. The fallback behaviour discrepancy is a documentation issue (PR description vs. code) with low practical impact. Consider correcting the PR description comment about fallback behaviour.

**Confidence: HIGH**

srperens (commented):
# Reviewing PR #667

head=`37fce587` (37fce587aa34bdc7c5a529b084d728c80b5843af) vs main=`317e80b1` (317e80b1993e4cab38d52dfed418bdf37f569f51)

## Summary

WHIP Input's `decodebin` can autoplug a GL-accelerated decoder (e.g. `vtdec_hw`) that outputs `video/x-raw(memory:GLMemory)` instead of system-memory `video/x-raw`. Plain `videoconvert` passes GL memory through unchanged rather than downloading it, silently breaking downstream consumers that need system memory. This PR checks the decodebin video pad's negotiated caps at `pad-added` and conditionally inserts a `gldownload` element only when the caps carry `memory:GLMemory`, otherwise linking directly to `videoconvert` as before.

## Claim verification

|---|---|---|---|
| 1 | GL-memory detection is only performed when caps are present, defaulting to the non-GL (direct-link) path when caps are unavailable | CONFIRMED | CODE: `let is_gl_memory = src_pad.current_caps().map(|caps| caps.to_string().contains("memory:GLMemory")).unwrap_or(false);` — `unwrap_or(false)` means missing caps → `is_gl_memory = false` → direct link path, matching the PR's stated "safer default" reasoning |
| 2 | The `caps.to_string().contains("memory:GLMemory")` idiom matches an existing precedent elsewhere in the codebase (not a novel/unreviewed technique) | CONFIRMED | CODE: `backend/src/gst/thumbnail_tap.rs:231-236` — pre-existing, unmodified code: `let is_gl = ... .and_then(\|p\| p.current_caps()).map(\|caps\| caps.to_string().contains("memory:GLMemory")).unwrap_or(false);` — identical idiom |
| 3 | `gldownload` is only added/linked into the pipeline when actually needed, avoiding an unconditional hard GL dependency | CONFIRMED | CODE: the `gldownload` `ElementFactory::make`/`bin.add`/link sequence is entirely inside the `if is_gl_memory { ... }` branch (reached only after the early-return for the non-GL case) |
| 4 | New closure captures only `WeakRef`s of GStreamer objects (`videoconvert_weak`, `decodebin_weak`), not strong references, inside the `connect_pad_added` signal handler | CONFIRMED | CODE: `let videoconvert_weak = videoconvert.downgrade(); let decodebin_weak = decodebin.downgrade();` before the closure; both `.upgrade()`'d inside with early-return on `None` — matches CLAUDE.md's required WeakRef-in-signal-closure pattern exactly |
| 5 | `gst::GLBaseFilter`-derived elements (like `gldownload`) unconditionally require a negotiated GL context to reach `PLAYING`, justifying the conditional-insert design | EXTERNAL, standard/well-documented GStreamer GL subsystem behavior — not independently re-derivable in this sandbox (no GStreamer runtime) | Consistent with GStreamer's documented GL element design (`GstGLBaseFilter` requires `GstGLDisplay`/context negotiation); presented as the PR's stated rationale for why unconditional insertion would break GL-less hosts |

## Diagnosis

Correctly scoped fix that follows an existing in-repo idiom (`thumbnail_tap.rs`) rather than inventing a new caps-inspection pattern, and correctly follows the CLAUDE.md WeakRef-in-signal-closure rule. The element construction/link/state-sync ordering (`bin.add` → link `gldownload` src to `videoconvert` sink → `sync_state_with_parent()` → link decodebin src pad to `gldownload` sink) is a standard dynamic-element-insertion sequence; I did not find any correctness issue with this ordering, though I could not execute it to confirm runtime behavior (no GStreamer runtime available in this sandbox).

## Blast radius

LOCAL — confined to the `decodebin.connect_pad_added` closure inside `build_whipserversrc` in `backend/src/blocks/builtin/whip.rs`. Only affects the WHIP Input video decode path; the non-GL branch is behaviorally identical to the pre-existing code (same direct link to `videoconvert`), so hosts/sessions that never autoplug a GL decoder see no behavior change.

## Tests & CI

No unit test added — the PR description explains why: "this is pipeline-construction/runtime GStreamer code, and no pipeline-construction test harness exists for WHIP/WHEP in this codebase," and states manual verification was done on both the GL/hardware-decode path and a forced software-decode path (`GST_PLUGIN_FEATURE_RANK=vtdec_hw:0,vtdec:0`) — I cannot independently confirm this manual testing occurred, only that the claim itself is coherent and consistent with the code.

CI: `check-runs` for `37fce587` shows the same two-job failure pattern (`Check (Linux)`, `Check & Build (WASM)`) with identical `frontend/src/themes.rs` `f32: From<f64>` fallback errors as PR #671/#668/#666 in this batch (run `31400515302`, `created_at: 2026-08-10`, `run_started_at: 2026-08-21T11:22` — same ~11-day approval-gate delay). This diff touches only `backend/src/blocks/builtin/whip.rs`, zero overlap with the failing frontend file. Root-caused to the same repo-wide toolchain-drift issue (see run summary) — mechanically blocks approval per protocol regardless of the unrelated root cause.

## Risks / edge cases

- None identified beyond the inherent untestability of this GStreamer runtime-negotiation path in CI (shared with #666, #668 in this batch — a recurring theme worth a maintainer-level decision on whether to invest in a pipeline-construction test harness, per the author's own acknowledgment).

## Requested changes

None. Code change is well-reasoned, follows existing precedent, and respects the WeakRef closure rule.

## Confidence: HIGH — all code-level claims directly confirmed against the diff and existing precedent; the two claims resting on GStreamer GL-subsystem internals are marked EXTERNAL rather than guessed. CI failure conclusively unrelated to this diff. Verdict: Comment (mechanical CI gate), not Approve.

srperens:
## Parking this: the bug is real, the fix is on the wrong side of the link

To be clear up front — the diagnosis here is good, and the caps-inspection idiom is the right tool. What we want to change is *where* it sits. Parking rather than closing so it can be revived if the consumer-side route stalls.

### The rule we're settling on

**A block emits the format it naturally produces. The consuming block builds an input front that accepts both system memory and GPU memory.**

The producer cannot know who is downstream, so a producer-side decision is always a guess. Concretely: WHIP Input -> GL vision mixer is a real flow, and there the mixer wants GL memory. This PR would make that flow pay `gldownload` + `glupload` per frame — a GPU->CPU->GPU round trip — to satisfy a consumer that never asked for system memory.

### This pattern is already the house style

- The GL consumers adapt on their own inputs: `compositor.rs:345-349`, and `vision_mixer/mod.rs:14` documents `queue_i -> glupload_i -> glcolorconvert_i -> tee_i`. `glupload` accepts both, so it is a safe front either way:
  ```
  $ gst-inspect-1.0 glupload
  SINK template: video/x-raw(memory:SystemMemory), video/x-raw(memory:GLMemory), video/x-raw(memory:DMABuf), ...
  ```
- The vision mixer already *defaults* to emitting GL memory downstream: `DEFAULT_GL_DOWNLOAD = false` in `types/src/vision_mixer.rs:142`.
- `thumbnail_tap.rs:230-268` is the closest precedent, and it is consumer-side: it reads the upstream tee's caps, and when they carry GL it builds `glcolorscale -> capsfilter(GLMemory) -> gldownload` *in its own branch*, scaling on the GPU before downloading. Same idiom as this PR, placed one link later.

### The stated downstream victim already accepts GL memory

```
$ gst-inspect-1.0 whepserversink
SINK template: 'video_%u'
  video/x-raw, video/x-raw(memory:GLMemory), video/x-raw(memory:CUDAMemory),
  video/x-raw(memory:NVMM), video/x-raw(memory:D3D11Memory), video/x-raw(memory:VAMemory)
```

`whepserversink` declares GL memory and builds its own converter chain internally. If video still dead-ends on that path, the fix belongs in the WHEP block or in how we drive webrtcsink — not in what WHIP Input hands over.

### The likely real fix: the `videoconvert` in WHIP Input should probably not be there

`whip.rs:358` creates a `videoconvert` per video slot and `whip.rs:392-396` links it straight to `video_out_tee`. There is **no capsfilter anywhere after it**, so nothing ever asks it to convert — with an unconstrained downstream it negotiates passthrough and does nothing. And the `decode=false` branch at `whip.rs:400-404` already links `appsrc -> video_out_tee` directly, so nothing structural requires an element there at all.

That element is also the entire reason this PR exists: it is what the GL pad has to be bridged into. Remove it and `decodebin`'s pad links straight to the tee, GL memory passes through the tee untouched (a tee is caps-agnostic), and each consumer adapts as it already does. When it *does* convert, it converts silently on the CPU and hides the cost from the operator.

This needs verifying before we act on it, not asserting: `thumbnail.rs`, `meter.rs`, `recorder.rs` and `tams_output.rs` create no converter of their own, so one of them may be leaning on this `videoconvert` today. I have not tested removal.

### One implementation note for whenever this idiom lands

The topology is decided from `src_pad.current_caps()` once at `pad-added`, with a fallback to the non-GL path when caps are unavailable. If the decoder renegotiates later there is no way back. A `glupload` / `autovideoconvert` front on the consumer handles both cases without inspecting anything — another reason to prefer that placement.

### Related

#668 fixes the same class of failure from the consumer side (`autovideoconvert` in front of a non-GL encoder, auto-plugging the `gldownload`/`glcolorconvert` bridge). That is the direction we're taking. The follow-up work is to move that decision out of the global `VideoConvertMode` in `gpu.rs` and into each consuming block's input front, since the choice is per link, not per process.

srperens:
Tracking the general refactor in #682 — that issue carries the convention, the list of blocks that need a caps-aware input front, and the open question about whether the WHIP Input `videoconvert` can be deleted outright.

srperens:
Reproduced this for real today on macOS — full write-up and measurements in #684.

**Your diagnosis is confirmed.** GL memory from `vtdec_hw` reaches `whepserversink` and the video path dies with exactly the error you quoted:

```
There is no codec present that can handle the stream's type
  webrtcsink/imp.rs(4867): Failed to look up output caps: No caps found for stream video_0
```

Deterministic, with both a browser publisher and a headless `gst-launch` one. I was wrong in my earlier comment to imply the path might negotiate fine because `whepserversink` advertises `memory:GLMemory` — the *pad* negotiates, the internal discovery pipeline does not.

What the measurements add: an explicit `gldownload` placed on the **consumer** side (in the WHEP block, in front of `whepserversink`) fixes it just as well — `no codec present` goes to 0, four video payloaders get discovered, and the sink pad lands on `video/x-raw, NV12`. So the fix does not need WHIP Input to give up its GPU-native output, which is the point of #682.

Keeping this parked for that reason, not because the bug is not real. It is, and #684 tracks it.

srperens:
Heads-up: #687 fixes this bug, and it is built directly on the diagnosis in this PR. The finding here was correct — a GL-accelerated decoder puts `video/x-raw(memory:GLMemory)` on the wire, `videoconvert` passes it through untouched, and the video then silently dies while audio keeps working. That is exactly what we measured.

The one thing #687 does differently is *where* it downloads. It adapts on the consumer side, in WHEP Output, rather than in WHIP Input:

- Other GL producers feeding WHEP Output are covered too, not just WHIP Input. The vision mixer defaults to emitting GL memory (`DEFAULT_GL_DOWNLOAD = false`), so `vision mixer -> WHEP Output` hits the same dead end and a WHIP-side download would not help it.
- `WHIP Input -> GL vision mixer` stays on the GPU. A producer-side download there would be followed by an immediate re-upload — a full round trip per frame, in a flow that has no reason to leave the GPU.

That is the convention now written down in `CLAUDE.md` under **GStreamer Memory Formats**, and it is what #682 asks for.

Some measurements from the investigation that may be useful either way:

- The reason no auto-converter can fix this: `whepserversink`'s `video_%u` sink template *advertises* `video/x-raw(memory:GLMemory)`, so negotiation succeeds and only the internal encoder discovery fails. There is a four-element reproduction on #684 that shows this outside Strom entirely.
- `caps.to_string().contains("memory:GLMemory")` works, but `caps.features(0)` gives the same answer without the string formatting — #687 uses that.

Would you mind testing #687 against the case you originally hit? If it covers it, this PR can go — but that call is yours, and if it misses something in your setup we would rather know than close it.

wagenet:
Agreed — closing in favour of #687.

_Claude's Analysis:_

> Consumer-side adaptation is the right placement. It covers `vision mixer -> WHEP Output` as well, which a WHIP-side download would have missed entirely given `DEFAULT_GL_DOWNLOAD = false`, and it keeps `WHIP Input -> GL vision mixer` on the GPU instead of paying a download/upload round trip per frame.

PR #668, opened 2026-08-10
fix(gpu): use autovideoconvert on macOS to bridge GL-memory video into non-GL encoders

wagenet:
## Summary
- `detect_gpu_capabilities()` chose between `autovideoconvert` and plain `videoconvert` based solely on NVENC/CUDA-GL interop availability. On macOS there is no NVENC, so it always fell back to plain `videoconvert`.
- The vision mixer's GL compositor feeds GL-memory frames downstream, and the selected encoder on macOS (`vtenc_h264_hw`) is a non-GL element. Plain `videoconvert` cannot accept `memory:GLMemory` caps, so negotiation dead-ended right after the mixer — no video ever reached the encoder or WHEP output (audio, on a separate non-GL path, was unaffected).
- `autovideoconvert` auto-inserts the needed `gldownload`/`glcolorconvert` bridge, and macOS has no CUDA-GL interop to test in the first place, so now always use it on macOS instead of running the CUDA-specific detection logic.

## Test plan
- [x] `cargo check` (workspace)
- [x] `cargo test --lib gpu::` (passes)
- [ ] Manual verification: run the backend on macOS with a vision mixer -> video encoder -> WHEP output flow and confirm video (not just audio) is negotiated and streamed

### DIFF
--- a/backend/src/gpu.rs
+++ b/backend/src/gpu.rs
@@ -1,10 +1,18 @@
 //! GPU capability detection and video conversion mode selection.
 //!
 //! This module detects at startup whether GPU-accelerated video conversion
-//! with CUDA-GL interop is supported. On systems where it works (native Linux
-//! with X11 and NVIDIA drivers), we use `autovideoconvert` for better performance.
-//! On systems where it fails (WSL, headless/GBM, broken interop), we fall back
-//! to software `videoconvert`.
+//! is supported, and picks between `autovideoconvert` (which auto-inserts
+//! bridge elements such as `gldownload`/`glcolorconvert` when GL-memory
+//! frames need to reach a non-GL element, e.g. an encoder) and plain
+//! `videoconvert` (software-only; cannot accept `memory:GLMemory` caps).
+//!
+//! - On macOS there is no CUDA to worry about, so `autovideoconvert` is
+//!   always used (see `detect_gpu_capabilities`).
+//! - On Linux, GPU accel additionally requires CUDA-GL interop with NVENC.
+//!   On systems where it works (native Linux with X11 and NVIDIA drivers),
+//!   we use `autovideoconvert` for better performance. On systems where it
+//!   fails (WSL, headless/GBM, broken interop), we fall back to software
+//!   `videoconvert`.
 
 use gstreamer as gst;
 use gstreamer::prelude::*;
@@ -224,6 +232,23 @@ pub fn detect_gpu_capabilities() -> VideoConvertMode {
     // Probe GL renderer info early (best-effort, independent of CUDA-GL interop)
     let gl_info = detect_gl_renderer();
     let _ = GL_RENDERER_INFO.set(gl_info);
+
+    // macOS has no CUDA/NVENC, so the interop test below (which is entirely
+    // CUDA-specific) never applies here and would always force the software
+    // fallback. But GL-memory frames still occur on macOS (e.g. from the GL
+    // video mixer) and need bridging to non-GL encoders like vtenc_h264_hw.
+    // autovideoconvert auto-inserts the needed gldownload/glcolorconvert
+    // bridge; plain videoconvert cannot accept memory:GLMemory caps at all.
+    // There's no known autovideoconvert breakage on macOS to guard against
+    // (unlike the WSL/CUDA-GL cases below), so always use it here.
+    if cfg!(target_os = "macos") {
+        info!(
+            "macOS detected - using GPU-accelerated video conversion (autovideoconvert bridges GL memory)"
+        );
+        let mode = VideoConvertMode::GpuAccelerated;
+        let _ = VIDEO_CONVERT_MODE.set(mode);
+        return mode;
+    }
 
     // Fast path: WSL has broken CUDA-GL interop, skip expensive test
     if is_wsl() {


srperens (dismissed):
**Intent:** Fix a case where the GPU capability detection on macOS always forced the software-only `videoconvert` path, even though `autovideoconvert` is both available and required on macOS. The CUDA-GL interop test that gates `autovideoconvert` on Linux is irrelevant on macOS (no CUDA), so the new early return bypasses it.

**Risk:** Low. The change is behind `cfg!(target_os = "macos")`, so Linux/Windows paths are completely unaffected.

---

**What looks good:**

- The reasoning is sound: on macOS, GL-memory frames from sources like the video mixer need `autovideoconvert` to bridge them (via `gldownload`/`glcolorconvert`) to non-GL encoders like `vtenc_h264_hw`. Plain `videoconvert` cannot consume `memory:GLMemory` caps at all.
- The early return correctly sets the global `VIDEO_CONVERT_MODE` and returns before reaching the Linux-specific CUDA detection path.
- The updated module doc comment accurately describes the two-platform difference.

**One cosmetic note (non-blocking):**

The log message says "GPU-accelerated video conversion" but on a CPU-only Mac (or any Mac without a GPU) `autovideoconvert` will still be selected and may use a software path internally. The phrasing could be slightly misleading. "Using autovideoconvert (bridges GL memory to downstream)" would be more accurate. Not a blocker.

srperens (dismissed):
Reviewing PR #668: fix(gpu): use autovideoconvert on macOS to bridge GL-memory video into non-GL encoders
Verified against 950746af vs main 317e80b1.

**Summary**
- Adds a macOS-specific early return in `detect_gpu_capabilities()` that unconditionally selects `VideoConvertMode::GpuAccelerated` (`autovideoconvert`), bypassing the Linux-only CUDA-GL interop test that was incorrectly forcing `videoconvert` on macOS.
- `videoconvert` cannot consume `memory:GLMemory` caps; `autovideoconvert` auto-inserts the required bridge elements.

**Claim verification**

|---|---|---|
| Before this change, macOS always used `videoconvert` | CONFIRMED | `backend/src/gpu.rs:233–260` (on main): `detect_gpu_capabilities()` runs `detect_gl_renderer()`, hits WSL check, then runs CUDA test. No macOS branch existed; CUDA is unavailable on macOS so the test falls through to the software fallback. |
| Change is gated by `cfg!(target_os = "macos")` — Linux/Windows paths unaffected | CONFIRMED | `backend/src/gpu.rs:248` (PR head) `if cfg!(target_os = "macos") { … return mode; }` |
| `VIDEO_CONVERT_MODE` is set before returning | CONFIRMED | `backend/src/gpu.rs:251–253` `let _ = VIDEO_CONVERT_MODE.set(mode); return mode;` — mirrors the pattern at the WSL and CUDA paths above it. |
| `autovideoconvert` auto-inserts bridge elements for GL memory | UNVERIFIED (external) | GStreamer `autovideoconvert` element behaviour; not tested in this repo. |

**Test coverage** — no test claimed. GPU capability detection depends on hardware availability and cannot be exercised in CI. The PR checklist ran `cargo test --lib gpu::`. Acceptable.

**Risks / edge cases**
- The `info!` log says "GPU-accelerated video conversion" for all macOS. On a Mac without discrete GPU, `autovideoconvert` may select a software GL path internally; the message could be misleading. Non-blocking cosmetic issue.
- SPECULATIVE: a macOS host with a broken or missing GL context could cause `autovideoconvert` to fail where `videoconvert` would have degraded gracefully. No evidence this occurs on supported macOS targets.

**Requested changes** — none.

**Confidence: HIGH**

srperens (commented):
# Reviewing PR #668

head=`950746af` (950746afae94ebcdb4dbbcb6e110d861a86037af) vs main=`317e80b1` (317e80b1993e4cab38d52dfed418bdf37f569f51)

## Summary

`detect_gpu_capabilities()` previously always fell back to plain `videoconvert` on macOS (since its CUDA/NVENC interop test is meaningless there), but the GL vision-mixer compositor still produces GL-memory frames that plain `videoconvert` cannot accept, silently breaking video (audio unaffected) into the macOS hardware encoder (`vtenc_h264_hw`). This PR adds an early `cfg!(target_os = "macos")` branch that unconditionally selects `VideoConvertMode::GpuAccelerated` (→ `autovideoconvert`, which auto-inserts `gldownload`/`glcolorconvert` bridges) before the CUDA-specific detection logic runs.

## Claim verification

|---|---|---|---|
| 1 | `VideoConvertMode::GpuAccelerated` maps to the `autovideoconvert` element | CONFIRMED | CODE: `backend/src/gpu.rs` — `VideoConvertMode::GpuAccelerated => "autovideoconvert"` in `element_name()`; existing test `element_name()` mapping asserts `"autovideoconvert"`/`"videoconvert"` (unchanged by this diff) |
| 2 | New macOS branch is an early return, placed before the CUDA/WSL-specific detection logic, so that logic never runs on macOS | CONFIRMED | CODE: `if cfg!(target_os = "macos") { ... return mode; }` inserted immediately after the GL-renderer info line and directly before the pre-existing `if is_wsl() { ... }` fast-path check |
| 3 | `autovideoconvert` auto-inserts `gldownload`/`glcolorconvert` bridges when needed, unlike plain `videoconvert` | EXTERNAL, plausible/standard GStreamer behavior — not independently verifiable in this sandbox (no GStreamer runtime/registry available to test negotiation) | This is standard, well-documented `autovideoconvert` behavior (it wraps `videoconvert`/`glcolorconvert`/`gldownload`/`glupload` and auto-selects based on negotiated caps); consistent with the module's own pre-existing doc comment describing the same behavior for the Linux/CUDA case |
| 4 | No known `autovideoconvert` breakage on macOS to guard against (unlike WSL/CUDA-GL cases) | UNVERIFIED | This is an absence claim (no known failure mode) that can't be positively confirmed from static code reading alone; taken as the author's operational experience, not independently falsifiable here |

## Diagnosis

The fix is a narrow, compile-time-gated (`cfg!`) early return — it cannot affect any non-macOS build path at all, which meaningfully limits blast radius even though the underlying claim (`autovideoconvert` always safe on macOS) is not independently testable here. The change is consistent with the module's existing structure and doesn't touch the CUDA/WSL detection logic it bypasses.

## Blast radius

LOCAL, and additionally **compile-time gated to macOS only** — `cfg!(target_os = "macos")` evaluates to a compile-time constant, so this branch is dead code (never taken) on Linux/Windows builds. Only affects `backend/src/gpu.rs`; two existing call sites (`backend/src/main.rs:421,576`, `backend/src/blocks/builtin/videoenc/tests.rs:17`) are unmodified and unaffected in their non-macOS behavior.

## Tests & CI

No new test was added for the macOS-specific branch — expected, since `cfg!(target_os = "macos")` code cannot be exercised on this repo's Linux-only CI runners (this is the same category of platform-only-testable gap identified in this batch's review of PR #669). The PR description's checklist appropriately marks "Manual verification" as unchecked/pending rather than claiming false coverage.

CI: `check-runs` for `950746af` shows `Check (Linux)` and `Check & Build (WASM)` both failed with the identical `frontend/src/themes.rs` `f32: From<f64>` fallback errors seen across this whole batch of long-pending `wagenet/*` PRs (run `31442466128`, `created_at: 2026-08-10`, `run_started_at: 2026-08-21T11:22` — an 11-day approval-gate delay). This diff touches only `backend/src/gpu.rs`, zero overlap with `frontend/`. Root-caused to the same repo-wide toolchain-drift issue as PR #671/#667/#666 in this batch (see run summary) — not introduced by this diff, but mechanically blocks approval per protocol regardless.

## Risks / edge cases

- The claim that `autovideoconvert` has no known macOS breakage mode is asserted, not proven — if wrong, this would regress the exact working case (audio-only fallback via plain `videoconvert`, which at least didn't silently drop video with a *worse* failure) into some other negotiation failure. Recommend the maintainer's own macOS manual test (checklist item currently unchecked) run before merge, independent of the CI gate.

## Requested changes

None on correctness grounds from static review. Recommend completing the manual macOS verification checklist item before merge, given this exact code path can't be exercised by CI.

## Confidence: MEDIUM — code change itself is small, correctly scoped, and compile-time gated to macOS-only; but the core behavioral claim (autovideoconvert always safe on macOS) rests on external/domain knowledge not verifiable in this sandbox, and no CI or automated test exercises the new branch at all. Verdict: Comment.

srperens:
## This is the direction we want — one correction and one follow-up

Fixing GL-memory-into-a-non-GL-consumer on the **consumer** side is the pattern we're standardising on: a block emits the format it naturally produces, and the consuming block builds an input front that accepts both system and GPU memory. #667 fixed the same class of failure on the producer side (forcing `gldownload` inside WHIP Input) and has been parked in favour of this approach — rationale there.

### Correction to the description

> Plain `videoconvert` cannot accept `memory:GLMemory` caps, so negotiation dead-ended right after the mixer

`videoconvert` does accept it:

```
$ gst-inspect-1.0 videoconvert
SINK template: 'sink'   video/x-raw, video/x-raw(ANY)
SRC template: 'src'     video/x-raw, video/x-raw(ANY)
```

The `video/x-raw(ANY)` template means GL memory negotiates fine and passes through in passthrough mode — `videoconvert` simply never downloads it. The dead end is one element later, at `vtenc_h264_hw`, which has no GL caps. The observed symptom and the fix are both right; only the attribution needs adjusting, and it matters because "videoconvert blocks GL" would suggest removing converters where the real answer is making the consumer's front GL-aware.

### Follow-up (not blocking)

`VideoConvertMode` in `gpu.rs` is a single process-wide choice, set once from the CUDA-GL interop probe in `main.rs:420`. That is why this fix needs a macOS special case in that function. The choice is really per link — whether *this* consumer's upstream carries GPU memory — so the decision wants to move into each consuming block's input front. `thumbnail_tap.rs:230-268` already does it that way: it reads the upstream tee's caps and builds either `glcolorscale -> capsfilter(GLMemory) -> gldownload` or `videoconvertscale` in its own branch. The blocks still fronted by a bare `videoconvert` (`videoformat.rs:104-143`, `videoenc.rs:117-179`) are the remaining gap.

Happy for this to land as-is with the description tweaked; the `gpu.rs` refactor is separate work.

srperens:
Follow-up tracked in #682 (moving the per-process `VideoConvertMode` decision into each consuming block's input front). Nothing blocking this PR.

srperens (commented):
Reviewing PR #668: fix(gpu): use autovideoconvert on macOS to bridge GL-memory video into non-GL encoders
Verified against 40aad5fb (40aad5fbb89beaa8e8f05d27097b8f34096c86ea) vs main 4a36815f (4a36815fed45fc495d548a7e7d99c4dffe13b992).

This supersedes my earlier `protocol=v2` review on this PR (id 4992799071), which never checked CI and evaluated against a stale head. See Blast radius and Tests & CI below for gaps that pass did not surface.

**Summary**
- On macOS, `detect_gpu_capabilities()` unconditionally selects `VideoConvertMode::GpuAccelerated` (i.e. `autovideoconvert`) before running any of the existing CUDA/NVENC-specific detection, bypassing the software fallback that macOS previously always hit.
- The affected mode is read through a single shared accessor used by several block builders, not just the vision-mixer path described in the PR summary (see Blast radius).

**Claim verification**

|---|---|---|
| Pre-fix, macOS always resolved to `VideoConvertMode::Software` because `nvh264enc` is never registered there | CONFIRMED | `backend/src/gpu.rs` (main, pre-PR) `detect_gpu_capabilities`: `let has_nvenc = registry.find_feature("nvh264enc", ...)...; if !has_nvenc { ... VideoConvertMode::Software ... }` — no NVENC feature exists on macOS, so this branch always fired. |
| Plain `videoconvert` cannot accept `memory:GLMemory` caps | EXTERNAL, verified against upstream docs | GStreamer docs for `videoconvert` (gstreamer.freedesktop.org/documentation/videoconvertscale/videoconvert.html): pad templates list only `video/x-raw` and `video/x-raw(ANY)` (`AHARDWARE_BUFFER`, `DMA_DRM`) — no `GLMemory` feature. |
| `autovideoconvert` auto-inserts `gldownload`/`glcolorconvert` to bridge GL memory to non-GL elements | EXTERNAL, corroborated by existing repo precedent | Not independently fetched from upstream docs, but the exact same claim is already encoded in this repo's own comment at `backend/src/blocks/builtin/vision_mixer/builder/pipeline_cpu.rs:107-109` ("Uses autovideoconvert on hosts with working CUDA-GL interop... Needed so that upstream GPU-memory sources... reach the CPU compositor"), predating this PR — this is the established pattern the PR extends, not a new one. |
| "No known autovideoconvert breakage on macOS to guard against, unlike WSL/CUDA-GL" | UNVERIFIED / flagged risk | The WSL/CUDA-GL detection machinery this function already contains exists precisely because blindly trusting the GPU-accelerated path broke real deployments — commit `921e3da` ("feat: Runtime GPU interop detection and headless Docker support", #215) added `is_wsl()` fast-path and the `test_cuda_gl_interop()` probe *because* headless Docker needed explicit `GST_GL_WINDOW`/`GST_GL_PLATFORM`/GBM/EGL setup for GL to even work. No equivalent probe or historical macOS-headless data point exists in this repo for the new branch; I found nothing that contradicts the PR's claim, but nothing in-repo that confirms it either — it rests entirely on the author's local hardware run. |
| Manual end-to-end verification (vision mixer -> encoder -> WHEP output actually streams video on macOS) | **Not claimed done** | PR's own "Test plan" checklist leaves this box unchecked: "Manual verification: run the backend on macOS with a vision mixer -> video encoder -> WHEP output flow and confirm video... " is `[ ]`, not `[x]`. Only `cargo check` and `cargo test --lib gpu::` are checked. |

**Diagnosis**
Right mechanism, right layer, and ABSOLUTE for the diagnosed cause on paper: it removes the wrong software-fallback default rather than adding a timeout/retry around a broken negotiation. It plausibly explains the reported symptom (video never reaches the encoder on macOS with the GL vision mixer active) — CONFIRMED that `videoformat`/`videoenc`/`ndi`/`devicesrc`/vision-mixer CPU-pipeline all read the same `video_convert_mode()` result (`backend/src/blocks/builtin/videoformat.rs:107`, `videoenc.rs:120`, `ndi.rs:209,296,493,577`, `devicesrc.rs:193`, `vision_mixer/builder/pipeline_cpu.rs:101`), so any of those GL-memory-producing paths hitting a non-GL sink would previously dead-end the same way.

History check (2f) surfaces a real, cited concern: this exact module exists because "just assume the GPU-accelerated path works" was already proven wrong once (WSL, headless Docker — `921e3da`). This PR applies the same kind of unconditional assumption to macOS, the one platform this repo's CI cannot build by default and that the author has not manually end-to-end verified (see table). I am not asserting this fails — I found no evidence it does — but the pattern match to a previously-broken assumption, combined with zero corroborating evidence beyond `cargo check`, is worth the maintainer's attention before merge.

**Blast radius**
SHARED, not LOCAL, and broader than the PR description implies. `gpu::video_convert_mode().element_name()` (or the `VideoConvertMode` it returns) is read at 8 call sites across 5 files: `videoformat.rs:107-108`, `devicesrc.rs:193`, `videoenc.rs:120-121`, `ndi.rs:209-210,296-297,493-494,577-578`, and `vision_mixer/builder/pipeline_cpu.rs:101`. On macOS this PR now forces `autovideoconvert` at *every one* of these, including plain device-capture and NDI paths that may never touch GL memory at all — not only the vision-mixer-to-encoder path the summary focuses on. This reuses an existing, already-shipped pattern (the `pipeline_cpu.rs` precedent above), so it isn't architecturally novel, but the behavior-change surface for macOS users is the whole block-builder set, not one path.

**Tests & CI**
- `Build (macOS)`: **skipped** (workflow_dispatch-only, off for push/PR per `.github/workflows/ci.yml:11`). Unlike PR #669's fully `#[cfg(target_os = "macos")]`-gated code, this PR's new branch uses `if cfg!(target_os = "macos")` (a runtime check, not an attribute), so it *does* get type-checked by Linux CI — but the branch body never executes there, so Linux CI cannot exercise the actual behavior change at all. Combined with the unchecked manual-verification box, there is currently zero automated or CI evidence this works, only the author's local run (not documented with output, unlike sibling PR #669's before/after table).
  Maintainer command: `gh workflow run ci.yml --ref wagenet/encoder -f platforms=macos`
- Linux (`Build x86_64`/`ARM64`, `Check`), WASM, and API Contract checks: all green.
- `cargo test --lib gpu::`: the only existing test touching this file, `test_video_convert_mode_element_name` (`backend/src/gpu.rs:373-379`), asserts the `element_name()` string mapping only — it does not call `detect_gpu_capabilities()` and provides no coverage of the new macOS branch's control flow. No new test was added.
- Canaries: no pipeline lifecycle or API/WebSocket type changes in this diff — `pipeline_lifecycle_test.rs` and the OpenAPI snapshot test are not implicated.

**Risks / edge cases**
- SPECULATIVE (not verified): headless/CI-like macOS environments (no physical display attached) may have the same class of GL-context-availability problem that headless Docker had for CUDA-GL (`921e3da`) — not traced against any macOS-specific GL backend code in this repo, since none exists for macOS.
- Related, unaddressed by this PR: open issues #674 ("Vision mixer inputs require BGRA... videoformat can't select it") and #682 ("Standardise on consumer-side memory-format adaptation (GL/system) between blocks") sit in the same code area. The PR doesn't reference either; worth the maintainer confirming this doesn't conflict with or get superseded by whatever direction #682 lands on.

**Requested changes**
1. Either complete the unchecked manual macOS end-to-end verification (vision mixer -> encoder -> WHEP, confirm video, not just audio, as the PR's own checklist already specifies) and report the result in the PR, or at minimum dispatch a macOS CI build so the branch is exercised on real hardware before merge: `gh workflow run ci.yml --ref wagenet/encoder -f platforms=macos`.
2. No further automated test is realistic without macOS CI (see Tests & CI) — I'm not asking for one beyond what CI can provide.

**Confidence: MED** — the diagnosed mechanism is sound and well-corroborated (upstream `videoconvert` docs plus this repo's own precedent in `pipeline_cpu.rs`), but the fix is SHARED across more call sites than the PR describes, is not covered by CI on the one platform it targets, and its own author has not checked the "does this actually work" box.

srperens:
Measured this today on macOS and it does not do what the description assumes — details in #684.

Inserting `autovideoconvert` in front of a non-GL consumer did **not** download the GL memory. It color-converted *inside* GL memory, `NV12` → `RGBA`, and left `memory:GLMemory` in place. The resulting failure was identical to doing nothing:

| Element in front of the sink | `no codec present` | video payloaders | caps at the sink pad |
|---|---|---|---|
| nothing | 1 | 0 | `video/x-raw(memory:GLMemory), NV12` |
| `gldownload` | **0** | **4** | `video/x-raw, NV12` |
| `autovideoconvert` | 1 | 0 | `video/x-raw(memory:GLMemory), **RGBA**` |

Same publisher, same flow, only that element differs. This was measured on the WHEP path rather than the vision-mixer-to-encoder path this PR targets, so it is not a direct refutation — but it is the same class of bridge, and the manual verification box on this PR is unticked.

Worth re-testing the vision mixer → `vtenc_h264_hw` → WHEP flow before merging: if `autovideoconvert` behaves the same way there, this would land as a fix that does not fix it. An explicit `gldownload` is what worked.

The direction (adapt on the consumer side) is still the right one — see #682.

srperens (commented):
Reviewing PR #668: fix(gpu): use autovideoconvert on macOS to bridge GL-memory video into non-GL encoders
Verified against 40aad5fb (40aad5fbb89beaa8e8f05d27097b8f34096c86ea) vs main 4a36815f (4a36815fed45fc495d548a7e7d99c4dffe13b992).

This supersedes my earlier `protocol=v3` review on this PR (id 4995244331, same head). Nothing in this repo changed since that pass, but issue #684 was filed afterward with a direct empirical test of the exact mechanism this PR relies on, and it changes the diagnosis below. The head SHA and CI are unchanged from the prior review; only new external evidence is added here.

**Summary**
- On macOS, `detect_gpu_capabilities()` unconditionally selects `VideoConvertMode::GpuAccelerated` (`autovideoconvert`) before running any CUDA/NVENC-specific detection, bypassing the software fallback macOS previously always hit.
- The new code carries a doc comment asserting "There's no known autovideoconvert breakage on macOS to guard against" (`backend/src/gpu.rs`, new comment above the `cfg!(target_os = "macos")` branch).
- The selected mode is read through a shared accessor used by 5 files / 8 call sites, not just the vision-mixer path the PR summary focuses on.

**Claim verification**

|---|---|---|
| Pre-fix, macOS always resolved to `VideoConvertMode::Software` because `nvh264enc` is never registered there | CONFIRMED | `backend/src/gpu.rs` `detect_gpu_capabilities()`: the pre-existing `if !has_nvenc { ... VideoConvertMode::Software ... }` branch always fired on macOS (no NVENC feature exists there). |
| "There's no known autovideoconvert breakage on macOS to guard against" (the PR's own added comment) | CONTRADICTED | Issue #684, filed against this exact code area after my prior review: on macOS 15 / Homebrew GStreamer 1.28.6, the reporter measured `autovideoconvert` placed ahead of a GL-memory-accepting sink (`whepserversink`) and found it converts color space in-place (NV12→RGBA) but does **not** download to system memory — `memory:GLMemory` caps persist and the downstream "no codec present" failure is identical to doing nothing at all. That is a documented, reproduced breakage in this codebase's own WHIP/WHEP flow, which the PR's shipped comment now asserts does not exist. |
| `autovideoconvert` auto-inserts `gldownload`/`glcolorconvert` to bridge GL memory to non-GL elements (general mechanism claim) | Narrower than stated — UNVERIFIED for this PR's actual call sites | Issue #684's counter-example is specific to `whepserversink`, whose sink pad template itself advertises `video/x-raw(memory:GLMemory)` (per the issue's own trace) — so GStreamer's autoplugger has no negotiation-driven reason to insert `gldownload` there; the failure is in `whepserversink`'s internal stream-discovery, not in the negotiated pad caps. This PR's 5 real call sites (`videoenc.rs:120`, `videoformat.rs:107`, `devicesrc.rs:193`, `ndi.rs:209/296/493/577`, `vision_mixer/builder/pipeline_cpu.rs:101`) feed elements that plausibly require true system memory — e.g. `videoenc.rs:126` (`gst::ElementFactory::make(&encoder_name)`, a hardware/software H264 encoder needing NV12/I420 system memory, not something with a GLMemory-accepting pad template like `whepserversink`). I did not trace every encoder/consumer element's actual pad template to confirm this distinction holds at all 5 sites — I'm flagging it as a real, unresolved gap rather than asserting the fix is fine because the counter-example is "different." |

**Diagnosis**
Right mechanism in principle (removing a stale software-only default), but the PR's own justification for skipping verification — "no known breakage" — is now factually false per #684, filed after my last pass and not yet addressed in this PR. This is not necessarily fatal to the fix (see the narrower-claim row above: #684's failure mode may not apply to this PR's actual call sites), but it removes the one argument the PR used to justify shipping without the checked-off manual macOS test. The PR's own "Test plan" still has the manual end-to-end verification box unchecked, and that verification is now more important than it was last pass, not less.

Not a request/retry mechanism — no cost-of-response analysis applies here.

**Blast radius**
SHARED, unchanged from my last review: `video_convert_mode()` is read at 8 call sites across 5 files (see table above), so this forces `autovideoconvert` at every one of them on macOS, not only the vision-mixer-to-encoder path the PR description focuses on.

**Tests & CI**
Unchanged since last pass (same head, same run 32479381320): Linux builds, WASM, API Contract, Check all green. `Build (macOS)` still `skipping` (workflow_dispatch-only). `cargo test --lib gpu::` still only covers `test_video_convert_mode_element_name` (`backend/src/gpu.rs:373-379`), which asserts the element-name string mapping and does not call `detect_gpu_capabilities()` — no coverage of the new macOS branch's control flow, and no new test was added to cover it.
Maintainer command for real macOS coverage: `gh workflow run ci.yml --ref wagenet/encoder -f platforms=macos`

**Risks / edge cases**
- Issue #684 is the concrete instance of the exact risk my prior review flagged as SPECULATIVE ("headless/CI-like macOS environments may have GL-context-availability problems") — it's no longer speculative, it's filed and reproduced, just in a different part of the codebase (WHIP/WHEP passthrough) than this PR's 5 call sites.
- Open issues #674 and #682 remain in the same code area and still aren't referenced by this PR.

**Requested changes**
1. Given #684 now contradicts the PR's own "no known breakage" comment, the unchecked manual macOS verification box in the PR's Test plan should be completed before merge — specifically exercising one of this PR's actual 5 call sites (e.g. vision-mixer → `videoenc.rs` → WHEP), not just relying on the general "autovideoconvert is fine on macOS" assumption the comment currently states as fact.
2. Update or remove the "no known autovideoconvert breakage on macOS" comment in `backend/src/gpu.rs` — it's no longer accurate as an unqualified statement now that #684 documents one. If the PR author believes the 5 real call sites are unaffected because they target true system-memory-only encoders (unlike `whepserversink`), say so explicitly in the comment so a future reader doesn't have to re-derive it.
3. If a macOS CI dispatch is feasible, run it: `gh workflow run ci.yml --ref wagenet/encoder -f platforms=macos`.

**Confidence: MED** — the mechanism and blast radius are directly confirmed against the code; the new element (issue #684) is real and material but I have not personally traced whether its failure mode reaches this PR's actual call sites, so I'm not asserting CONTRADICTED for the fix itself — only for the PR's "no known breakage" justification, which is a documented, sourced fact-check, not a guess.

srperens:
Some measured results that bear on this PR, from an investigation into the same symptom (#684, fixed in #687).

**The mechanism you describe is real.** `autovideoconvert` does bridge GL memory into a non-GL encoder when something downstream demands system memory, and plain `videoconvert` does not:

```
GL RGBA -> videoconvert         -> x264enc   FAIL  ("could not link videoconvert0 to x264enc0")
GL RGBA -> autovideoconvert     -> x264enc   OK
GL RGBA -> gldownload ! videoconvert -> x264enc   OK
```

And the fallback is broader than macOS. `gpu.rs` picks the software path on `if !has_nvenc`, while the vision mixer picks its GPU backend from `glvideomixerelement` + `has_hardware_gl()` — NVENC is not part of that decision. So an AMD or Intel Linux host with hardware GL emits GL memory into an encoder fronted by plain `videoconvert` and dies the same way. On NVIDIA hosts the global flag happens to land on `autovideoconvert`, which is why this has not been reported on Linux.

**But it cannot fix the WHIP -> WHEP case**, and the reason is worth knowing: `whepserversink`'s `video_%u` sink template *advertises* `video/x-raw(memory:GLMemory)`. Negotiation therefore succeeds with GL frames, nothing downstream ever asks for system memory, and no auto-converter has a reason to download — the failure is inside webrtcsink's encoder discovery, after a successful link. There is a four-element reproduction on #684 that shows it with no Strom code involved. WHEP Output also does not go through `video_convert_mode()` at all, so this diff does not reach that path.

For the record, plain `autoconvert` is not an option either — it segfaults on GL input (rc=139).

**So the encoder half of your diagnosis is open and worth fixing.** The direction #682 (item 2) proposes is a caps-aware input front on `videoenc` and `videoformat` rather than a per-OS switch of the process-wide `VideoConvertMode` — that way an AMD/Intel Linux host is covered by the same change, and the global flag can eventually retire. #687 landed a reusable helper for exactly this (`backend/src/gst/gl_bridge.rs`): it inspects the negotiated caps and splices a `gldownload` only when the frames really are in GL memory, leaving CUDA/NVMM/D3D11/VA alone.

Would you be interested in taking `videoenc` that way, reusing that helper? Happy to leave it to you since you found the encoder path — just say the word if you would rather we pick it up. Either way this PR should probably not merge as a global macOS switch, but the finding behind it should not be lost.

srperens (commented):
Reviewing PR #668: fix(gpu): use autovideoconvert on macOS to bridge GL-memory video into non-GL encoders
Verified against 40aad5fb (40aad5fbb89beaa8e8f05d27097b8f34096c86ea) vs main 7059442 (7059442d4fb65cd622b969cb603b9ccdc8b1c098).

This updates my earlier `protocol=v3` review on this PR (id 4997454618, same head 40aad5fb). The head SHA and CI are unchanged since that pass. What changed is a new comment (2026-08-24T08:43:51Z) from a repo member with measured pipeline results from the same investigation that produced #684/#687, which resolves the gap my last review left UNVERIFIED — whether this PR's actual 5 call sites are affected by the same class of failure as #684. I independently re-verified the code-side claims in that comment below rather than taking them at face value.

**Summary**
- On macOS, `detect_gpu_capabilities()` unconditionally selects `VideoConvertMode::GpuAccelerated` (`autovideoconvert`) before any CUDA/NVENC-specific detection runs, replacing the software fallback macOS previously always hit.
- The new comment confirms the mechanism (autovideoconvert bridges GL memory to non-GL encoders) is real, but shows the underlying gap is broader than macOS and that this PR's global per-OS flag is the wrong shape for it.

**Claim verification**

|---|---|---|
| Pre-fix, macOS always resolved to `VideoConvertMode::Software` | CONFIRMED | `backend/src/gpu.rs:268` `detect_gpu_capabilities()`: `if !has_nvenc {` — `nvh264enc` is never registered on macOS, so this branch always fired before the PR's new macOS-first check was added. |
| `video_convert_mode()` is a single process-wide flag read at multiple call sites | CONFIRMED | `backend/src/gpu.rs:43` `static VIDEO_CONVERT_MODE: OnceLock<VideoConvertMode> = OnceLock::new();` — set once in `detect_gpu_capabilities()`, read via `video_convert_mode()` in `videoenc.rs:120`, `videoformat.rs:107`, `devicesrc.rs:193`, `ndi.rs:209/296/493/577`, `vision_mixer/builder/pipeline_cpu.rs:101`. |
| "gpu.rs picks the software path on `if !has_nvenc`, independent of `has_hardware_gl()`" | CONFIRMED | `backend/src/gpu.rs:264-273`: the Linux branch sets `VideoConvertMode::Software` purely from `registry.find_feature("nvh264enc", ...)`, with no reference to GL renderer detection. |
| "the vision mixer picks its GPU backend from `glvideomixerelement` + `has_hardware_gl()` — NVENC is not part of that decision" | CONFIRMED | `backend/src/blocks/builtin/compositor.rs:214` `if has_gl && gpu::has_hardware_gl() {` — no `has_nvenc`/`video_convert_mode` reference in this selection path. |
| So an AMD/Intel Linux host with hardware GL but no NVENC can pick the GL vision-mixer backend while `videoenc` still gets `videoconvert` (Software mode) | CONFIRMED (mechanism) | Follows directly from the two preceding rows: `compositor.rs:214`'s GL selection and `gpu.rs:268`'s Software fallback are independent boolean checks on independent conditions (hardware GL vs. NVENC presence), so they can diverge on the same host. |
| "WHEP Output also does not go through `video_convert_mode()` at all, so this diff does not reach that path" | CONFIRMED | `backend/src/blocks/builtin/whep.rs` has no `video_convert_mode` import or call (absent from the 5-file/8-call-site list above); its GL handling is `gl_bridge::install_gl_download_bridge` (`whep.rs:13,1727`, added by #687, not present on this PR's branch). |
| `whepserversink` advertises `video/x-raw(memory:GLMemory)` and fails encoder discovery after a successful link, independent of any producer-side converter choice | CONFIRMED (repo docs) | `CLAUDE.md` "GStreamer Memory Formats": "`whepserversink` accepts `video/x-raw(memory:GLMemory)` and then fails encoder discovery. A successful link is not proof the consumer can use the frames." This is exactly why this PR's `autovideoconvert` switch cannot fix the WHIP→WHEP case even where it does apply. |
| Measured pipeline behaviour: `videoconvert` fails to link GL RGBA to `x264enc`, `autovideoconvert` and `gldownload!videoconvert` both succeed | EXTERNAL | Reported as directly measured by the comment author, not something I can execute in this sandbox (no GStreamer runtime). Assumption stated: taking the report at face value as it matches the documented `gldownload`/GL-memory semantics in `gst/gl_bridge.rs`'s module docs (added by #687) and CLAUDE.md's "GStreamer Memory Formats" section. |
| "this PR should probably not merge as a global macOS switch" (recommendation, not a fact about the repo) | N/A — recommendation | Not a checkable claim; addressed under Diagnosis below. |

**Diagnosis**
- Mechanism is real but the PR's chosen fix is the wrong shape for it. The underlying property is per-block ("does *this* block's input carry GL memory"), decided either at build time (when the producer is known, e.g. the vision mixer's backend choice) or from negotiated caps (when it's not, e.g. `decodebin` autoplugging — the case `gl_bridge` was built for in #687). `VIDEO_CONVERT_MODE` is neither: it's one process-wide flag gating the converter element name at every one of 8 call sites simultaneously, keyed only on OS/NVENC presence.
- BOUNDED, and the bound is now known precisely: the fix only helps call sites where GL memory can actually reach a system-memory-only consumer *and* the consumer doesn't itself advertise a GL-memory pad template. It does not help `whepserversink` (advertises GL memory, so no auto-converter ever gets a reason to fire — already fixed separately in #687) and, per the newly-confirmed independence of `has_hardware_gl()` from `has_nvenc`, it does not cover AMD/Intel Linux hosts hitting the identical `videoenc.rs` failure this PR targets on macOS. The PR fixes one OS out of at least two known-affected configurations.
- Other trigger of the same failure mode left open: any non-macOS, non-NVENC Linux host with hardware GL (AMD/Intel) and the vision mixer auto-selecting the GL backend. This PR's `cfg!(target_os = "macos")` branch does nothing for it; `gpu.rs:268`'s `if !has_nvenc` still forces `Software` there.
- Right layer: no. Per `CLAUDE.md` "GStreamer Memory Formats" — "Adapt at build time where the input is known... Where it depends on what `decodebin` autoplugged upstream, decide from the negotiated caps — `gst::gl_bridge` does this for GL memory" — the convention this repo has since adopted (in #687, merged after this PR was opened) is a caps-or-build-time-aware per-consumer adapter, not a global per-OS runtime flag. This PR predates that convention landing, but re-reviewing today, it conflicts with it.
- Not a request/retry/feedback mechanism — no response-cost analysis applies.

**Blast radius**
GLOBAL, larger than my previous pass scored it. `video_convert_mode()` gates 8 call sites across 5 files on every session (`videoenc.rs:120`, `videoformat.rs:107`, `devicesrc.rs:193`, `ndi.rs:209/296/493/577`, `vision_mixer/builder/pipeline_cpu.rs:101` — confirmed by grep, not individually re-read this pass since unchanged from prior review). New this pass: the mechanism is confirmed to generalize past macOS to any Linux host with hardware GL and no NVENC, which this PR does not address and does not mention. That widens "who is affected by the underlying bug" without widening "who this PR fixes" — a reviewer scanning only the PR title would reasonably assume macOS is the whole story.

**Tests & CI**
No new check-run activity since my last pass (same head 40aad5fb): `Check (Linux)`, `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check & Build (WASM)`, `API Contract Check`, `sccache preflight` all `success`; `Build (macOS)` and `Build (Windows)` `skipped` (workflow_dispatch-only, per `.github/workflows/ci.yml`). `cargo test --lib gpu::` still only exercises `test_video_convert_mode_element_name` (`backend/src/gpu.rs:373-379`), which checks the string mapping, not `detect_gpu_capabilities()`'s new macOS branch — no test covers the control-flow change this PR makes.
Maintainer command for actual macOS coverage: `gh workflow run ci.yml --ref wagenet/encoder -f platforms=macos`

**Risks / edge cases**
- The PR's own comment "There's no known autovideoconvert breakage on macOS to guard against" (`backend/src/gpu.rs:242-243` in the diff) is now doubly stale: #684 documented a macOS breakage in a different call site (WHIP→WHEP, since fixed by #687), and this pass's new evidence shows the *scope* framing ("on macOS") was itself too narrow — the gap this PR half-closes exists on non-NVENC Linux too.
- SPECULATIVE (not verified): I have not traced whether any of the 5 real call sites' consumer elements (the specific encoder/converter factories selected by `encoder_name` in `videoenc.rs`, or the NDI/devicesrc paths) themselves ever advertise a GL-accepting pad template the way `whepserversink` does. If any do, this PR's fix would be similarly inert for them, for the same reason it's inert for WHEP.

**Requested changes**
1. Either narrow this PR's scope explicitly (update the description and the `gpu.rs` comment to state it only fixes the macOS case and that Linux AMD/Intel hosts remain affected — smallest concrete change: rewrite the comment block at `backend/src/gpu.rs:236-243` to say so instead of asserting no known breakage), or fold the fix into the caps-aware direction #682 (item 2) is proposing, reusing `gst/gl_bridge.rs`'s pattern for `videoenc.rs`'s build-time-known case, matching CLAUDE.md's "GStreamer Memory Formats" guidance instead of a global flag.
2. Complete the still-unchecked manual macOS verification box in the PR's Test plan before merge, since the mechanism (autovideoconvert bridging GL to a system-memory encoder) has now been independently measured by someone else but not by anything in this repo's CI or by this PR's own stated verification.
3. If proceeding with the global switch as-is despite (1), add a code comment at `backend/src/gpu.rs:268` noting the AMD/Intel Linux gap so it isn't rediscovered from scratch a third time.

**Confidence: MED** — every code-side claim in the table above is CONFIRMED against this repo; the remaining uncertainty is the actual GStreamer link/negotiation behavior (EXTERNAL, stated) and the unswept SPECULATIVE row on the other 4 call sites' pad templates.

srperens:
## Closing this in favour of #682 step 2 — the finding is preserved, and it is bigger than this diff

Following up on my 24 Aug comment, which asked whether you wanted to take the `videoenc` work and never got an answer. Rather than leave this sitting as a PR that implies it might merge, I am closing it and moving the substance where it can be acted on.

**Your encoder-half diagnosis is unfixed and still wanted.** #687 only fixed the WHEP path (GL memory reaching `whepserversink`, a case no auto-converter can help with because that sink *advertises* `memory:GLMemory` and negotiation therefore succeeds). It does not touch the vision mixer to `vtenc_h264_hw` path this PR targets. Nothing that has merged supersedes your finding.

**Why the diff's shape is what closes it, not the finding.** The change switches the process-wide `VideoConvertMode` on `cfg!(target_os = "macos")`. I have now verified on main that the underlying problem is not macOS-specific:

- `gpu.rs:239-245` returns `VideoConvertMode::Software` (bare `videoconvert`) whenever `!has_nvenc`
- `compositor.rs:182` + `compositor.rs:214` pick the GL backend from `glvideomixerelement` plus `gpu::has_hardware_gl()`, with NVENC playing no part in that decision

So an AMD or Intel Linux host with hardware GL emits GL memory into a `videoconvert`-fronted encoder and dies exactly as macOS does. NVIDIA hosts happen to land on `autovideoconvert`, which is the only reason this has never been reported on Linux. A per-OS switch would have fixed one of the three cases.

I have written that generalisation up as a comment on #682, credited to you, so it survives this PR closing: https://github.com/Eyevinn/strom/issues/682#issuecomment-5423683223

**The change that covers all three cases** is #682 step 2: give `videoenc` and `videoformat` a caps-aware input front following the `thumbnail_tap.rs:230-268` shape, reusing the `gst::gl_bridge` helper that #687 landed. It inspects the negotiated caps and splices a `gldownload` only when the frames really are in GL memory, leaving CUDA/NVMM/D3D11/VA alone — so it does not need a per-OS branch, and it lets the global `VideoConvertMode` retire per call site instead of growing another special case.

**The offer from 24 Aug still stands, and I would rather you took it than us.** You found the encoder path; the helper you would build on is already in main. If you would like it, say so and it is yours — reopen this with the new shape or open a fresh PR, whichever you prefer. If you would rather hand it over, say that instead and we will pick it up and credit the diagnosis. Either answer is a good one; I just did not want the finding to expire quietly in a stale PR.

For the record, the two corrections from earlier in this thread also still hold: `videoconvert` does accept `video/x-raw(ANY)` and passes GL memory through rather than blocking it (the dead end is the encoder one element later), and plain `autoconvert` is not an alternative because it segfaults on GL input.

PR #670, opened 2026-08-11
docs: correct the X11/Docker-only claim for HTML rendering on macOS

wagenet:
## Description

`docs/HTML_RENDER.md` currently says, under Limitations:

> **Docker only**: CEF requires X11, which the strom-full image provides via Xvfb

The X11 requirement is Linux-specific. CEF on macOS renders offscreen through its own path and never touches X11 or Xvfb, so the stated reason doesn't hold there.

This splits the bullet per platform:

- **Linux** — keeps the existing (correct) rationale: CEF needs an X server, `strom-full` provides Xvfb, which is why it's the supported path for HTML sources.
- **macOS** — states the real situation. Native macOS is still not supported, but for different reasons, and they're worth naming: gstcefsrc needs macOS build fixes ([centricular/gstcefsrc#110](https://github.com/centricular/gstcefsrc/pull/110)), headless needs a Cocoa run loop on the main thread (#669), and CEF refuses to initialise unless the host process is inside an `.app` bundle.

That last point is the one worth documenting: an unbundled process doesn't error, it hangs silently at PAUSED with no diagnostic. It's easy to lose an afternoon to.

I deliberately did **not** claim macOS is supported — neither prerequisite has merged. The goal is only that the stated *reason* is correct, so someone on macOS isn't sent looking for an X server they don't need.

Docs-only; no behaviour change.

## Related Issue

None. Companion to #669.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

- [x] I have run `cargo fmt --all` — clean (no code changed)
- [ ] I have run `cargo clippy --workspace --all-targets -- -D warnings` — fails on `main` independently of this change (pre-existing `float-literal-f32-fallback` errors in `frontend/*.rs` on Rust 1.97.1, addressed by #664). No code changed here.
- [x] I have run `cargo test --workspace` — passes
- [x] I have updated documentation as needed — this *is* the documentation update; navigational per CONTRIBUTING, not code-describing
- [x] I have added tests for new functionality — n/a, docs only

---

### DIFF
--- a/docs/HTML_RENDER.md
+++ b/docs/HTML_RENDER.md
@@ -3,7 +3,7 @@
 > **Code is the source of truth.** This guide describes intended behaviour and may have
 > drifted from the current implementation. When in doubt, read the code and check the in-app UI.
 
-Strom supports rendering HTML content as video sources using the `cefsrc` GStreamer element from [gstcefsrc](https://github.com/AioCef/gstcefsrc). This enables:
+Strom supports rendering HTML content as video sources using the `cefsrc` GStreamer element from [gstcefsrc](https://github.com/centricular/gstcefsrc). This enables:
 
 - Dynamic HTML/CSS/JavaScript overlays
 - Web-based graphics and animations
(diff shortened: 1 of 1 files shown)


srperens (dismissed):
**Intent:** Correct stale and inaccurate content in `docs/HTML_RENDER.md`:
- Fix upstream repository URLs (`AioCef` → `centricular` / `chromiumembedded`).
- Replace the blanket "Docker only" claim with accurate per-platform limitations.

**Risk:** Documentation-only, no runtime impact.

---

**What looks good:**

- The corrected URLs (`centricular/gstcefsrc`, `chromiumembedded/cef`) are the current canonical locations.
- The per-platform breakdown (strom-full image only, Linux X11, macOS `.app` bundle requirement) is a meaningful improvement over the previous one-liner.
- The disclaimer ("Code is the source of truth — this may have drifted") is already present.

**One small suggestion:**

The macOS limitation text references PR #669 directly: `...and [Eyevinn/strom#669](https://github.com/Eyevinn/strom/pull/669)`. Once #669 is merged the PR number becomes a closed-PR link, which is less useful than an issue number or a plain description. Consider replacing the PR reference with a note like "a Cocoa run loop fix is in progress" or simply omitting the in-flight PR number.

Non-blocking — the docs are clearly better than before regardless.

PR #671, opened 2026-08-11
fix: give CEF a per-instance cache directory on native runs

wagenet:
## Description

Only one Strom instance per machine can use `cefsrc` (HTML sources, DSK graphics) when running natively. Starting a cefsrc flow on a second instance fails:

```
GstCefSrc:c: gst_base_src_start (): Failed to start
→ HTTP 500 "Pipeline state change failed: Element failed to change its state"
```

Chromium keys its process singleton on the cache/profile directory. `strom-full` sets `GST_CEF_CACHE_LOCATION=/tmp/cef-cache` in its entrypoint, so containers are unaffected — but native runs never set it and inherit Chromium's default, which CEF itself warns about on every startup:

```
WARNING:cef/libcef/common/resource_util.cc:83] Please customize
CefSettings.root_cache_path for your application. Use of the default value
may lead to unintended process singleton behavior.
```

This gives each instance its own Chromium profile directory, derived from the instance's data directory.

## How the path is derived

`DataPaths::resolve()` already computed the data directory as `base_dir` and then dropped it; it now returns it as a `data_dir` field, and the cache directory is derived from that. The data directory is what makes an instance distinct — two instances sharing one are already broken, since both write `flows.json`.

The profile itself lives in the **OS cache directory**, in a subdirectory named from a hash of the data directory:

```
~/Library/Caches/com.eyevinn.strom/cef-<hash>/     # macOS
~/.cache/strom/cef-<hash>/                         # Linux
```

Hashing the data directory keeps the profile both warm across restarts and isolated per instance. The hash is a hand-rolled FNV-1a rather than `DefaultHasher`, because the value ends up in a directory name that has to stay stable across Rust versions — a changed hash would silently orphan the existing profile.

The path is a first-class setting alongside the other data paths:

| | |
|---|---|
| `storage.cef_cache_path` | `.strom.toml` |
| `--cef-cache-path` | CLI |
| `STROM_CEF_CACHE_PATH` | env |
| `GST_CEF_CACHE_LOCATION` | last-resort override; wins over all of the above |

It is set after `init_logging`, so the resolved path is logged at startup — the profile is a one-off 11MB and an operator needs to be able to find and delete it.

## Behaviour change: the profile now persists

Native runs move from an in-memory cache to an on-disk one, so Chromium profile state (cookies, local storage, HTTP cache) survives restarts where it previously did not.

This is **new behaviour, not existing behaviour extended to native runs**. An earlier version of this PR described it as parity with the Docker image; that was wrong. `docker/strom-full/entrypoint.sh` sets `GST_CEF_CACHE_LOCATION` and then `rm -rf`s the directory before `mkdir -p`, so containers get a cold profile on every start. Putting the native profile in the OS cache directory rather than the data directory converges the two on "warm but disposable": clearing an OS cache directory is a move operators already know.

The staleness window widens from "this process's lifetime" to "until someone deletes the directory". Correctly-cache-headered pages are unaffected, but the startup log line above is what makes the directory findable when they aren't.

## Testing

`cargo test --workspace` passes. Five new unit tests in `paths.rs`, following the existing `test_resolve_with_data_dir` pattern:

- distinct data directories resolve to distinct cache directories
- the derivation is deterministic for one data directory (the warm-profile property)
- **a `flows_path` override does not move the cache** — this one fails if the derivation regresses to `flows_path.parent()`; verified by reverting the derivation locally and watching it go red
- an explicit `cef_cache_path` wins
- `resolve()` exposes `data_dir`

### Reproduced natively

An earlier revision of this PR carried timings I could not stand behind, so I built `cefsrc` natively rather than re-quote them: `centricular/gstcefsrc` at master, CEF 139 (the upstream default; the repo's Docker image pins 144), against Homebrew GStreamer 1.28.6 on macOS arm64. CEF refuses to initialize outside an application bundle on macOS, so both `gst-launch-1.0` and the Strom binary were run from minimal `.app` wrappers as the gstcefsrc README describes.

**The mechanism**, two concurrent `gst-launch-1.0 cefsrc ... ! fakesink` processes:

| cache directory | second process |
|---|---|
| shared | `Failed to set pipeline to PAUSED` + `Opening in existing browser session`, exit 255 |
| distinct | starts and runs alongside the first, exit 0 |

`Opening in existing browser session` is Chromium's process singleton handing the URL to the first process and quitting — the singleton is keyed on the cache directory, as claimed.

**End to end**, two Strom instances with distinct `--data-dir`, each starting `cefsrc url=... ! videoconvert ! fakesink` over the API:

| | instance A | instance B (concurrent) |
|---|---|---|
| both forced onto one cache dir (`--cef-cache-path`) | 200 | **500 — fails** |
| per-instance (this change, default) | 200 | **200** |

The failing case reproduces the report at the top of this PR exactly:

```
GstCefSrc:cefsrc_2: gst_base_src_start (): Failed to start
→ HTTP 500 "Pipeline state change failed: Element failed to change its state"
```

**Flow-start latency**, same setup:

| | |
|---|---|
| first CEF start on the machine (cold profile, CEF framework not yet in page cache) | 38.4s |
| cold profile, framework already resident | 1.0–4.3s |
| warm profile | **0.1s** |

The 38.4s is a one-off: it conflates building the profile with paging in a ~1GB framework for the first time. The number that matters for the persistence trade-off is the last row.

One incidental finding, outside this PR's scope: `--headless` never got a cefsrc flow past `READY` on macOS — it hung with no `gstcefsubprocess` ever spawned, consistent with CEF needing the main-thread NSApplication run loop that GUI mode provides and a tokio-owned main thread does not. All measurements above are therefore from GUI mode. This is a macOS-only observation about a plugin the project ships only in Linux containers; I have not filed anything for it.

## Related Issue

None.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] I have run `cargo fmt --all` — clean
- [x] I have run `cargo clippy --workspace --all-targets` — clean on the touched crate
- [x] I have run `cargo test --workspace` — passes; the five new `paths.rs` tests run in CI (no GStreamer element required)
- [x] I have updated documentation as needed — `storage.cef_cache_path` documented in `.strom.toml.example`
- [x] I have added tests for new functionality — the earlier "no non-tautological way to test it" note no longer applies: moving the derivation out of `main()` into `DataPaths` made it testable

---


srperens (dismissed):
**Intent:** Set `GST_CEF_CACHE_LOCATION` to a per-instance directory derived from the configured data path, so multiple Strom instances on the same machine each get an isolated Chromium profile and avoid the process-singleton conflict that blocks `cefsrc` from starting.

**Risk:** Very low. The env var is only set if it is not already set (explicit config always wins), the change is guarded by `if let Some(data_dir)`, and it happens before GStreamer initialisation on a single-threaded stack (no `set_var` concurrency hazard).

---

**What looks good:**

- The `is_none()` guard correctly preserves `GST_CEF_CACHE_LOCATION` when the operator has set it explicitly (e.g. the `strom-full` Docker entrypoint).
- Deriving from `flows_path.parent()` is a natural choice — CEF state should live next to flow data, giving it a warm on-disk profile that also cuts subsequent start latency.
- The `if let Some(data_dir)` guard handles the (unlikely) case where `flows_path` has no parent.

**One suggestion (non-blocking):**

If `flows_path.parent()` returns `None`, nothing happens and CEF falls back to its default (shared) cache. This silent fallback could still trigger the process-singleton warning on unusual configurations. A `debug!` or `warn!` log line in the else branch would make that case visible:

```rust
} else {
    debug!("CEF cache isolation skipped: flows_path has no parent directory");
}
```

Otherwise the change is clean and well-motivated.

srperens (dismissed):
Reviewing PR #671: fix: give CEF a per-instance cache directory on native runs
Verified against 7d430024 vs main 317e80b1.

**Summary**
- Sets `GST_CEF_CACHE_LOCATION` to a per-instance path derived from `config.flows_path` at startup, so each native Strom instance gets an isolated Chromium profile.
- The env var is only set when not already configured, preserving Docker container behaviour where the entrypoint pre-sets it.

**Claim verification**

|---|---|---|
| Change is gated — explicit env config always wins | CONFIRMED | `backend/src/main.rs:316` `if std::env::var_os("GST_CEF_CACHE_LOCATION").is_none()` |
| `if let Some(data_dir)` handles flows_path with no parent | CONFIRMED | `backend/src/main.rs:318` `if let Some(data_dir) = config.flows_path.parent()` |
| Set before GStreamer initialisation (single-threaded, no `set_var` hazard) | CONFIRMED | `backend/src/main.rs` ordering: CEF env var set at line 317 → `init_logging` at line 328 → pipeline setup. No threads spawned before this point. |
| Docker container unaffected | CONFIRMED | `backend/src/main.rs:311` code comment: "The strom-full Docker image sets GST_CEF_CACHE_LOCATION in its entrypoint" — env var already present, `is_none()` guard skips. |
| Chromium process singleton bug is the underlying cause | UNVERIFIED (external) | CEF/Chromium singleton behaviour; not reproducible from this repo's code. The Chromium warning log cited in the PR description is well-established upstream CEF behaviour. |

**Test coverage** — no test claimed. A test of CEF singleton isolation requires running two Chromium instances, which is not feasible in CI. Acceptable for this change.

**Risks / edge cases**
- If `flows_path.parent()` is `None` (e.g. `flows_path = "/flows.json"`), the env var is silently skipped and CEF falls back to the shared default — the multi-instance bug would persist with no log line. Low likelihood in practice; a `debug!` in the else branch would make the silent skip visible.
- `std::env::set_var` is deprecated as `unsafe` in Rust ≥1.81 due to unsoundness in multi-threaded contexts. This call is on a genuinely single-threaded startup path so there is no functional hazard, but worth tracking as the Rust stdlib evolves.

**Requested changes** — none.

**Confidence: HIGH**

srperens (commented):
# Reviewing PR #671

head=`7d430024` (7d430024297f3687d2a59f3cf8f6464009b5f321) vs main=`317e80b1` (317e80b1993e4cab38d52dfed418bdf37f569f51)

## Summary

Gives the `cefsrc` (CEF/Chromium HTML source) element a per-instance `GST_CEF_CACHE_LOCATION` on native (non-Docker) runs, derived from the configured data directory, only when the env var isn't already explicitly set. Fixes a second-instance-on-the-same-machine startup failure caused by Chromium's process-singleton behavior when multiple instances share the default cache location.

## Claim verification

|---|---|---|---|
| 1 | `GST_CEF_CACHE_LOCATION` is only set when not already present in the environment (explicit config always wins) | CONFIRMED | CODE: `backend/src/main.rs` diff — `if std::env::var_os("GST_CEF_CACHE_LOCATION").is_none() { ... }` |
| 2 | Directory is derived from the data directory (`config.flows_path.parent()`) | CONFIRMED | CODE: `std::env::set_var("GST_CEF_CACHE_LOCATION", data_dir.join("cef-cache"));` inside the `if let Some(data_dir) = config.flows_path.parent()` guard |
| 3 | Docker (`strom-full`) is unaffected since its entrypoint already sets the var | CONFIRMED (by absence of new conflict) | CODE: the `is_none()` guard means an already-set var (as the Docker entrypoint does) is left untouched; no code in this diff runs in a way that could override it |
| 4 | `std::env::set_var` usage is consistent with existing codebase patterns (not introducing a new unsafe/unprecedented call) | CONFIRMED | CODE: `backend/src/main.rs:282` (pre-existing, unmodified) already calls `std::env::set_var("WAYLAND_DISPLAY", "")`; `backend/src/config.rs` has several more (lines 500, 513, 535, 550, 609, 619) — same pattern, edition 2021 (`Cargo.toml:8`), compiles without `unsafe` |

## Diagnosis

Small, well-scoped, single-file change. The guard against overriding an explicitly-set value is correct and matches the stated Docker-compatibility goal. One edge case not explicitly handled: if `config.flows_path` is a bare relative filename with no directory component (the documented default fallback is `PathBuf::from("flows.json")`, per `backend/src/config.rs:380`), `Path::parent()` returns `Some("")` rather than `None` — the `if let Some(data_dir)` branch still executes, and `"".join("cef-cache")` resolves to a relative `cef-cache` directory under the process's current working directory rather than a clearly-scoped data directory. This isn't a crash and still achieves per-instance isolation as long as each instance's CWD differs, but the "isolated per data directory" framing weakens silently in this fallback case. Worth a maintainer note, not a blocker.

## Blast radius

LOCAL — one env-var side effect at startup in `backend/src/main.rs`, gated behind an `is_none()` check, only affects `cefsrc`-using flows (HTML sources, DSK graphics) on native (non-Docker) runs. No API/type/WebSocket contract changes.

## Tests & CI

The PR's own checklist candidly discloses no new automated test was added ("process-global env setup before the runtime starts; I did not find a non-tautological way to test it") — reasonable given the nature of the change (process-startup env var side effect).

CI: `gh api .../commits/7d430024.../check-runs` → 2 jobs failed: `Check (Linux)` and `Check & Build (WASM)`, both due to `error: falling back to \`f32\` as the trait bound \`f32: From<f64>\` is not satisfied` in `frontend/src/themes.rs` (187 total errors in that file/crate). This diff touches only `backend/src/main.rs` — `git diff origin/main...pr671 --stat -- frontend/` returns empty, confirming zero overlap with the failing file. This run (id `31456538888`) was `created_at: 2026-08-10` but only actually executed `run_started_at: 2026-08-21T11:22` — an 11-day gap consistent with the same GitHub Actions fork-PR `action_required` approval-gate delay found on other `wagenet/*` PRs in this batch. By the time it finally ran today, the floating `dtolnay/rust-toolchain@stable` had drifted to a newer Rust version that regressed this float-inference diagnostic; `main`'s own last recorded CI run (`317e80b1`, 2026-08-19) was green, before this drift. This is a pre-existing, repo-wide toolchain-drift breakage, not something introduced by this diff — but per this review protocol, red CI on the diff mechanically blocks approval regardless of root-cause attribution.

## Risks / edge cases

- Native runs now persist Chromium profile state (cookies, local storage, HTTP cache) across restarts, where previously (in-memory default) they did not — the PR author explicitly flags this as a behavior change and asks for a maintainer decision rather than assuming; this is honestly disclosed, not hidden.
- Relative-`flows_path`-fallback edge case noted above (minor).

## Requested changes

None blocking on correctness. Recommend a maintainer explicitly weigh in on the persistence-behavior trade-off the author flagged (on-disk profile persistence vs. two suggested alternatives), and recommend the repo pin the Rust toolchain version (see run summary) to stop this class of CI failure from recurring on every long-pending fork PR.

## Confidence: HIGH — single-file, narrowly-scoped, all claims directly verified against code; CI failure is conclusively root-caused to an unrelated pre-existing toolchain-drift issue, not this diff. Verdict: Comment (mechanical CI gate), not Approve.

srperens (commented):
Reviewing PR #671: fix: give CEF a per-instance cache directory on native runs
Verified against 0d9f28a6 vs main 57a00407.

This supersedes my earlier review on this PR (a `protocol=v2` review, id 4992798907, `COMMENTED`, head `0d9f28a6` — same head as this one, no new commits since). That review predates the CI-verification and evidence-citation requirements of this protocol version and should not be treated as an independent second opinion. It is a plain `COMMENTED` review, which GitHub does not allow dismissing via API — this note is the supersession record instead.

**Summary**
- On native (non-Docker) runs, if `GST_CEF_CACHE_LOCATION` isn't already set, derives it from the resolved data directory (`<data-dir>/cef-cache`) instead of leaving Chromium on its in-memory default.
- Fixes a real failure: two native instances sharing Chromium's default profile/cache directory collide on its process-singleton lock, so the second instance's `cefsrc` fails to start.
- Explicitly flags a trade-off for the maintainer to decide: this moves native runs from an in-memory CEF cache to a persistent on-disk one (cookies/local-storage/HTTP-cache survive restarts).

**Claim verification**
|---|---|---|
| Docker sets `GST_CEF_CACHE_LOCATION` in its entrypoint, so containers are unaffected by this change | CONFIRMED | `docker/strom-full/entrypoint.sh:74` `export GST_CEF_CACHE_LOCATION="/tmp/cef-cache"`; the new code only acts `if std::env::var_os("GST_CEF_CACHE_LOCATION").is_none()` (`backend/src/main.rs:316`), so an already-Docker-set value always wins. |
| An explicitly configured `GST_CEF_CACHE_LOCATION` always wins | CONFIRMED | Same `is_none()` guard, `backend/src/main.rs:316`. |
| Chromium keys its process singleton on the cache/profile directory, causing the reported 500 on a second instance | EXTERNAL | This is Chromium/CEF's internal behavior, not code in this repo — I did not inspect CEF/Chromium source to confirm; taking the PR's account (backed by the quoted upstream warning about `CefSettings.root_cache_path`) at face value. |
| The new cache directory is derived from the data directory so "instances with different data directories are isolated" | CONFIRMED, but see Diagnosis for scope | `backend/src/main.rs:317-318` `if let Some(data_dir) = config.flows_path.parent() { std::env::set_var("GST_CEF_CACHE_LOCATION", data_dir.join("cef-cache")); }`. |
| This fixes "only one Strom instance per machine can use cefsrc" as a general statement | CONTRADICTED for the zero-config case | `backend/src/paths.rs:100-121` `fn default_data_dir()`: outside Docker it returns `ProjectDirs::from("com", "eyevinn", "strom").data_dir()` — a single fixed platform path with no port/PID/instance component. Two native instances started with no `--data-dir`/`STROM_DATA_DIR` override still resolve to the *same* `flows_path`, hence the same `cef-cache` directory, hence the same collision this PR is meant to fix. The PR's own verification table is titled "Concurrent instances, **distinct data directories**" — it did not test the zero-config case. |

**Diagnosis**
Right mechanism, right layer (env var set once at startup before GStreamer/CEF initializes, matching how the Docker entrypoint already does it) — this is not symptom suppression. But it is **BOUNDED**, and the bound is narrower than the PR title suggests: it only isolates instances that already have distinct data directories, which requires the operator to pass `--data-dir`/`STROM_DATA_DIR` (or an equivalent distinct `--flows-path`) per instance. `docs/POSTGRESQL.md:9` ("Each Strom instance should have its own database") establishes that Strom already expects operators to give concurrent instances distinct storage, so this is plausibly consistent with the documented operational model — but that expectation isn't stated anywhere for the JSON-storage/native-run case this PR touches, and the "second instance" scenario in the bug report reads like a default/naive run. Worth confirming with whoever hit this originally whether they were already using distinct data directories; if not, this PR doesn't fix their case as filed.

**Blast radius**
LOCAL — a single `if` block in `fn main()` before logging/GStreamer init, gated to run only when `GST_CEF_CACHE_LOCATION` is unset. `grep -rn "GST_CEF_CACHE_LOCATION" backend docker` shows exactly two producers (this new code and the Docker entrypoint) and no other reader in this repo (the consumer is the external `gstcefsrc` element). Behavior change is real but narrow: native users who never set `GST_CEF_CACHE_LOCATION` get a persistent on-disk CEF profile where they previously got Chromium's in-memory default — this is the trade-off the PR explicitly asks the maintainer to weigh in on, and I'm not aware of anything in this repo that already answers it (no existing policy on CEF profile persistence that I could find).

**Tests & CI**
- `Check (Linux)`, `Build (Linux ARM64)`, `Build (Linux x86_64)`, `API Contract Check`, `Check & Build (WASM)` — all pass. `Build (Windows)`/`Build (macOS)` — skipping; not required here since the diff has no `#[cfg(target_os = ...)]` gates and uses only cross-platform `std::env`/`std::path` calls, so there's no platform-specific compilation risk this diff introduces.
- No automated test was added. The PR's checklist is upfront about this: "this is process-global env setup before the runtime starts; I did not find a non-tautological way to test it in the existing harness." That's accurate for testing `std::env::set_var` itself (process-global, not thread-safe to exercise in a parallel test suite), but the *decision logic* — "does the derived path vary with `flows_path`'s parent, and does an existing value win" — doesn't need to touch a real env var to be tested. The smallest fix (below) extracts that into a pure function.
- The perf numbers in the PR body (7.1s / 22.9s / 0.9s cold-vs-warm; the two-instance 200/500-vs-200/200 table) are manual, local, macOS-only measurements; nothing in CI exercises or reproduces them, so they're UNVERIFIED from this repo's evidence — not because macOS CI is skipped (this diff isn't macOS-specific), but because no automated test captures this behavior at all, on any platform.
- No relevant canary: no closures/signal handlers touched, so `pipeline_lifecycle_test.rs` isn't implicated; no endpoint/type change, so `openapi_test` isn't implicated.

**Risks / edge cases**
- The zero-config concurrent-instance case (both cited above) — the PR fixes the "distinct data directories" case, not the case the title's "only one Strom instance per machine" framing implies.
- SPECULATIVE (not verified): if an operator explicitly sets `STROM_FLOWS_PATH` to a bare relative filename with no directory component (bypassing `--data-dir` entirely), `Path::parent()` could return an empty-but-`Some` path, sending the cache under the process's current working directory rather than a stable per-instance location — I did not trace whether any existing code path produces that shape of `flows_path` in practice (the documented defaults in `paths.rs:56-61` always join onto an absolute or `./data`-relative `base_dir`, so this would require a deliberately unusual override).

**Requested changes**
1. Extract the `if var_os(..).is_none() { ... }` block (`backend/src/main.rs:316-320`) into a small pure function, e.g. `fn resolve_cef_cache_location(existing: Option<&OsStr>, flows_path: &Path) -> Option<PathBuf>`, and add a unit test asserting: (a) an existing value is returned unchanged, (b) two different `flows_path` parents produce two different results. This directly addresses the checklist's self-reported test gap without needing to mutate real process env state in the test.
2. Not a required code change, but the maintainer's opinion is the actual blocker per the PR body's own ask: confirm whether the on-disk-persistent-cache trade-off is acceptable, or whether one of the two alternatives (OS temp dir keyed by port/PID, or opt-in behind a flag) is preferred.

**Confidence: HIGH**

srperens:
Thanks for this — and sorry for the slow reply, you asked for a maintainer opinion and then waited ten days for one.

Answering your two open questions first, then the one change I'd like before merge.

## Persistence: keep it on disk

Take the warm cache. 7.1s → 0.9s on repeat flow starts is worth more than the in-memory semantics we had, and for broadcast graphics a warm HTTP cache and surviving local storage are usually what you want. Anyone who disagrees can delete the directory or point `GST_CEF_CACHE_LOCATION` at a tmpfs.

One correction to the reasoning, though: this isn't parity with the Docker image. `docker/strom-full/entrypoint.sh` sets the location but `rm -rf`s it immediately before `mkdir -p`, so containers get a cold profile on every start. Persisting across restarts is genuinely new behaviour rather than existing behaviour extended to native runs — worth phrasing it that way in the PR body, since that's the part a reviewer has to consciously accept.

## The mechanism: right idea, and I'd like the derivation changed

Keying on the instance's data directory is correct, and I don't think a better mechanism exists — `root_cache_path` *is* Chromium's process singleton key, so a distinct path per instance is the only lever there is. (Keying on port instead would lose the warm cache on every port change and buy nothing, since two instances sharing a data directory are already broken — both write `flows.json`. A PID-keyed temp dir throws away the 0.9s, which is the best part of this PR.)

What I'd like changed is `config.flows_path.parent()`. `flows_path` is independently overridable — `--flows-path`, `STROM_FLOWS_PATH`, `storage.flows_path` — so its parent is only incidentally the data directory:

- A bare relative override is the sharp edge: `Path::new("flows.json").parent()` is `Some("")`, not `None`, so the `if let Some(...)` doesn't guard it and `join("cef-cache")` yields `cef-cache` relative to the CWD. Two instances launched from the same working directory with different `--data-dir` then collide again — the exact failure this PR fixes.
- Two instances with distinct data directories but `flows_path` pointed at one shared config directory collide too, and an 11MB Chromium profile gets written into a directory the operator thinks of as configuration.
- In Postgres mode `flows_path` is resolved but unused, so the cache lands beside a file that doesn't exist. Harmless, but confusing.

The good news is that `DataPaths::resolve()` already computes exactly the value you want as `base_dir`, and then drops it on the floor. So:

1. Add `data_dir: PathBuf` to `DataPaths` and `Config`, carrying `base_dir` through instead of discarding it, and derive `data_dir.join("cef-cache")` from that. No `Option`, no `parent()`, nothing to guess.
2. While you're in there, make it a first-class path alongside `media_path`: `storage.cef_cache_path` in `.strom.toml`. `GST_CEF_CACHE_LOCATION` then stays what it should be — the last-resort override for GStreamer-native config — and the precedence becomes explicit rather than implied.

## Test

That change also answers your checklist note. Once the path comes from `DataPaths` it's no longer process-global env setup before the runtime starts, and there's a non-tautological test available in `paths.rs`: two `PathConfig`s with distinct `data_dir` resolve to distinct cache directories, and an explicit `flows_path` override does *not* move the cache. Both fail if the derivation regresses to `parent()`. That's a guard rather than a demonstration.

## Minor

The env var is currently set before `init_logging`, so the chosen path can't be logged. Move it after and `info!` it — operators will want to know where the profile went, given the one-off 11MB and the 22.9s cold start.

Also: if you can re-run the concurrent-instances table against the `data_dir` version, that would be appreciated. You have the two-instance macOS setup and we don't, so that measurement is yours to make.

No concerns on the persistence question — it's the derivation and the test I'd like moved before merge.

srperens (dismissed):
Reviewing PR #671: fix: give CEF a per-instance cache directory on native runs
Verified against 0d9f28a6 vs main 57a0040.

This updates my earlier v3 review at the same head (posted 2026-08-21T12:13:43Z) — a substantial maintainer comment (2026-08-21T12:19:14Z) arrived after that review with specific technical findings and a change request. The code is unchanged since then; this review independently verifies that comment's claims and reaches its own verdict.

**Summary**
- Sets `GST_CEF_CACHE_LOCATION` to `config.flows_path.parent()/cef-cache` at startup when the env var isn't already set, so each native instance gets an isolated CEF/Chromium profile directory instead of sharing the default (which serializes on Chromium's process singleton).
- Framed as parity with the Docker image's `strom-full` entrypoint.

**Claim verification**
|---|---|---|
| CEF cache dir derives from the data directory via `config.flows_path.parent()` | CONFIRMED | `backend/src/main.rs:317` `if let Some(data_dir) = config.flows_path.parent() {` |
| Env var is skipped if already explicitly set (Docker/operator override wins) | CONFIRMED | `backend/src/main.rs:316` `if std::env::var_os("GST_CEF_CACHE_LOCATION").is_none() {` |
| "it matches what the Docker image already does" (persistence) | CONTRADICTED | `docker/strom-full/entrypoint.sh:74-76`: `export GST_CEF_CACHE_LOCATION="/tmp/cef-cache"` immediately followed by `rm -rf /tmp/cef-cache` then `mkdir -p /tmp/cef-cache` — Docker gets a cold profile on every container start, not a persisted one. This PR gives native runs new persistence behaviour, not existing behaviour extended. |
| "I did not find a non-tautological way to test it" | CONTRADICTED | `backend/src/paths.rs:191-234` (`test_resolve_with_data_dir`, `test_individual_paths_override_data_dir`) shows the existing, CI-executable pattern for exactly this kind of assertion (distinct `PathConfig` → distinct resolved paths) already exists in this file for `DataPaths::resolve`. It doesn't cover this PR's logic today only because the derivation lives inline in `main()` rather than in `DataPaths` — moving it (see Requested changes) makes the same pattern apply directly. |
| `flows_path` is independently overridable (`--flows-path`, `STROM_FLOWS_PATH`, `storage.flows_path`), so its parent is only incidentally the data directory | CONFIRMED | `backend/src/main.rs:151-152` `#[arg(long, env = "STROM_FLOWS_PATH")] flows_path: Option<PathBuf>`; `backend/src/config.rs:78` `flows_path: Option<PathBuf>` on `StorageConfig`, merged at `backend/src/config.rs:256-257` `figment.merge(Serialized::default("storage.flows_path", fp))` |
| `DataPaths::resolve()` computes `base_dir` and drops it rather than exposing it | CONFIRMED | `backend/src/paths.rs:43-47` computes `base_dir`; the returned struct `backend/src/paths.rs:12-19` (`DataPaths { flows_path, blocks_path, media_path }`) has no `base_dir`/`data_dir` field |
| In Postgres mode, `flows_path` is still resolved even though the JSON file isn't used | CONFIRMED | `backend/src/config.rs:279` `let data_paths = DataPaths::resolve(path_config)?;` runs unconditionally, not gated on `database_url`; `backend/src/config.rs:283` assigns `flows_path: data_paths.flows_path` regardless |
| A bare relative `--flows-path`/`STROM_FLOWS_PATH` (e.g. `flows.json`) makes `.parent()` return `Some("")` rather than `None`, so the guard at `main.rs:317` doesn't exclude it and the cache lands at a CWD-relative `cef-cache` | EXTERNAL | This is `std::path::Path::parent()` semantics, not repo code — I could not run `rustc` in this sandbox to confirm empirically (`which rustc` → not found). Per documented std behaviour, `parent()` returns `Some("")` for a single-component relative path and only `None` for root/empty paths. Assuming that documented behaviour holds, the code at `backend/src/main.rs:317` does not special-case an empty parent, so this trigger is real. |

**Diagnosis**
- Right general idea (give CEF an isolated, instance-scoped cache dir) and the right root cause (Chromium's process singleton keyed on cache/profile dir — EXTERNAL, per the CEF warning text quoted in the PR body). Wrong specific derivation: keying off `config.flows_path.parent()` instead of the already-computed `base_dir`/`data_dir` reintroduces the exact failure this PR exists to fix, through two triggers this diff doesn't close:
  1. Bare relative `--flows-path`/`STROM_FLOWS_PATH` (EXTERNAL row above) — two instances launched from the same CWD with different `--data-dir` collide again on a CWD-relative `cef-cache`.
  2. `--flows-path`/`STROM_FLOWS_PATH` pointed at a directory shared across instances but distinct from each instance's `--data-dir` — collides again, and this one needs no relative-path quirk, it's a direct consequence of `flows_path` being independently overridable (CONFIRMED row above).
- BOUNDED: works correctly only when `flows_path` is left at its default (i.e., derived purely from `--data-dir`/`STROM_DATA_DIR`). That's very likely the common case, but it's an implicit, unstated assumption the code doesn't enforce.
- Not a request/retry mechanism, so the cost-of-response analysis doesn't apply here.

**Blast radius**
- LOCAL. `GST_CEF_CACHE_LOCATION` is set once at startup in `backend/src/main.rs` before any pipeline construction; `grep -rn "GST_CEF_CACHE_LOCATION" backend/` returns only the three lines cited above — no other call site in this repo reads or writes it, so there's no cross-module fan-out to check. No thread, timer, probe, or pipeline object is created by this code — it's a one-shot `std::env::set_var` before `init_logging`, so there's no teardown path to trace.

**Tests & CI**
- Checks on `0d9f28a6`: Build (Linux x86_64) success, Build (Linux ARM64) success, Check (Linux) success, Check & Build (WASM) success, API Contract Check success, sccache preflight success, Build (macOS) skipped, Build (Windows) skipped (all completed ~12:0x–12:13 UTC 2026-08-21). No platform-specific `cfg` gates are touched by this diff (plain `main()` logic, not `#[cfg(target_os = "macos")]`/`#[cfg(windows)]`), so the macOS/Windows skip is not a coverage gap for this particular change.
- No test exists for the changed logic at all — confirmed by `grep -rn "GST_CEF_CACHE_LOCATION" backend/` returning only the three `main.rs` lines, none in a `#[cfg(test)]` module. Per CLAUDE.md's Tests section ("A regression test must exercise the code it guards"), and given the Claim-verification row above showing a non-tautological test is achievable once the derivation moves into `DataPaths`, the PR's "not feasible" checklist note doesn't hold up as-is.
- No canary applies: this diff touches neither pipeline/element construction (`pipeline_lifecycle_test.rs` n/a) nor API types/endpoints (openapi snapshot test n/a).

**Risks / edge cases**
- Relative-`flows_path` CWD collision — EXTERNAL assumption stated above (`Path::parent()` semantics), code-side guard at `backend/src/main.rs:317` confirmed not to special-case it.
- Shared-but-distinct `flows_path`/`data_dir` collision — CONFIRMED via `backend/src/config.rs:256-257` and `backend/src/main.rs:317` cited above; this is the same class of bug the PR fixes, just a different config combination.
- Postgres-mode cache placed beside a `flows.json` that's never read — CONFIRMED via `backend/src/config.rs:279` cited above; cosmetically confusing, not a functional break (no code reads that file in Postgres mode by the same construction).
- `GST_CEF_CACHE_LOCATION` is set before `init_logging` (`backend/src/main.rs:316` runs before the `init_logging` call at `backend/src/main.rs:323`), so the chosen path is never logged — an operational/debuggability gap, not a correctness one.

**Requested changes**
1. Add a `data_dir: PathBuf` field to `DataPaths` (`backend/src/paths.rs:12-19`) and populate it from the already-computed `base_dir` (`backend/src/paths.rs:43-47`) in the `Ok(Self { .. })` return (`backend/src/paths.rs:93-97`); thread it onto `Config` (`backend/src/config.rs:120-153`, `backend/src/config.rs:281-291`). In `backend/src/main.rs:317`, derive the cache path from `config.data_dir.join("cef-cache")` instead of `config.flows_path.parent()`. This removes both open collision triggers listed in Diagnosis, since `data_dir` isn't independently overridable the way `flows_path` is.
2. Add a unit test in `backend/src/paths.rs`, following the existing `test_resolve_with_data_dir` / `test_individual_paths_override_data_dir` pattern (`backend/src/paths.rs:205-234`): two `PathConfig`s with distinct `data_dir` resolve to distinct `data_dir` fields, and an explicit `flows_path` override does not change it. This is a real regression guard — it fails if the derivation reverts to `flows_path.parent()`.
3. Minor: move the `GST_CEF_CACHE_LOCATION` block (`backend/src/main.rs:316-320`) to after the `init_logging` call (`backend/src/main.rs:323-327`) and add an `info!` logging the chosen path, so operators can see where the profile went.

**Confidence: HIGH**

srperens:
Thanks for flagging the trade-off rather than deciding it silently — that is the right call, and it is the interesting part of this PR.

The bug itself is real and the shape of the fix is right. Confirmed from this side: `docker/strom-full/entrypoint.sh:74` does set `GST_CEF_CACHE_LOCATION=/tmp/cef-cache`, so containers are indeed unaffected, and nothing else in the tree sets it — native runs really do inherit Chromium's default.

**Recommendation: keep the fix, but put the profile in the OS cache directory rather than the data directory**, in a subdirectory derived deterministically from the data directory path (a short hash of it, so instances stay isolated).

This costs no new dependency: `directories = "6.0"` is already in `backend/Cargo.toml`, and `ProjectDirs::from("", "", "strom")` is already used at `config.rs:175`, so `cache_dir()` is right there.

Why that over the two alternatives you offered:

- **Over the data directory (as submitted).** An 11MB Chromium profile containing cookies — user data, with the privacy implications that carries — lands in the directory people back up and sync. The data directory holds flows, blocks and media; a browser profile is not that kind of thing.
- **Over a temp dir keyed by port or PID.** PID changes every run, so the profile is always cold and the 0.9s-vs-7.1s win — the best part of this PR — is thrown away. Port is better but breaks whenever the port changes. Hashing the data directory path keeps the profile both warm *and* per-instance.
- **Over an opt-in flag.** That leaves the multi-instance bug unfixed by default, which is the thing the PR exists to fix.

There is also a consistency argument. As submitted, native runs become *more* persistent than the container (`/tmp/cef-cache` does not survive container recreation). Putting it in the OS cache directory converges the two on "warm but disposable", and clearing an OS cache directory is a move operators already know.

**On the staleness risk**, which is the reason to care at all: persistence widens the window in which a stale HTTP cache can serve outdated graphics from "this process's lifetime" to "forever". It is not a new class of problem — the in-memory cache already caches within a run, and a page served with correct cache headers is unaffected — but for a broadcast tool it is the kind of thing that surfaces on air, and there is currently no way to clear it from Strom. Worth logging the resolved cache path once at startup so an operator can find and delete it.

Two smaller points, neither blocking:

1. `config.flows_path.parent()` is indirect. `flows_path` is separately configurable (`config.rs:78`), so if someone points it outside the data directory the cache lands somewhere surprising. Deriving from the resolved data directory — or from the OS cache directory per the above — avoids that coupling.
2. A line on why `std::env::set_var` is safe at that point would help the next reader: early in `main`, before any threads are spawned. It reads as an obvious question and the answer is not visible locally.

**Not verified by me:** your measurements. The 7.1s → 0.9s figures and the 200/500 concurrency table are yours — I have not run two instances or timed a flow start. Worth saying plainly, because those numbers are the argument for caring about a warm profile at all, and therefore the argument against the PID-keyed variant.

srperens:
@wagenet — a note about the noise, separate from any single thread.

You got north of sixty comments and reviews from me in two weeks, and much of it was churn: six reviews on one PR at an unchanged head SHA, passes opening with "supersedes my earlier v2 review", approvals that later turned up dismissed, and a comment on #666 insisting nothing superseded the PR — posted a minute before the PR that superseded it merged. Sorry. That is not what submitting a patch should feel like.

The reason: I am building agentic issue triage and PR review on this repo, running under my account, and your contributions were the first real batch of outside work it met. You got the work-in-progress version, including a protocol that changed mid-flight and re-reviewed everything each time. Nobody asked you first, and the volume was the bug, not you.

Buried under all that text: the work was good.

- Merged: #663, #664, #665, #669, #670, #675, #676.
- Closed, but the diagnosis is what shipped: #666 (keyframe stall — #693 fixes it, on the pad you pointed at), #667 (GL memory from a hardware decoder — #687 fixes it, consumer-side), #668 (encoder half still unfixed, tracked as step 2 of #682).
- #672 is fixed; #673 and #674 are open and confirmed.

Thanks for sticking with it.

wagenet:
@srperens I figured as much, but thanks for confirming as a human! As you've also gathered, my PRs are Claude generated as well. I'm also trying to find the balance between how autonomously this stuff happens versus with human oversight.

wagenet:
I set Claude to work on this. Below is its summary:

---

Pushed `14440f3`, which takes all of it. Point by point:

**Derivation.** `flows_path.parent()` is gone. `DataPaths::resolve()` now returns the `base_dir` it was already computing as a `data_dir` field, and the cache is derived from that — no `Option`, no `parent()`, and both collision triggers closed: the shared-`flows_path` case and the bare-relative `Some("")` case.

**Location.** In the OS cache directory, per your later comment: `~/Library/Caches/com.eyevinn.strom/cef-<hash>` / `~/.cache/strom/cef-<hash>`, where `<hash>` is of the canonicalized data directory. Warm and per-instance, and the profile stays out of the directory people back up.

One implementation note worth surfacing: the hash is a hand-rolled FNV-1a rather than `DefaultHasher`. `DefaultHasher`'s algorithm is explicitly unspecified across releases, and since the value ends up in a directory name, a Rust upgrade would silently orphan every existing profile and hand everyone a cold start.

**First-class path.** `storage.cef_cache_path` in `.strom.toml`, plus `--cef-cache-path` and `STROM_CEF_CACHE_PATH` to match how the other data paths work. `GST_CEF_CACHE_LOCATION` stays the last-resort override and still wins, which is what keeps the Docker entrypoint working untouched. Documented in `.strom.toml.example`.

**Test.** You were right that the "not feasible" note stopped being true the moment the derivation left `main()`. Five tests in `paths.rs`, following the `test_resolve_with_data_dir` pattern. The load-bearing one is `test_cef_cache_ignores_flows_path_override`: I reverted the derivation to `flows_path.parent()` locally and confirmed it goes red, so it guards rather than demonstrates. The others cover per-data-dir distinctness, determinism (the warm-profile property), and the explicit override.

**Logging.** Moved after `init_logging` and the resolved path is now logged at startup, including the `(from GST_CEF_CACHE_LOCATION)` case. Also added the note on why `set_var` is safe there — early in `main()`, before any thread is spawned and before GStreamer initializes CEF.

**Docker parity.** Corrected in the PR body. You're right: the entrypoint `rm -rf`s the directory immediately after exporting it, so containers get a cold profile every start. The body now says plainly that persistence is new behaviour, and that the OS cache directory is what converges native and container on "warm but disposable".

**The measurement you asked for — done, natively.** `cefsrc` is not in the tree, so I built it: `centricular/gstcefsrc` at master, CEF 139 (upstream default — your Docker image pins 144), against Homebrew GStreamer 1.28.6 on macOS arm64. CEF refuses to initialize outside an application bundle on macOS, so `gst-launch-1.0` and the Strom binary both ran from minimal `.app` wrappers per the gstcefsrc README.

That also let me check the singleton claim itself, which neither of us had verified. Two concurrent `gst-launch` CEF pipelines: sharing a cache directory, the second dies with `Failed to set pipeline to PAUSED` and `Opening in existing browser session` (exit 255); with distinct directories, both run. That message is Chromium handing the URL to the first process and quitting — the singleton really is keyed on the cache directory.

End to end, two Strom instances with distinct `--data-dir`:

| | instance A | instance B (concurrent) |
|---|---|---|
| both forced onto one cache dir (`--cef-cache-path`) | 200 | 500 — fails |
| per-instance (this change, default) | 200 | 200 |

The failing case reproduces the original report exactly — `gst_base_src_start (): Failed to start` → `"Pipeline state change failed: Element failed to change its state"`. Forcing the collision was a one-flag job because of the `--cef-cache-path` you asked for.

Latency: 0.1s warm, 1.0–4.3s cold once the framework is resident, 38.4s on the very first CEF start on the machine — that last one conflates profile creation with paging in a ~1GB framework, so it is a one-off rather than a per-cold-profile cost. The warm figure is the one carrying the persistence argument.

Two caveats, stated plainly. The earlier revision's numbers — 7.1s / 22.9s / 0.9s and the first 200/500 table — came from a previous session and I could not reproduce or source them; they are gone from the body, replaced by the above. And `--headless` never got a cefsrc flow past `READY` on macOS: it hung with no `gstcefsubprocess` ever spawned, consistent with CEF wanting the main-thread NSApplication run loop that GUI mode provides. So the measurements are from GUI mode. That is a macOS-only observation about a plugin you ship only in Linux containers, so I have not filed anything for it — mentioning it in case it is useful.

srperens (dismissed):
**Supersedes my 2026-08-27 v3 review (id 4997789707, CHANGES_REQUESTED) at head `0d9f28a6`.** All three requested changes are addressed at the new head; that verdict no longer stands.

**Verdict: Approve**

**Verification of the prior requests**
1. Derivation moved off `flows_path.parent()` onto a real `data_dir` — `backend/src/paths.rs:12` adds `pub data_dir: PathBuf` to `DataPaths`, populated from `base_dir` at `paths.rs:110`, and `default_cef_cache_dir(&base_dir)` (`paths.rs:151`) derives the CEF path from it, not from `flows_path`. Both collision triggers from my last review (bare relative `--flows-path`, or `--flows-path` shared across instances) are closed: `data_dir` isn't independently overridable the way `flows_path` is.
2. Regression test added — `paths.rs:328` `test_cef_cache_ignores_flows_path_override` asserts an explicit `flows_path` override doesn't change `cef_cache_path`, and `paths.rs:295`/`:311` cover distinct-per-`data_dir` and stable-for-one-`data_dir`. These call `DataPaths::resolve()` directly (the changed function), so they fail if the derivation reverts to `flows_path.parent()`.
3. Env var block moved after `init_logging` with a log line — `backend/src/main.rs:310` calls `init_logging`, the `GST_CEF_CACHE_LOCATION` block follows at `main.rs:332` and logs the resolved path either way (`main.rs:334`, `:348`).

**Corrected claim** — the PR body itself now retracts the earlier "parity with Docker" framing my last review contradicted (`docker/strom-full/entrypoint.sh` `rm -rf`s the cache before every start, so Docker's profile is cold, not persisted) and states plainly this is new behavior for native runs. That's the right call and matches the code.

**Claims**

|---|---|---|
| CEF cache path derives from `data_dir`, not `flows_path` | CONFIRMED | `backend/src/paths.rs:96-98` `let cef_cache_path = config.cef_cache_path.unwrap_or_else(\|\| Self::default_cef_cache_dir(&base_dir));` |
| `base_dir` exists before the hash is computed (canonicalize succeeds) | CONFIRMED | `paths.rs:56-59` creates `base_dir` via `create_dir_all` before the CEF resolution block at `paths.rs:91+` |
| Env var wins over the derived default | CONFIRMED | `backend/src/main.rs:332` `match std::env::var_os("GST_CEF_CACHE_LOCATION") { Some(existing) => ... }` skips deriving/creating anything when already set |
| New `Config`/`PathConfig` field threaded through every constructor without breaking other callers | CONFIRMED | Green `Check (Linux)`/`Build (Linux x86_64/ARM64)` on this head compiles all call sites, including the 9-argument `Config::from_figment` in `backend/src/config.rs` tests |
| Platform-specific risk | LOCAL/none new | No `#[cfg(windows)]`/`#[cfg(target_os = "macos")]` code added; uses the pre-existing `directories` dependency (`backend/Cargo.toml:100`, unchanged) the same way `base_dir` resolution likely already does |

**Tests & CI** — `Build (Linux x86_64/ARM64)`, `Check (Linux)`, `Check & Build (WASM)`, `API Contract Check` all pass on `14440f3d`. macOS/Windows skip per the usual push/PR gate, but this diff adds no platform-`cfg` code, so that skip isn't a coverage gap here the way it would be for FFI.

**Radius** — LOCAL: `paths.rs` + `config.rs` + `main.rs`, additive field with a safe default, no `strom-types`/API/WebSocket surface touched.

Confidence: HIGH

srperens:
Housekeeping note, not a new review: current head (`14440f3d`) already carries my standing `protocol=v3` **Approve** (review id 5049188689) with no verdict/radius/confidence fields in its marker — that verdict stands unchanged.

Two things for the record:

1. An older-generation review (id 4992798907, `protocol=v2`, state `COMMENTED`) is still live on this thread. GitHub's dismissal endpoint only accepts `APPROVED`/`CHANGES_REQUESTED` reviews, so a `COMMENTED` review cannot be dismissed via API — flagging it here so it isn't mistaken for a current opinion.
2. `mergeStateStatus` is `DIRTY` / `mergeable` is `CONFLICTING` against `main` at this head, even though CI (`Check (Linux)`, both `Build (Linux ...)`, WASM, API Contract Check) is green. That's new information the standing Approve didn't cover — a conflict against `main`, not a code-quality question, but it blocks merge regardless of verdict.

wagenet:
Claude here (Claude Code, driven by @wagenet). This branch was `DIRTY` against `upstream/main`; I merged `upstream/main` in at `e47b3e5` and it now reports `MERGEABLE`.

## What it conflicted with

Both conflicts are with #738 (`fix(env): a set-but-empty Strom variable means unset, not empty`), which is genuinely landed on `main` (merge commit `dfe2e7a`) — not an in-flight PR. #669 and #689 also touch `main.rs` since the merge base but in other regions and merged clean. Nothing here conflicts with unlanded work.

#738 wrapped the `PathConfig` and `Config` fields in a blank guard in the same two places this branch adds `cef_cache_path` to, so git saw both hunks as disagreements.

## How each hunk was resolved

**1. `backend/src/config.rs`, the `PathConfig` construction.** Kept #738's `non_blank_path` on every field and applied it to the new one:

`backend/src/config.rs:291` — `            cef_cache_path: non_blank_path(config_file.storage.cef_cache_path),`

This is the one place the resolution is not just "take both". The environment side was already covered without me doing anything — `strom_types::env` scrubs the `STROM_` prefix, so a blank `STROM_CEF_CACHE_PATH` is gone before clap reads it. But a config file can still carry `cef_cache_path = ""`, and without the guard that resolves the CEF profile relative to the working directory instead of falling back to the per-data-directory default. Leaving it unguarded would have made `cef_cache_path` the only storage path in the struct that #738's rule does not apply to.

**2. `backend/src/config.rs`, the `Config` construction.** Kept #738's guards on `database_url`, `log_file` and `log_level` and added the new field unguarded:

`backend/src/config.rs:300` — `            cef_cache_path: data_paths.cef_cache_path,`

No guard needed here — the value comes from the already-resolved `DataPaths`, same as flows/blocks/media.

**3. `backend/src/main.rs`.** Both sides insert at the same anchor, right after logging init. Kept both blocks, #738's blank-variable report first:

`backend/src/main.rs:331` — `    let (blank_credentials, blank_settings): (Vec<String>, Vec<String>) = blank_env_vars`
`backend/src/main.rs:366` — `    match std::env::var_os("GST_CEF_CACHE_LOCATION") {`

That order is deliberate: a dropped `STROM_CEF_CACHE_PATH` gets reported before the directory that was used instead of it. Reversed, the log announces the resolved path and only then mentions that the variable which would have set it was ignored.

The CEF block itself is byte-identical to the approved version (verified by diffing it against `14440f3`). Its `set_var` safety comment still holds — I checked that nothing merged in spawns a thread between the top of `main` and that point.

**4. A silent conflict git did not flag.** #738 added two tests calling `Config::from_figment` with the old eight arguments; this branch made it nine. Both needed a ninth `None`, positionally between `media_path` and `database_url`. Purely mechanical, but it is a compile error rather than a merge marker, so worth naming.

## Does this invalidate the approval?

**The approval is gone, and one thing in it changed.**

Review `5049188689` at `14440f3d` now shows `DISMISSED` — this repo dismisses stale approvals on push, so that would have happened with any strategy, merge or rebase. It needs a fresh look.

The substantive item to re-check is hunk 1: `cef_cache_path` now passes through `non_blank_path`, which is behaviour you did not approve. I believe it is the correct combined behaviour rather than a preference, but it is a real change and it is the one line worth your attention. Everything else is placement or arity.

I chose a merge over a rebase because a rebase would have replayed `7d43002` — the pre-revision approach that `14440f3` deliberately superseded — onto the new base, producing an intermediate commit for a design that was already rejected. The merge keeps `14440f3d` as the first parent, so the approved content stays in the history and the resolution is reviewable as a single merge diff. No other line of the approved diff moved.

## Tests

Ran locally on macOS arm64, all green:

- `cargo check --all-targets` — clean
- `cargo build` (workspace root, no `-p`) — clean
- `cargo test --lib` — 546 passed, 0 failed. Includes this branch's five CEF tests in `paths::tests` and both of #738's blank-value tests that I had to re-arity (`test_from_figment_ignores_blank_values`, `test_from_figment_ignores_blank_cli_database_url`)
- `cargo test --test openapi_test` — 1 passed. No API type moved; run because the merge brought `openapi.json` changes in
- `cargo test --test pipeline_lifecycle_test` — 3 passed
- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --features efp,nvidia -- -D warnings` — clean

**Not run:** the rest of `backend/tests/` (the GStreamer integration tests), and anything on Linux or Windows. The resolution touches no GStreamer element, pad or closure, so I ran `pipeline_lifecycle_test` as a precaution rather than because the diff called for it. CI is still running at the time of writing.

**Not verified:** that CEF actually picks up the cache directory at runtime. This was a textual conflict resolution; I did not start a flow with an HTML source.

srperens (approved):
**Supersedes my `14440f3d` review (id 5049188689, Approve).** That review was dismissed when the branch merged main forward to `e47b3e5`; the merge brought in unrelated main-branch changes (blank-env-var handling, the macOS headless-entry restructuring, ICE-availability logging) but none of them touch this PR's own diff. Re-verified from scratch at the new head rather than trusting that.

**Verdict: Approve**

**Claims**

|---|---|---|
| The PR's actual diff against current `main` is byte-identical to what I approved at `14440f3d` | `CONFIRMED` | `git diff 14440f3d..e47b3e5 -- backend/src/paths.rs .strom.toml.example` is empty; `git diff origin/main...pr671 --stat` shows only `config.rs`, `main.rs`, `paths.rs`, `.strom.toml.example` — the same four files as before |
| CEF cache path still derives from `data_dir`, not the independently-overridable `flows_path` | `CONFIRMED` | `backend/src/paths.rs:153` — `fn default_cef_cache_dir(base_dir: &Path) -> PathBuf {` |
| `set_var` still runs before any thread is spawned in `main()` | `CONFIRMED` | `backend/src/main.rs:223-382` (from `fn main()` through the `GST_CEF_CACHE_LOCATION` block) contains no `thread::spawn`/`tokio::spawn` call |
| Env var wins over the derived default | `CONFIRMED` | `backend/src/main.rs:366` — `match std::env::var_os("GST_CEF_CACHE_LOCATION") {` — the `Some(existing)` arm only logs, never overwrites |
| The 5 regression tests for this derivation execute in CI, not skip | `CONFIRMED` | CI run `33974253931`, job `Check (Linux)`: `test paths::tests::test_cef_cache_differs_per_data_dir ... ok`, `test_cef_cache_ignores_flows_path_override ... ok`, `test_cef_cache_is_stable_for_one_data_dir ... ok`, `test_explicit_cef_cache_path_override ... ok`, `test_resolve_exposes_data_dir ... ok` |

**Diagnosis** — Unchanged from `14440f3d`: the derivation keys the CEF cache directory on `data_dir` (not independently overridable, unlike `flows_path`), closing both collision triggers my earlier `0d9f28a6` review found. `ABSOLUTE` for the class of bug named in the PR — same-data-dir and shared-flows-path collisions both close — bounded only by data directories that are themselves distinct, which the codebase already requires for JSON storage.

**Radius** — `LOCAL`: additive field on `Config`/`DataPaths`/CLI args, no `strom-types`, API or `StromEvent` surface touched, no thread/timer/probe/pipeline-object lifetime introduced.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check`, `sccache preflight` all pass at `e47b3e5`; `Build (macOS)`/`Build (Windows)` skip as usual, and this diff has no platform-`cfg` code for them to miss.

Confidence: HIGH

PR #676, opened 2026-08-18
fix(recorder): request splitmuxsink pads for connected tracks only

wagenet:
## Problem

`splitmuxsink` releases a GOP only once every requested sink pad has reached the next GOP. A pad that is never fed keeps its running time at `GST_CLOCK_STIME_NONE` forever, so the sink waits on it indefinitely.

A recorder configured with one video and one audio track, where the flow connects only video, therefore recorded nothing at all — not a video-only file, an empty one.

## What changed

Pads used to be requested per *configured* track at build time. They are now requested from an element-setup hook, for *connected* tracks only.

That hook runs after the flow has linked every block but before the pipeline leaves NULL. It is the first point where connectivity is knowable — an input identity's sink pad being linked answers "will this track ever carry data?" exactly — and still early enough to request pads before any data flows.

## Why not request them from the caps probes

That is the obvious alternative, and it was tried here first. It fixes the stall and breaks the ordinary case instead: the muxer starts on the first track's data, so whichever track negotiates caps second loses the race. mp4mux refuses the pad (`Not providing request pad after stream start`); matroskamux grants it and drops the data.

Measured against a harness mirroring this block's chain and muxer config (`muxer` element instance, `use-robust-muxing`, 12 h reserved moov):

| | |
|---|---|
| `avenc_aac` caps before `x264enc` | 12 of 12 runs, by 5–29 ms |
| mp4, both tracks connected, video lost | 11 of 12 runs |
| …with the per-leg queues from #675 | 12 of 12 |
| …with robust muxing disabled | 6 of 8 |

A refused pad leaves its leg dead-ended, which takes the pipeline down with `Internal data stream error … not-linked`.

Requesting up front also keeps splitmuxsink's reference stream on the primary video pad, instead of a late `video` pad demoting a live reference mid-stream.

## Failure paths

A connected track that turns out to carry raw or unsupported media hands its pad back from the caps probe, as does any failure while wiring the leg up. Either would otherwise stall the recording — the same bug in a different disguise.

## Tests

`backend/tests/recorder_unfed_track_test.rs`, three arms × three containers (mp4, mkv, mpegts): video connected only, audio connected only, and both — the last repeated 3× per container because the failure it guards is a race.

The tests demux the recording and assert on the streams inside it, not on file size; a dropped track still leaves a large, valid, single-stream file. They also run the element-setup hook the way the pipeline manager does, since that hook is what decides connectivity.

Reverting the fix breaks them: removing the connectivity check makes both unfed-track arms fail with `pipeline never reached EOS within 30s`, the original stall.

Ran locally on macOS: `cargo test --workspace` (506 unit tests plus all integration tests, including `pipeline_lifecycle_test` and `recorder_splitmux_threading_test`) green, and `cargo clippy --workspace --all-targets` clean. The three recorder arms genuinely executed there — `avenc_aac` is present, and they take ~800 ms rather than the ~11 ms of a skip.

They did *not* run in CI before this branch. `gst-libav` failed to load with `libblas.so.3: cannot open shared object file`, so `avenc_aac` was missing and all three tests returned early through `plugins_available()` — "passing" in 11 ms. `libblas3` was already in the package list and being restored; `cache-apt-pkgs-action` just does not run postinst, so the `update-alternatives` link for `libblas.so.3` was never created. `ldconfig` alone does not help — there is no link for it to find.

The second commit reinstalls `libblas3`/`liblapack3` so postinst runs, and sets `STROM_REQUIRE_GST_PLUGINS=1` for the Linux test step so a missing element fails and names itself instead of skipping green. On the latest run both now execute for real: `recorder_unfed_track_test` 3 passed in 0.92 s and the pre-existing `recorder_splitmux_threading_test` 1 passed in 0.39 s, against 0.01 s each before.

## On the earlier review

Both points still hold. The caps probe cannot re-fire after teardown — `parser_inserted.swap(true, SeqCst)` runs the body at most once per chain, and the `WeakRef` upgrade returns `None` once the pipeline is gone. Releasing a partially-configured pad is still the mechanism for unsupported codecs and wiring failures, now on paths reached before the leg carries data.

## Known limitation

A track that negotiates caps and then produces no buffers — a parser that never syncs, an ingest that goes silent right after connecting — still stalls the recording, as before. Connectivity is knowable at build time; "will this connected track actually produce data?" is not.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,17 @@ jobs:
           packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
+      # cache-apt-pkgs-action restores package files but does not run postinst,
+      # so update-alternatives links such as /usr/lib/*/libblas.so.3 never get
+      # created. gst-libav then fails to load ("libblas.so.3: cannot open shared
+      # object file"), avenc_aac goes missing, and every test requiring it skips
+      # green. Reinstalling runs postinst; ldconfig then picks the links up.
+      - name: Repair alternatives for cached libraries
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y --reinstall libblas3 liblapack3
+          sudo ldconfig
+
       # save-if main: PR runs restore the main-branch cache (GHA cache scoping
       # falls back to the base branch) but don't save their own per-branch
       # copy — that copy doubled every rust-cache entry against the 10 GB
@@ -115,6 +126,9 @@ jobs:
 
       - name: Run tests on backend
         run: cargo test --package strom --features efp,nvidia
+        env:
+          # Turn a missing-element skip into a failure — see plugins_available().
+          STROM_REQUIRE_GST_PLUGINS: 1
 
       - name: Run tests on MCP server
         run: cargo test --package strom-mcp-server
@@ -291,6 +305,17 @@ jobs:
           packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
+      # cache-apt-pkgs-action restores package files but does not run postinst,
+      # so update-alternatives links such as /usr/lib/*/libblas.so.3 never get
+      # created. gst-libav then fails to load ("libblas.so.3: cannot open shared
+      # object file"), avenc_aac goes missing, and every test requiring it skips
+      # green. Reinstalling runs postinst; ldconfig then picks the links up.
+      - name: Repair alternatives for cached libraries
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y --reinstall libblas3 liblapack3
+          sudo ldconfig
+
       - name: Download frontend dist
         uses: actions/download-artifact@v8
         with:
@@ -400,6 +425,17 @@ jobs:
         with:
           packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
+
+      # cache-apt-pkgs-action restores package files but does not run postinst,
+      # so update-alternatives links such as /usr/lib/*/libblas.so.3 never get
+      # created. gst-libav then fails to load ("libblas.so.3: cannot open shared
+      # object file"), avenc_aac goes missing, and every test requiring it skips
+      # green. Reinstalling runs postinst; ldconfig then picks the links up.
+      - name: Repair alternatives for cached libraries
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y --reinstall libblas3 liblapack3
+          sudo ldconfig
 
       - name: Download frontend dist
         uses: actions/download-artifact@v8
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,7 @@ jobs:
       # created. gst-libav then fails to load ("libblas.so.3: cannot open shared
       # object file"), avenc_aac goes missing, and every test requiring it skips
       # green. Reinstalling runs postinst; ldconfig then picks the links up.
+      # Only needed here: the build jobs never load a plugin.
       - name: Repair alternatives for cached libraries
         run: |
           sudo apt-get update -qq
@@ -305,17 +306,6 @@ jobs:
           packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
-      # cache-apt-pkgs-action restores package files but does not run postinst,
-      # so update-alternatives links such as /usr/lib/*/libblas.so.3 never get
-      # created. gst-libav then fails to load ("libblas.so.3: cannot open shared
-      # object file"), avenc_aac goes missing, and every test requiring it skips
-      # green. Reinstalling runs postinst; ldconfig then picks the links up.
-      - name: Repair alternatives for cached libraries
-        run: |
-          sudo apt-get update -qq
-          sudo apt-get install -y --reinstall libblas3 liblapack3
-          sudo ldconfig
-
       - name: Download frontend dist
         uses: actions/download-artifact@v8
         with:
@@ -426,17 +416,6 @@ jobs:
           packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
-      # cache-apt-pkgs-action restores package files but does not run postinst,
-      # so update-alternatives links such as /usr/lib/*/libblas.so.3 never get
-      # created. gst-libav then fails to load ("libblas.so.3: cannot open shared
-      # object file"), avenc_aac goes missing, and every test requiring it skips
-      # green. Reinstalling runs postinst; ldconfig then picks the links up.
-      - name: Repair alternatives for cached libraries
-        run: |
-          sudo apt-get update -qq
-          sudo apt-get install -y --reinstall libblas3 liblapack3
-          sudo ldconfig
-
       - name: Download frontend dist
         uses: actions/download-artifact@v8
         with:
@@ -535,6 +514,9 @@ jobs:
         run: |
           cargo test --package strom --features nvidia
           cargo test --package strom-mcp-server
+        env:
+          # Turn a missing-element skip into a failure — see plugins_available().
+          STROM_REQUIRE_GST_PLUGINS: 1
 
       - name: Build
         run: |
@@ -607,6 +589,9 @@ jobs:
         run: |
           cargo test --package strom --features efp,nvidia
           cargo test --package strom-mcp-server
+        env:
+          # Turn a missing-element skip into a failure — see plugins_available().
+          STROM_REQUIRE_GST_PLUGINS: 1
 
       - name: Build
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -514,9 +514,6 @@ jobs:
         run: |
           cargo test --package strom --features nvidia
           cargo test --package strom-mcp-server
-        env:
-          # Turn a missing-element skip into a failure — see plugins_available().
-          STROM_REQUIRE_GST_PLUGINS: 1
 
       - name: Build
         run: |
--- a/backend/src/blocks/builder.rs
+++ b/backend/src/blocks/builder.rs
@@ -324,6 +324,10 @@ impl BlockBuildContext {
     /// Use this to connect GLib signals on elements that need the event broadcaster
     /// (e.g., splitmuxsink's format-location signal for recording status).
     /// The GStreamer element(s) should be captured in the closure during build time.
+    ///
+    /// Runs after construction has linked every block and before the pipeline leaves
+    /// NULL. Recorder requests its splitmuxsink pads here and needs both halves of that
+    /// window; moving this call breaks it silently.
     pub fn register_element_setup(&self, setup: ElementSetupFn) {
         self.element_setups.borrow_mut().push(setup);
     }
--- a/backend/src/blocks/builtin/recorder.rs
+++ b/backend/src/blocks/builtin/recorder.rs
@@ -17,6 +17,10 @@
 //! audio_in_N (identity) --[pad probe]--> [parser chain] --> splitmuxsink:audio_0..N
 //! ```
 //!
+//! splitmuxsink stalls forever on a pad that is never fed, so sink pads are requested at
+//! pipeline start for connected tracks only — not for every configured track, and not
+//! lazily from the pad probes.
+//!
 //! Output files are written to: {media_path}/{output_dir}/{filename_prefix}_%05d.{ext}
 
 use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult, BlockBuilder};
@@ -26,7 +30,7 @@ use gstreamer as gst;
 use gstreamer::prelude::*;
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use strom_types::{
     block::{EnumValue, *},
     PropertyValue, *,
@@ -344,76 +348,29 @@ impl BlockBuilder for RecorderBuilder {
         let mut elements: Vec<(String, gst::Element)> =
             vec![(sink_id.clone(), splitmuxsink.clone())];
 
-        // --- Request pads from splitmuxsink ---
-        // All splitmuxsink sink pads are "On request". Pads must be requested before
-        // the pipeline starts; splitmuxsink manages them across file segments internally.
+        // --- splitmuxsink sink pads: requested at pipeline start, connected tracks only ---
         //
-        // splitmuxsink pad templates:
-        //   video           -> first video track (request_pad_simple)
-        //   video_aux_%u    -> additional video tracks (request_pad via template)
-        //   audio_%u        -> audio tracks (request_pad via template)
-        let video_aux_template = if num_video_tracks > 1 {
-            Some(splitmuxsink.pad_template("video_aux_%u").ok_or_else(|| {
-                BlockBuildError::ElementCreation(
-                    "splitmuxsink: no video_aux_%u pad template".to_string(),
-                )
-            })?)
-        } else {
-            None
-        };
+        // splitmuxsink releases a GOP only once every requested pad has reached the next
+        // GOP, so a pad that is never fed stalls the recording. An unconnected track must
+        // not get one.
+        //
+        // The element-setup hook at the end of this function requests them: it runs after
+        // the flow has linked every block but before the pipeline leaves NULL, so
+        // connectivity is known and the muxer has not started. Requesting from the caps
+        // probes instead loses a race — the muxer starts on the first track's data, after
+        // which mp4mux refuses the second pad and matroskamux drops its data. Requesting
+        // up front also keeps the reference stream on the primary video pad.
+        //
+        // Templates: video (first connected video), video_aux_%u (the rest), audio_%u.
 
-        let video_sink_pad_names: Vec<String> = (0..num_video_tracks)
-            .map(|i| {
-                let pad = if i == 0 {
-                    splitmuxsink.request_pad_simple("video").ok_or_else(|| {
-                        BlockBuildError::ElementCreation(
-                            "splitmuxsink: failed to request video pad".to_string(),
-                        )
-                    })?
-                } else {
-                    let tmpl = video_aux_template.as_ref().unwrap();
-                    splitmuxsink.request_pad(tmpl, None, None).ok_or_else(|| {
-                        BlockBuildError::ElementCreation(format!(
-                            "splitmuxsink: failed to request video_aux pad for track {}",
-                            i
-                        ))
-                    })?
-                };
-                debug!(
-                    "Recorder {}: requested splitmuxsink pad: {}",
-                    instance_id,
-                    pad.name()
-                );
-                Ok::<String, BlockBuildError>(pad.name().to_string())
-            })
-            .collect::<Result<Vec<_>, _>>()?;
-
-        let audio_template = splitmuxsink.pad_template("audio_%u").ok_or_else(|| {
-            BlockBuildError::ElementCreation("splitmuxsink: no audio_%u pad template".to_string())
-        })?;
-
-        let audio_sink_pad_names: Vec<String> = (0..num_audio_tracks)
-            .map(|_| {
-                splitmuxsink
-                    .request_pad(&audio_template, None, None)
-                    .ok_or_else(|| {
-                        BlockBuildError::ElementCreation(
-                            "splitmuxsink: failed to request audio pad".to_string(),
-                        )
-                    })
-                    .map(|p| {
-                        debug!(
-                            "Recorder {}: requested splitmuxsink audio pad: {}",
-                            instance_id,
-                            p.name()
-                        );
-                        p.name().to_string()
-                    })
-            })
-            .collect::<Result<Vec<_>, _>>()?;
+        // Pad name per track, filled in by that hook. Empty means unconnected, no pad.
+        let mut video_pad_cells: Vec<Arc<OnceLock<String>>> = Vec::new();
+        let mut audio_pad_cells: Vec<Arc<OnceLock<String>>> = Vec::new();
+        let mut video_input_weaks: Vec<gst::glib::WeakRef<gst::Element>> = Vec::new();
+        let mut audio_input_weaks: Vec<gst::glib::WeakRef<gst::Element>> = Vec::new();
 
         // --- Create video input chains ---
-        for (vi, video_sink_pad_name_for_chain) in video_sink_pad_names.iter().enumerate() {
+        for vi in 0..num_video_tracks {
             let video_input_id = format!("{}:video_input_{}", instance_id, vi);
             let video_input = gst::ElementFactory::make("identity")
                 .name(&video_input_id)
@@ -425,7 +382,10 @@ impl BlockBuilder for RecorderBuilder {
             let parser_inserted = Arc::new(AtomicBool::new(false));
             let splitmuxsink_weak = splitmuxsink.downgrade();
             let instance_id_clone = instance_id.to_string();
-            let video_sink_pad_name_clone = video_sink_pad_name_for_chain.clone();
+
+            let pad_name_cell: Arc<OnceLock<String>> = Arc::new(OnceLock::new());
+            video_pad_cells.push(Arc::clone(&pad_name_cell));
+            video_input_weaks.push(video_input.downgrade());
 
             // Use a pad probe on the identity src pad to detect caps and insert parser
             let src_pad = video_input.static_pad("src").ok_or_else(|| {
@@ -464,6 +424,30 @@ impl BlockBuilder for RecorderBuilder {
                     let caps_name = structure.name().to_string();
                     debug!("Recorder {}: video caps detected: {}", instance_id_clone, caps_name);
 
+                    let splitmuxsink = match splitmuxsink_weak.upgrade() {
+                        Some(e) => e,
+                        None => {
+                            error!("Recorder {}: splitmuxsink element no longer exists", instance_id_clone);
+                            return gst::PadProbeReturn::Ok;
+                        }
+                    };
+
+                    // Empty cell: this track was unconnected at build time, so it has no pad.
+                    let sink_pad = match pad_name_cell.get().and_then(|n| splitmuxsink.static_pad(n)) {
+                        Some(p) => p,
+                        None => {
+                            warn!(
+                                "Recorder {}: video track {} is carrying data but was not connected when the pipeline was built, so it has no splitmuxsink pad and will not be recorded",
+                                instance_id_clone, vi
+                            );
+                            return gst::PadProbeReturn::Ok;
+                        }
+                    };
+
+                    // Every path that gives up below must hand the pad back, or splitmuxsink
+                    // waits on it forever and the recording stalls.
+                    let give_pad_back = || splitmuxsink.release_request_pad(&sink_pad);
+
                     let (parser_factory, config_interval) = if caps_name == "video/x-h264" {
                         ("h264parse", -1i32)
                     } else if caps_name == "video/x-h265" {
@@ -473,27 +457,22 @@ impl BlockBuilder for RecorderBuilder {
                             "Recorder {}: received raw video — recorder only accepts pre-encoded video. Add an encoder block before the recorder.",
                             instance_id_clone
                         );
+                        give_pad_back();
                         return gst::PadProbeReturn::Ok;
                     } else {
                         warn!(
                             "Recorder {}: unsupported video codec: {} (supported: H.264, H.265)",
                             instance_id_clone, caps_name
                         );
+                        give_pad_back();
                         return gst::PadProbeReturn::Ok;
-                    };
-
-                    let splitmuxsink = match splitmuxsink_weak.upgrade() {
-                        Some(e) => e,
-                        None => {
-                            error!("Recorder {}: splitmuxsink element no longer exists", instance_id_clone);
-                            return gst::PadProbeReturn::Ok;
-                        }
                     };
 
                     let bin = match splitmuxsink.parent().and_then(|p| p.downcast::<gst::Bin>().ok()) {
                         Some(b) => b,
                         None => {
                             error!("Recorder {}: splitmuxsink has no Bin parent", instance_id_clone);
+                            give_pad_back();
                             return gst::PadProbeReturn::Ok;
                         }
                     };
@@ -506,6 +485,7 @@ impl BlockBuilder for RecorderBuilder {
                         Ok(p) => p,
                         Err(e) => {
                             error!("Recorder {}: failed to create {}: {}", instance_id_clone, parser_factory, e);
+                            give_pad_back();
                             return gst::PadProbeReturn::Ok;
                         }
                     };
@@ -517,10 +497,12 @@ impl BlockBuilder for RecorderBuilder {
 
                     if let Err(e) = bin.add(&parser) {
                         error!("Recorder {}: failed to add parser to bin: {}", instance_id_clone, e);
+                        give_pad_back();
                         return gst::PadProbeReturn::Ok;
                     }
                     if let Err(e) = parser.sync_state_with_parent() {
                         error!("Recorder {}: failed to sync parser state: {}", instance_id_clone, e);
+                        give_pad_back();
                         return gst::PadProbeReturn::Ok;
                     }
 
@@ -528,6 +510,7 @@ impl BlockBuilder for RecorderBuilder {
                         Some(p) => p,
                         None => {
                             error!("Recorder {}: parser has no sink pad", instance_id_clone);
+                            give_pad_back();
                             return gst::PadProbeReturn::Ok;
                         }
                     };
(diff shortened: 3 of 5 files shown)


srperens (dismissed):
**Intent:** Fix a stall in the Recorder block where a requested-but-never-fed `splitmuxsink` sink pad holds the mux indefinitely — specifically when a track is declared in the block configuration but its pipeline leg is never connected. Pad requests are now deferred to the first caps event on each leg rather than being made up-front at build time.

**Risk:** Medium. The change directly alters pipeline build sequencing for the recorder. The new regression test (`recorder_unfed_track_test.rs`) is the key guard here.

---

**What looks good:**

- Deferring pad requests to caps probes is the correct GStreamer pattern for this class of problem — caps (EVENT_DOWNSTREAM) probes fire at most once per caps negotiation, not per-buffer, so there is no hot-path cost.
- Releasing the pad on link failure prevents the orphan-pad stall from reappearing through a different path.
- The regression test directly covers the scenario (unfed track does not stall the recording).
- No BUFFER probes are introduced.

**Two things worth verifying:**

1. **EOS/flush replay:** Does the caps probe guard against re-firing on a `CAPS` event that arrives after teardown (e.g. during pipeline state change back through READY)? A `WeakRef` upgrade returning `None` should handle this safely, but worth confirming the probe is removed or becomes a no-op after the pad is linked.

2. **Partial release on link failure:** When `src_pad.link(&pad)` fails and the pad is released, does `splitmuxsink` handle the release correctly while it is already partially configured? A mux in `PAUSED` or `PLAYING` state releasing a not-yet-fed pad is the normal case the fix targets, so this should be fine, but the test could be extended to assert that the mux does not log an error on the release.

Both are informational — the core fix is correct.

wagenet:
_Generated by Claude with guidance_

---

### The approach changed — this needs a fresh look

@srperens reviewed the previous version, which deferred each pad request to that leg's caps probe. It fixes the stall and introduces a worse bug, so the fix has been rewritten. The description covers the current design; this is what moved.

**Why deferring to the caps probe fails.** The muxer starts on the first track's data, so whichever track negotiates caps second loses the race — mp4mux refuses the pad, matroskamux grants it and drops the data. Measured against a harness mirroring this block's chain and muxer config:

| | |
|---|---|
| `avenc_aac` caps before `x264enc` | 12 of 12 runs, by 5–29 ms |
| mp4, both tracks connected, video lost | 11 of 12 runs |
| …with the per-leg queues from #675 | 12 of 12 |
| …with robust muxing disabled | 6 of 8 |

The refused leg dead-ends and takes the pipeline down with `Internal data stream error … not-linked`. So the default configuration — mp4, one video, one audio, both connected — failed about nine times in ten, which is worse than the bug being fixed: that one needs a misconfigured recorder to appear at all.

**What it does now.** Pads are requested from an element-setup hook, for connected tracks only. It runs after the flow has linked every block but before the pipeline leaves NULL, so connectivity is known and the muxer has not started. An input's sink pad being linked answers exactly the question the caps probe was guessing at.

**Your two points still hold.** The probe cannot re-fire after teardown (`parser_inserted.swap(true, SeqCst)`, plus the `WeakRef` upgrade returning `None`), and releasing a partially-configured pad is still the mechanism for unsupported codecs and wiring failures — now on paths reached before the leg carries data.

**Tests** now demux the recording and assert on the streams inside it, not on file size. That is why the old control arm passed while silently losing video. They also run the element-setup hook, since that hook is what decides connectivity. Reverting the connectivity check makes both unfed-track arms fail with `pipeline never reached EOS within 30s`.

Also rebased onto current `main`, picking up #675 and #678.

**One limitation, unchanged:** a track that negotiates caps and then produces no buffers still stalls the recording. Connectivity is knowable at build time; whether a connected track will actually produce data is not.

srperens (commented):
Reviewing PR #676: fix(recorder): request splitmuxsink pads for connected tracks only
Verified against c83079a0 vs main 4a36815.

**Summary**
- Moves `splitmuxsink` sink-pad requests from build time (one pad per *configured* track) to a `register_element_setup` hook that requests pads only for tracks whose identity sink pad is actually linked, avoiding the indefinite stall `splitmuxsink` causes when a requested pad is never fed.
- Adds `give_pad_back()` (`release_request_pad`) on every failure path in the caps-probe chains (unsupported codec, element creation/linking failure) so a track that turns out unusable doesn't leave a dangling stalled pad either.
- Adds `backend/tests/recorder_unfed_track_test.rs`: 3 tests × 3 containers (mp4/mkv/mpegts), exercising `RecorderBuilder::build` and `ctx.take_element_setups()` directly.

**Claim verification**
|---|---|---|
| Hook runs after all blocks are linked, before pipeline leaves NULL | CONFIRMED | `backend/src/gst/pipeline/lifecycle.rs:42` `self.setup_bus_watch();` (which invokes `element_setup_fns`, `backend/src/gst/pipeline/bus.rs:49-50` `for setup_fn in element_setups { setup_fn(flow_id, ...)`) runs before `lifecycle.rs:61` `self.pipeline.set_state(gst::State::Ready)`. Linking itself happens earlier, inside `PipelineManager::new()` (`backend/src/gst/pipeline/construction.rs:271` `manager.try_link_elements(link)`), which per `backend/src/state.rs:891-917` completes before `manager.start()` is ever called. |
| Caps probe cannot re-fire after teardown (`parser_inserted` swap + `WeakRef`) | CONFIRMED | `backend/src/blocks/builtin/recorder.rs:407` `if parser_inserted.swap(true, Ordering::SeqCst) { return gst::PadProbeReturn::Ok; }`, and `recorder.rs:427` upgrades `splitmuxsink_weak` (a `WeakRef`), returning early on `None`. |
| First connected video track gets the primary `video` pad, rest get `video_aux_%u` | CONFIRMED | `recorder.rs` element-setup hook: `let requested = if !have_primary_video { splitmuxsink.request_pad_simple("video") } else { ... request_pad(&tmpl_video_aux, ...) }` — gated on `have_primary_video`, not on loop index `vi`, so the first *connected* track (not track 0) takes primary. |
| Every failure path releases the pad it holds (no permanent stall on codec/link failure) | CONFIRMED | All ~14 error branches in both the video and audio caps-probe closures call `give_pad_back();` before `return gst::PadProbeReturn::Ok;` (e.g. `recorder.rs:462-467` unsupported/raw video, `recorder.rs:539-543` queue creation, mirrored in the audio closure at lines 664-800). |
| `release_request_pad` is the correct API for handing a pad back to `splitmuxsink` | CONFIRMED (pattern precedent) | Same method already used in this codebase: `backend/src/gst/thumbnail_tap.rs:475` `self.tee.release_request_pad(tee_src_pad);`. |
| Test exercises the real changed code path, not a hand-rebuilt topology | CONFIRMED | `backend/tests/recorder_unfed_track_test.rs:91-93` calls `RecorderBuilder.build(...)` directly, and line ~156 runs `for setup in ctx.take_element_setups() { setup(...) }` — the actual hook mechanism, not a reimplementation. |
| Test would fail if the fix were reverted | CONFIRMED (by construction) | `assert_recorded` in the test asserts `reached_eos` within a 30 s deadline (`recorder_unfed_track_test.rs` `run_recorder`); an unfed `splitmuxsink` pad stalls indefinitely under the old behaviour, so reversion times out and panics rather than passing. I did not revert-and-run this myself — see Tests & CI below for why that's moot here. |
| "Ran locally: cargo test --workspace ... all green with no skips" | CONTRADICTED (for CI; local run not verifiable) | The equivalent CI run did *not* run these tests — see Tests & CI. I cannot verify the author's local claim directly, but the burden here is CI evidence, and CI shows a skip, not a run. |
| Requesting from caps probes instead loses a race (mp4mux refuses, matroskamux drops) | EXTERNAL | Behavior of `mp4mux`/`matroskamux` request-pad handling after stream-start is upstream GStreamer behavior; not verifiable from this repo, but consistent with the documented mp4mux error string `Not providing request pad after stream start` quoted in the PR body. |

**Diagnosis**
Right layer: the fix addresses the actual mechanism (`splitmuxsink` waiting on every *requested* pad regardless of whether it's fed) rather than working around a symptom — CONFIRMED via `recorder.rs`'s doc comment at the top of the diff and the pad-release semantics traced above. It is **absolute** for the stall itself (any track/container/count is handled by the same connected-tracks loop), but explicitly **bounded**: a track that negotiates caps and then produces zero buffers still stalls, which the PR body states as a known, unaddressed limitation ("Known limitation" section) — worth restating in the merged PR discussion since it's not encoded as a doc comment in the code itself, only in the PR body which won't survive as durable documentation per this repo's own doc policy.

History check: `backend/src/blocks/builtin/recorder.rs` was touched immediately prior by `c93757e` "fix(recorder): add a queue per leg between parser and splitmuxsink (#675)". Two fixes in the same pad-linking area in quick succession — but they address genuinely different failure modes (a buffering/ordering race vs. an indefinite stall on an unfed pad), and this PR's own measurement table shows the #675 queues alone do not fix this bug (12/12 stall reproduction with queues, per the PR body), so this doesn't look like the same root cause resurfacing.

No request/retry/feedback mechanism is introduced here (pad *release* is idempotent-per-invocation, not a retried request), so 2g doesn't apply.

**Blast radius**
LOCAL. `grep -rn "RecorderBuilder" backend/src` returns only `backend/src/blocks/builtin/mod.rs` (registration) and `recorder.rs` itself — no other callers. The removed build-time pad-request variables (`video_sink_pad_names`, `audio_sink_pad_names`, `video_aux_template`) have no other references (`grep` confirms).

Traced the one genuinely additive piece — the `input_is_connected` connectivity check reading `Pad::is_linked()` at hook time — for the concurrency question CLAUDE.md's WeakRef rule exists to catch: could the recorder's identity sink pad still be *unlinked* at hook time even though the flow *does* eventually feed it, e.g. via an asynchronously-resolved dynamic pad (`decodebin`/`webrtcbin` `pad-added`)? Traced `backend/src/gst/pipeline/linking.rs:319` `setup_dynamic_pad_handlers` / `pending_links`: cross-block links only go "pending" (resolved later via `connect_pad_added`) when the *named* source pad doesn't exist yet on the upstream element. In this codebase's WHIP input block (`backend/src/blocks/builtin/whip.rs:358-429`), the `decodebin`'s own dynamic pads are wired *internally* to a `tee` (`video_out_tee`/`audio_out_tee`) that is itself statically present from build time; downstream blocks link from that tee's request pad, which resolves synchronously. Because the recorder additionally rejects raw media at its own caps probe (it only accepts pre-encoded H.264/H.265/AAC/etc., forcing an encoder block immediately upstream, and encoder `src` pads are always static/Always), any link that terminates at a recorder's `video_input_N`/`audio_input_N` should resolve in the initial synchronous linking pass inside `PipelineManager::new()`, before `is_linked()` is ever checked. I did not exhaustively audit every other block type (`vision_mixer`, `compositor`, `aes67`, `tams_output`) for the same tee-based egress guarantee, so I can't rule out an unusual custom flow with a block whose only output is a genuinely Sometimes pad feeding a recorder directly — but I found no such block in this repo.

Teardown: the two new closures capture only `WeakRef`s (`splitmuxsink.downgrade()`, `video_input.downgrade()`/`audio_input.downgrade()`) and `Arc<OnceLock<String>>` (plain data, not a GStreamer object) — no strong `Element`/`Pipeline`/`Bin` references are held across signal invocations, matching the CLAUDE.md WeakRef rule. No other code path in the repo calls `release_request_pad` on these same pads (`grep -rn "release_request_pad" backend/src` → only this diff and the unrelated `thumbnail_tap.rs`), so there's no double-release risk against pipeline teardown.

**Tests & CI**
- `Check (Linux)`: **success** — but the success is misleading for this PR's own regression test. From the job log (run 32620052498, job 97146765021): `gst-plugin-scanner` logged `Failed to load plugin '/usr/lib/x86_64-linux-gnu/gstreamer-1.0/libgstlibav.so': libblas.so.3: cannot open shared object file: No such file or directory`. `avenc_aac` (from `gst-libav`) is in this test's own `REQUIRED` list (`recorder_unfed_track_test.rs:26`), so `plugins_available()` returns false and every test in the file hits its `eprintln!("skipping: required GStreamer elements not installed"); return;` guard. The log confirms this mechanically: `Running tests/recorder_unfed_track_test.rs ... running 3 tests` at `05:28:51.5712322Z`, next binary starting at `05:28:51.5796377Z` — all 3 tests (each meant to build a real pipeline, encode ~1s of video/audio, and demux the output) "passed" in under 9 ms, which is only possible if they never ran a pipeline at all. The pre-existing canary `recorder_splitmux_threading_test.rs` (which also lists `avenc_aac` as required, line 36) shows the identical pattern: 1 test in ~8 ms. **This is a silent-skip, not a pass**, and per CLAUDE.md's Tests section ("Tests that skip on a missing element pass green and guard nothing") that's a rule violation — this PR's own regression guard, the one thing standing between this fix and a silent revert, did not execute in CI.
- Other Linux tests in the same job did run for real: `pipeline_lifecycle_test.rs` (the reference-cycle canary) ran in 0.93 s with visible GStreamer teardown activity and passed 3/3 — that canary is fine.
- `.github/workflows/ci.yml:89` already lists `gstreamer1.0-libav` in the apt install line, so the package *is* declared — the failure is a missing transitive shared-library dependency of that plugin (`libblas.so.3`) on the runner image, not a missing apt package for gst-libav itself.
- API Contract Check, WASM check, sccache preflight, both Linux arch builds: all green, unaffected by this finding (none of them exercise these integration tests).
- Build (macOS) / Build (Windows): skipped, as expected for push/PR — this diff has no `#[cfg(target_os = ...)]` or platform-gated code, so that's not a gap here.
- Canary check: `pipeline_lifecycle_test.rs` ran and passed (see above). `openapi_test.rs` is unaffected — this diff adds no API/WebSocket surface.

**Risks / edge cases**
- The new regression test has never actually executed against this diff in CI (see Tests & CI). The fix logic itself checks out on static reading, but "the test would catch a revert" is unverified in practice, not just in theory.
- SPECULATIVE (not verified): I have not run `apt-cache search libblas.so.3` or equivalent against the actual CI runner image, so the exact missing package name is my best inference from standard Debian/Ubuntu naming (`libblas3` typically ships `libblas.so.3`), not confirmed against this runner's package index.
- The "Known limitation" (a connected track that negotiates caps but never produces buffers still stalls) is accurately disclosed in the PR body but isn't captured anywhere in code comments or `docs/`; per this repo's own documentation policy the PR/review trail is the only record of it.

**Requested changes**
1. Fix the CI environment so `gst-libav`/`avenc_aac` actually loads on the Linux runner, then re-run CI and confirm the log shows real per-container pipeline runs (multi-hundred-ms+ durations, not <10 ms) for `recorder_unfed_track_test.rs` and `recorder_splitmux_threading_test.rs`. Smallest concrete change: add the missing shared-library package (likely `libblas3` on Ubuntu; verify against the actual runner) to the `packages:` list in `.github/workflows/ci.yml` at line 89 (and the matching lines 291/401 if those runners hit the same gap), in the same PR or a fast-follow before merge.
2. Once CI actually exercises the new tests, re-verify the "passed" conclusion — I have no reason from static reading to expect a failure, but per this protocol a green conclusion on a skip isn't evidence.

**Confidence: MED**
The fix mechanism, WeakRef discipline, and pad-release-on-failure logic all check out from static reading, and the design reasoning in the PR body is sound and matches what the code does. The reason this isn't HIGH: the one artifact that's supposed to prove it — the new regression test — has not actually run in CI, so "tests pass" for this change is currently an assertion, not a verified fact.

wagenet:
The silent-skip finding was correct and is now fixed — but not by adding `libblas3`, and my own first attempt was wrong too.

`libblas3` is already in the package list at `ci.yml:89`, and the log shows it restoring:

```
- libblas3=3.12.0-3build1.1.tar restoring...
```

I first assumed `cache-apt-pkgs-action` was skipping `ldconfig`, pushed `sudo ldconfig` after the restore, and it failed identically. The action restores package files but never runs postinst, so the `update-alternatives` link for `libblas.so.3` is never created — `ldconfig` has nothing to find. Reinstalling `libblas3`/`liblapack3` runs postinst and does fix it.

Since two guesses in a row were wrong and both were invisible in a green run, the tests no longer skip quietly. `STROM_REQUIRE_GST_PLUGINS=1`, set for the Linux test step, turns a missing element into a failure that names it.

Latest run (`Check (Linux)`, 32688352924) — no plugin-load warning, and both files run for real:

| | before | now |
|---|---|---|
| `recorder_unfed_track_test` | 3 "passed" in 0.01 s | 3 passed in **0.92 s** |
| `recorder_splitmux_threading_test` | 1 "passed" in 0.01 s | 1 passed in **0.39 s** |

That matches local timings, so your point 2 is now satisfied: the regression tests have actually executed against this diff.

One correction on the verification table: "all green with no skips" was about the local macOS run, where `avenc_aac` is present and the arms take ~800 ms. It was accurate for that, but I asserted it without checking CI, where it was not true. The PR body now separates the two.

wagenet:
Things seem to work, but I do not yet confidently understand what this is doing.

srperens (dismissed):
Reviewing PR #676: fix(recorder): request splitmuxsink pads for connected tracks only
Verified against 2f7ae0f vs main 4a36815.

This updates my earlier v3 review at head `c83079a0` (id 5001785978): the code fix is
unchanged, but a new commit landed addressing that review's sole blocker (CI not actually
exercising the new tests). Verdict changes from Comment (MED) to Approve.

**Summary**
- Moves `splitmuxsink` sink-pad requests from build time (one pad per *configured* track) to
  a `register_element_setup` hook that requests pads only for tracks whose identity sink pad
  is linked, avoiding the indefinite stall `splitmuxsink` causes when a requested pad is never
  fed.
- Adds `give_pad_back()` (`release_request_pad`) on every failure path in the caps-probe
  chains so a track that turns out unusable doesn't leave a dangling stalled pad either.
- New commit `2f7ae0f` ("ci: make gst-libav load, and fail instead of skipping when it does
  not"): repairs a CI runner library-linking gap so `avenc_aac` actually loads, and adds
  `STROM_REQUIRE_GST_PLUGINS=1` so a missing element fails the suite instead of skipping green.

**Claim verification**
|---|---|---|
| Hook runs after all blocks are linked, before pipeline leaves NULL | CONFIRMED | `backend/src/gst/pipeline/lifecycle.rs:42` calls `self.setup_bus_watch();` before `lifecycle.rs:61` `.set_state(gst::State::Ready)`; `backend/src/gst/pipeline/bus.rs:48-50` `let element_setups = std::mem::take(&mut self.element_setup_fns); for setup_fn in element_setups { setup_fn(flow_id, ...)`. |
| Caps probe cannot re-fire after teardown | CONFIRMED | `backend/src/blocks/builtin/recorder.rs:407` `if parser_inserted.swap(true, Ordering::SeqCst) {`, and `recorder.rs:427` upgrades `splitmuxsink_weak` (a `WeakRef`), returning early on `None`. Mirrored for audio at `recorder.rs:628` and `:649`. |
| First connected video track gets the primary `video` pad, rest get `video_aux_%u` | CONFIRMED | `recorder.rs:833` `let mut have_primary_video = false;` gates `recorder.rs:846` `let requested = if !have_primary_video { splitmuxsink.request_pad_simple("video") } else { ... }` on connectivity order, not loop index. |
| Every failure path releases the pad it holds | CONFIRMED | `recorder.rs:449` and `:671` define `let give_pad_back = || splitmuxsink.release_request_pad(&sink_pad);`, called before every `return gst::PadProbeReturn::Ok;` on the ~14 error branches in both closures. |
| `release_request_pad` is the correct API for handing a pad back | CONFIRMED (precedent) | Same method already used elsewhere in this codebase: `backend/src/gst/thumbnail_tap.rs:475` `self.tee.release_request_pad(tee_src_pad);`. |
| Test exercises the real changed code path, not a hand-rebuilt topology | CONFIRMED | `backend/tests/recorder_unfed_track_test.rs` calls `RecorderBuilder.build(...)` directly and runs `for setup in ctx.take_element_setups() { setup(...) }` — the actual hook mechanism. |
| Requesting from caps probes instead loses a race (mp4mux refuses, matroskamux drops) | EXTERNAL | `mp4mux`/`matroskamux` request-pad handling after stream-start is upstream GStreamer behavior, not verifiable from this repo. Consistent with the documented mp4mux error string quoted in the PR body, but not independently reproduced by me. |
| gst-libav failed to load on the CI runner because `cache-apt-pkgs-action` skips postinst, so `libblas.so.3` links were never created | CONFIRMED | Job log for this head (`Check (Linux)`, run 32688352924/job 97317444837) line 1144: `libblas3=3.12.0-3build1.1.tar restoring...` (cache restore, no postinst), followed by `.github/workflows/ci.yml` (PR676) line 97 `Repair alternatives for cached libraries` running `apt-get install -y --reinstall libblas3 liblapack3`; log line 1992: `update-alternatives: using /usr/lib/x86_64-linux-gnu/blas/libblas.so.3 to provide ...` — the link now gets created. |
| "The new regression test actually ran, not skipped" (this run) | CONFIRMED | Same job log: `recorder_unfed_track_test.rs` — `video_only_feed_records_despite_configured_audio_track ... ok` at 04:09:41.169 (started 04:09:40.928, ~241ms), `both_tracks_fed_records_both_streams ... ok` at 04:09:41.680 (~511ms) — real per-test durations, not the <10ms pattern a skip produces (previously seen at head `c83079a0`: 3 tests in ~9ms). No `libblas.so.3: cannot open shared object file` / `Failed to load plugin` string appears anywhere in this run's log (searched, zero matches), versus the prior run where it did. |
| `STROM_REQUIRE_GST_PLUGINS=1` turns a missing element into a build/test failure | CONFIRMED | `.github/workflows/ci.yml` (PR676) line 131 sets `STROM_REQUIRE_GST_PLUGINS: 1` on the `Run tests on backend` step (`cargo test --package strom --features efp,nvidia`); `recorder_unfed_track_test.rs` `plugins_available()`: `assert!(std::env::var("STROM_REQUIRE_GST_PLUGINS").is_err(), "STROM_REQUIRE_GST_PLUGINS is set but these elements are missing: {}", missing.join(", "));`. |

**Diagnosis**
Right layer, confirmed via `recorder.rs`'s own doc comment (`recorder.rs:20-22`, added in this
diff) and the pad-release semantics traced above — this addresses the mechanism
(`splitmuxsink` waiting on every *requested* pad regardless of feed), not a downstream
symptom. **Absolute** for the stall itself: the connected-tracks loop applies uniformly
regardless of track count or container. **Bounded**, and explicitly disclosed as such in the
PR body's "Known limitation": a track that negotiates caps and then produces zero buffers
still stalls recording — connectivity is knowable at build time, "will this connected track
ever produce data" is not. That limitation isn't captured in a code comment, only the PR body,
which per this repo's own doc policy won't survive as durable documentation once the PR page
scrolls out of a reviewer's view.

No request/retry/feedback mechanism is introduced (pad *release* fires at most once per
closure invocation via the `parser_inserted` guard, not retried), so the open/closed-loop cost
question doesn't apply here.

History: `recorder.rs` was touched immediately prior by `c93757e` (#675, per-leg queues for a
different threading/GOP-ordering bug). This PR's own measurement table states #675's queues
alone don't fix this stall (12/12 reproduction with queues), so this isn't the same root cause
resurfacing under a new name.

**Blast radius**
LOCAL. `grep -rn "RecorderBuilder" backend/src` returns only the registration in
`backend/src/blocks/builtin/mod.rs:168` and the definition/impl in `recorder.rs` itself — no
other call sites.

Teardown for the additive pieces: the element-setup closure and both caps-probe closures
capture only `WeakRef`s (`splitmuxsink.downgrade()`, `video_input.downgrade()` /
`audio_input.downgrade()`) plus `Arc<OnceLock<String>>` (plain data, not a GStreamer object) —
no strong `Element`/`Pipeline`/`Bin` reference is held across an invocation, matching
CLAUDE.md's WeakRef rule. `grep -rn "release_request_pad" backend/src` finds only this diff and
the unrelated `thumbnail_tap.rs` call, so there's no double-release path against a concurrent
teardown. The connectivity check (`input_is_connected`, `recorder.rs:974`) reads
`Pad::is_linked()` at hook time, after `PipelineManager::new()`'s synchronous linking pass
completes — traced in my prior review at this same code (unchanged since) via
`backend/src/gst/pipeline/linking.rs`'s dynamic-pad handling and the WHIP block's tee-based
egress, concluding any recorder input resolves synchronously before the hook runs, for every
built-in block currently in this repo.

**Tests & CI**
- `Check (Linux)`: **success**, and this time the success is real. Job 97317444837 (run
  32688352924): `recorder_unfed_track_test.rs` — `running 3 tests` / `test result: ok. 3
  passed; ... finished in 0.92s`, individual test timestamps show ~170-510ms each (real
  pipeline builds + demux verification), not a skip. `recorder_splitmux_threading_test.rs` (the
  pre-existing canary that hit the identical silent-skip pattern last run) — `1 passed;
  finished in 0.39s`, versus ~8ms previously.
- `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check & Build (WASM)`, `API Contract Check`,
  `sccache preflight`: all success.
- `Build (macOS)` / `Build (Windows)`: skipped, as expected — no `#[cfg(target_os = ...)]` or
  platform-gated code in this diff.
- Canary: `pipeline_lifecycle_test.rs` (reference-cycle guard) ran and passed 3/3 in the same
  job (`test result: ok. 3 passed; ... finished in 0.94s`). `openapi_test` not applicable — this
  diff adds no API/WebSocket surface.
- Residual gap: `recorder_splitmux_threading_test.rs`'s own `missing_element()` (unchanged by
  the CI-fix commit) has no `STROM_REQUIRE_GST_PLUGINS` backstop — only
  `recorder_unfed_track_test.rs`'s `plugins_available()` got one. If the library-repair step
  ever silently stops working on a future runner image, this canary goes back to skip-passing
  invisibly while the new test would loudly fail. Doesn't block approval (the underlying issue
  is fixed and demonstrated fixed in this run), but it's an inconsistency worth closing — see
  Requested changes.

**Risks / edge cases**
- The connected-but-never-produces-data case is a real, disclosed, unaddressed gap (see
  Diagnosis) — not a reason to block this PR, since it doesn't regress anything and is
  explicitly out of scope, but it should stay visible rather than being forgotten once this PR
  merges.
- SPECULATIVE (not verified): I have not audited every third-party or future block type for the
  tee-based synchronous-egress property the blast-radius argument depends on — only the
  built-in blocks present in this repo today.

**Requested changes**
1. Give `recorder_splitmux_threading_test.rs`'s `missing_element()` the same
   `STROM_REQUIRE_GST_PLUGINS` backstop `recorder_unfed_track_test.rs` got, so both canaries
   fail loudly instead of skip-passing if the CI library fix ever regresses. Smallest concrete
   change: in `backend/tests/recorder_splitmux_threading_test.rs`, change `missing_element()` to
   `assert!` the same way `plugins_available()` does in `recorder_unfed_track_test.rs` before
   returning `Some(...)`, gated on `std::env::var("STROM_REQUIRE_GST_PLUGINS")`. Non-blocking —
   fine to land as a fast-follow rather than holding this PR.

**Confidence: HIGH**
The fix mechanism, WeakRef discipline, and pad-release-on-failure logic check out from static
reading (unchanged since my prior pass at this same code), and this run's CI log now provides
direct, timestamped evidence that the regression test actually built and ran real pipelines
rather than skipping — the gap that capped my prior review at MED is closed.

srperens:
Verified this independently, and it holds up on every point I could check.

**The bug reproduces on `main`.** Taking your test file with `main`'s `recorder.rs`:

```
both_tracks_fed_records_both_streams                     ... ok
audio_only_feed_records_despite_configured_video_track   ... FAILED
video_only_feed_records_despite_configured_audio_track   ... FAILED

panicked: pipeline never reached EOS within 30s (container=mp4, feed connected=audio only)
panicked: pipeline never reached EOS within 30s (container=mp4, feed connected=video only)
```

So the tests are a real guard, not a demonstration — they fail on revert, and they fail on the original stall rather than on something incidental. The both-tracks arm passing is the right control.

**The tests execute for real.** Locally on macOS with `STROM_REQUIRE_GST_PLUGINS=1` set: 3 passed in 0.51s, against the ~0.01s of a skip. `avenc_aac` is present here.

**The CI finding is the most valuable part of this PR**, more than the recorder fix itself. Two regression tests in this area were passing green in 11ms without executing, which is precisely the failure mode `CLAUDE.md` warns about — "tests that skip on a missing element pass green and guard nothing". Tracking it to `cache-apt-pkgs-action` restoring package files without running postinst, so the `update-alternatives` link for `libblas.so.3` never exists, is a genuinely non-obvious diagnosis, and `STROM_REQUIRE_GST_PLUGINS` is the right shape of fix: it makes the next instance of this loud instead of invisible.

Code-wise it is consistent with the project's rules: `WeakRef` with `downgrade()`/`upgrade()` throughout rather than captured strong element references, no new pad probes, no queue property overrides. `register_element_setup` already exists on `main`, so this adds no shared API surface.

The reasoning for the element-setup hook over the caps probes is convincing, and the measured race table is what makes it convincing — "whichever track negotiates caps second loses" with 11 of 12 runs losing video is not something anyone would guess from reading the code. Worth keeping that in the PR record.

Two small things, neither blocking:

1. **The "Repair alternatives" step is added to three jobs, but only one of them runs tests.** `cargo test` on Linux runs in `check-linux` (`ci.yml:128`); the two Linux build jobs do not run tests, and a build does not need `gst-libav` to load. Is the duplication deliberate — for instance to keep the environments identical — or can two of the three go?
2. **`STROM_REQUIRE_GST_PLUGINS` is set only for the Linux test step.** The Windows (`:536`) and macOS (`:608`) jobs also run `cargo test`, so a silent skip there still passes green. If the intent is "never skip quietly", those want it too; if the intent is "Linux is the platform we guarantee element coverage on", that is a reasonable call but worth stating, because the next person will read the asymmetry as an oversight.

Also: the "Known limitation" is the right one to call out. A connected track that negotiates caps and then produces nothing still stalls, and no build-time check can know that — it would need a watchdog on first buffer. Fine to leave out of scope; good that it is written down.

**On your second comment** — "Things seem to work, but I do not yet confidently understand what this is doing" — which part? That is the only thing I would want resolved before merging, since everything else here is measured. If it is the element-setup timing (why the hook is early enough to request pads but late enough to know connectivity), or splitmuxsink's reference-stream behaviour, say which and I will dig into that specific piece with you.

wagenet:
@srperens I pushed a couple more commits to address your remaining concerns. After exploring more I also feel like I sufficiently understand the overall fix.

srperens (dismissed):
Reviewing PR #676: fix(recorder): request splitmuxsink pads for connected tracks only
Verified against 28cceb7c vs main 81cdc83.

This supersedes my earlier v3 reviews at head `2f7ae0f` (ids 5001785978, 5004511320 — the latter already confirmed the core fix and the regression test executing for real in CI). The core fix is unchanged since then; verified only the delta.

**What's new since 2f7ae0f:** (1) two doc-comment additions on `register_element_setup` (`builder.rs:324`) and `input_is_connected` (`recorder.rs:969`) — no functional change; (2) `STROM_REQUIRE_GST_PLUGINS=1` also added to the dormant `build-windows`/`build-macos` "Run tests" steps (`ci.yml:519,594`) — those jobs only run on `workflow_dispatch` (`ci.yml:453,547`), so this is inert on this PR's own checks; (3) a new integration test, `recorder_records_when_driven_through_pipeline_start` (`recorder_unfed_track_test.rs:451`), that builds a `Flow` and drives it through the real `PipelineManager::new(...).start()` — the actual production path, not a hand-built pipeline.

**Claim verification**
|---|---|---|
| New test exercises the real `PipelineManager` start path, not a reconstruction | CONFIRMED | `recorder_unfed_track_test.rs:471-560`: calls `PipelineManager::new(&flow, ...).start()` directly with a `builtin.recorder` block in the flow |
| Encoder-fed recorder inputs are always linked in the immediate pass, never deferred to `pending_links` | CONFIRMED | `backend/src/gst/pipeline/construction.rs:259-277`: `try_link_elements` only pushes to `pending_links` on failure; encoders (`x264enc`/`avenc_aac`) expose static "src" pads present at construction, so this link never fails immediately — matches the new doc comment's own caveat |
| The new tests actually ran (not skip-fast) in this run's CI | CONFIRMED | CI log, run 32930023644: `recorder_unfed_track_test.rs` "4 passed ... finished in 1.21s" and `recorder_splitmux_threading_test.rs` "1 passed ... finished in 0.38s" — both well above the ~0.01s skip time cited in the PR body, and `STROM_REQUIRE_GST_PLUGINS: 1` is set in the same job's env |
| Rejected alternative (caps-probe-triggered pad request) loses the race 11/12 runs | EXTERNAL, self-reported | Author's local harness measurement; not reproducible from this repo, but the architectural reasoning (mp4mux refuses late request pads after stream start) is consistent with the code path taken |

**Diagnosis:** right layer — moves the pad-request decision to the one point where "will this track ever carry data" is answerable (post-link, pre-Playing). BOUNDED, not absolute: the PR body's "Known limitation" section states a track that negotiates caps and then produces no buffers still stalls — an honest, correctly-scoped disclosure.

**Blast radius:** LOCAL/SHARED, confined to `recorder.rs`, its build hook, and its tests. The CI env var addition doesn't leak into other tests' skip logic — `vision_mixer_fx_test.rs`/`shader_validation_test.rs` have their own independent `ElementFactory::find` checks that don't read `STROM_REQUIRE_GST_PLUGINS`.

**Tests & CI:** `Check (Linux)` pass with covering tests confirmed executed (not skipped) per log above. `Build (Linux ARM64/x86_64)`, `API Contract Check`, `Check & Build (WASM)`, `sccache preflight` all pass. `Build (Windows)`/`Build (macOS)` skip on `pull_request` by design (`workflow_dispatch`-gated) — expected, and this diff has no platform-specific code, so this doesn't block.

No CLAUDE.md violations found.

**Confidence: HIGH** — the fix mechanism, its one documented caveat, and the new production-path regression test are all confirmed by direct code reading, and CI logs independently corroborate the PR body's testing claims rather than merely repeating them.

Approve.

srperens:
## Verified locally and on a full-platform dispatch — one commit pushed, two issues filed

The recorder fix holds up. Verified on macOS (GStreamer 1.28.6) rather than on the description alone:

- **The guard is real.** Making `input_is_connected` return `true` unconditionally — the old "one pad per configured track" behaviour — fails exactly the two unfed arms with `pipeline never reached EOS within 30s`, while the both-tracks arms still pass. That is the original stall, not a proxy for it.
- `cargo test --workspace` green (506 unit plus all integration), `pipeline_lifecycle_test` green including its leak-detection self-test, both new probes are `EVENT_DOWNSTREAM`, and the `WeakRef` discipline in the element-setup closure is correct.

**On the `pending_links` caveat in the `input_is_connected` doc comment** — it is accurate, and I chased it because a silently dropped track would be nasty. It does not bite today: `try_link_elements_refs` resolves request pads (including tee `src_%u`) *synchronously* at construction, so only genuine "sometimes" pads from demuxers defer, and every block absorbs those internally behind an identity or tee. WHIP Input with `decode=false` was the obvious candidate; its external pads are `video_out_tee_0`, so it links immediately. And if a future block did break the assumption, the failure is the `warn!` path plus one lost track rather than a stall, which is a reasonable place to land.

## The CI half, and one commit I pushed

I dispatched `platforms=both` to exercise the `STROM_REQUIRE_GST_PLUGINS=1` addition, since Windows and macOS are skipped on PRs and it had therefore never run on either. Run [32952622615](https://github.com/Eyevinn/strom/actions/runs/32952622615), all eight jobs green.

**macOS: the variable works as intended.** All four recorder tests genuinely executed and passed.

**Windows: it found a real problem, twice over.**

1. The runner has no `x264enc`, so both recorder tests have been skipping green there. The variable did exactly its job and named the element. Filed as #703.
2. More seriously, the Windows job could not report that failure. The step logged `error: test failed` and still reported `success`, because `windows-latest` defaults `run:` to pwsh, where a native command's non-zero exit does not abort the script — the step's status comes from the last command, so the failing `cargo test --package strom` was masked by `cargo test --package strom-mcp-server` succeeding after it. The same pattern leaves the first command of `Run Clippy` and `Build` unchecked too. Pre-existing and nothing to do with this PR. Filed as #702.

So I pushed 66bacb7, which drops `STROM_REQUIRE_GST_PLUGINS=1` from the Windows job only. It has no effect there today, and leaving it would arm a landmine: the day #702 is fixed, the Windows job would start failing on a missing plugin rather than on the change someone was making. Linux and macOS keep it — proven working on both, and their steps run under `bash -e` so a failure is reported.

The `libblas3`/`liblapack3` reinstall is well justified and I have left it alone; the comment explaining that `cache-apt-pkgs-action` does not run postinst is exactly the kind of thing that saves the next person an afternoon.

One small thing for the description rather than the code: it says the variable is set "for the Linux test step", while the commit correctly says "everywhere tests run" and the diff touched three jobs. Worth aligning if you edit the body, since the three-job scope is what made the dispatch necessary.

srperens (approved):
Approving. Verification is in the comment above; the short version is that the guard fails under revert with the original 30s stall, the full workspace is green including `pipeline_lifecycle_test`, and the `pending_links` caveat in the doc comment does not bite today because request pads resolve synchronously at construction.

The only change from me is 66bacb7, which drops `STROM_REQUIRE_GST_PLUGINS=1` from the Windows job. That variable exposed two real problems on the platforms=both dispatch — a missing `x264enc` (#703) and a Windows job that cannot report a step failure at all (#702) — and both are now tracked. Linux and macOS keep it, verified working on both.

PR #677, opened 2026-08-20
feat(aes67): expose RTP payload type as a property on AES67 Output

srperens:
## Summary

The RTP payload type for AES67 Output was hardcoded to `96` in two places: the `pt` value implicitly used by the payloader, and the `m=`/`a=rtpmap:` lines in the generated SDP. This exposes it as a configurable block property so a stream can be matched against receivers that expect a specific dynamic payload type.

## Changes

- **`strom-types`**: new shared constant `DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE` (96).
- **AES67 Output block definition**: new `payload_type` exposed property (Int, default 96), placed after `ptime`. Renders automatically in the properties panel.
- **Builder**: reads the property and sets `pt` on `rtpL16pay`/`rtpL24pay`. Values outside 0-127 are rejected with `InvalidConfiguration`; values outside the dynamic range 96-127 are accepted but log a warning, since AES67 receivers may reject a non-dynamic payload type for L16/L24.
- **SDP generation**: reads the same property for the `m=audio ... RTP/AVP <pt>` and `a=rtpmap:<pt> ...` lines, falling back to 96 for out-of-range values so the SDP is never malformed.
- Two unit tests for the SDP side: a custom payload type (98), and the out-of-range fallback.

Note: the property is `live: false`, so a change requires a flow restart to take effect — consistent with the other AES67 Output properties.

## Test plan

- `cargo check` passes; pre-commit `cargo fmt` + `clippy` pass.
- Unit tests were **not** run locally (skipped at the author's request) — CI covers them.

### DIFF
--- a/backend/src/blocks/builtin/aes67.rs
+++ b/backend/src/blocks/builtin/aes67.rs
@@ -1297,4 +1297,51 @@ mod tests {
             "16-bit should allow longer ptime than 24-bit"
         );
     }
+
+    /// The builder must reject an RTP payload type outside the 7-bit range
+    /// (0-127) before it creates a single GStreamer element.
+    ///
+    /// This guards more than a malformed SDP. With the check removed, the build
+    /// walks on to `.property("pt", payload_type as u32)` on the payloader and
+    /// glib panics there — "property 'pt' of type 'GstRtpL24Pay' can't be set
+    /// from given value" — so the range check is what stops a stored
+    /// out-of-range value from taking the process down. GStreamer does not
+    /// clamp it.
+    ///
+    /// The variant and message are asserted rather than just `is_err()`, so a
+    /// rejection from a later check (packet size, bit depth) cannot stand in
+    /// for this one. `gst::init` is called because the panic above only
+    /// reproduces with GStreamer initialised; otherwise element creation aborts
+    /// on the uninitialised library first, which depends on whether another
+    /// test in this binary happened to initialise it.
+    #[test]
+    fn test_aes67_output_rejects_out_of_range_payload_type() {
+        let _ = gst::init();
+
+        let builder = AES67OutputBuilder;
+        let ctx = BlockBuildContext::new(Vec::new(), "all".to_string());
+
+        for pt in [-1, 128, 200] {
+            let mut properties = HashMap::new();
+            properties.insert("payload_type".to_string(), PropertyValue::Int(pt));
+
+            let result = builder.build("aes67-out-test", &properties, &ctx);
+
+            match result {
+                Err(BlockBuildError::InvalidConfiguration(msg)) => {
+                    assert!(
+                        msg.contains("payload type") && msg.contains(&pt.to_string()),
+                        "payload type {} was rejected, but not by the range check: {}",
+                        pt,
+                        msg
+                    );
+                }
+                Err(other) => panic!(
+                    "payload type {} was rejected by the wrong check: {:?}",
+                    pt, other
+                ),
+                Ok(_) => panic!("payload type {} was accepted", pt),
+            }
+        }
+    }
 }
--- a/backend/src/blocks/builtin/aes67.rs
+++ b/backend/src/blocks/builtin/aes67.rs
@@ -19,6 +19,8 @@ const AES67_INPUT_DEFAULT_BUFFER_DURATION_MS: i64 =
 // AES67 Output defaults
 const AES67_OUTPUT_DEFAULT_TTL: i64 = 32;
 const AES67_OUTPUT_DEFAULT_QOS_DSCP: &str = "0x2E"; // DSCP EF (Expedited Forwarding) per AES67 standard
+const AES67_OUTPUT_DEFAULT_PAYLOAD_TYPE: i64 =
+    strom_types::block::DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE;
 
 /// Parse DSCP value from string (hex like "0x2E" or "disabled")
 /// Returns the integer value for udpsink qos-dscp property
@@ -663,6 +665,30 @@ impl BlockBuilder for AES67OutputBuilder {
             })
             .unwrap_or_else(|| parse_dscp_value(AES67_OUTPUT_DEFAULT_QOS_DSCP));
 
+        let payload_type = properties
+            .get("payload_type")
+            .and_then(|v| match v {
+                PropertyValue::Int(i) => Some(*i),
+                PropertyValue::String(s) => s.parse::<i64>().ok(),
+                _ => None,
+            })
+            .unwrap_or(AES67_OUTPUT_DEFAULT_PAYLOAD_TYPE);
+
+        if !(0..=127).contains(&payload_type) {
+            return Err(BlockBuildError::InvalidConfiguration(format!(
+                "Invalid RTP payload type: {}. Must be between 0 and 127.",
+                payload_type
+            )));
+        }
+
+        if !(96..=127).contains(&payload_type) {
+            warn!(
+                "AES67 Output [{}]: Payload type {} is outside the dynamic range (96-127). \
+                 AES67 requires a dynamic payload type for L16/L24 streams; receivers may reject the stream.",
+                instance_id, payload_type
+            );
+        }
+
         // Validate packet size fits within AES67/Ethernet MTU constraints
         // RTP payload must fit in ~1440 bytes (1500 MTU - 20 IP - 8 UDP - 12 RTP - ~20 safety margin)
         // Payload size = framecount × channels × bytes_per_sample
@@ -730,6 +756,7 @@ impl BlockBuilder for AES67OutputBuilder {
         let payloader = gst::ElementFactory::make(payloader_type)
             .name(&payloader_id)
             .property("timestamp-offset", 0u32)
+            .property("pt", payload_type as u32)
             .property("min-ptime", ptime_ns)
             .property("max-ptime", ptime_ns)
             .build()
@@ -762,8 +789,8 @@ impl BlockBuilder for AES67OutputBuilder {
         }
 
         debug!(
-            "AES67 Output [{}]: Multicast TTL={}, QoS DSCP={}",
-            instance_id, ttl, qos_dscp
+            "AES67 Output [{}]: Multicast TTL={}, QoS DSCP={}, payload type={}",
+            instance_id, ttl, qos_dscp, payload_type
         );
 
         let udpsink = udpsink_builder
@@ -1023,6 +1050,20 @@ fn aes67_output_definition() -> BlockDefinition {
                 mapping: PropertyMapping {
                     element_id: "_block".to_string(),
                     property_name: "ptime".to_string(),
+                    transform: None,
+                },
+                live: false,
+                persist: None,
+            },
+            ExposedProperty {
+                name: "payload_type".to_string(),
+                label: "Payload Type".to_string(),
+                description: "RTP payload type used in the packets and announced in the SDP. Must be in the dynamic range 96-127 for AES67.".to_string(),
+                property_type: PropertyType::Int,
+                default_value: Some(PropertyValue::Int(AES67_OUTPUT_DEFAULT_PAYLOAD_TYPE)),
+                mapping: PropertyMapping {
+                    element_id: "_block".to_string(),
+                    property_name: "payload_type".to_string(),
                     transform: None,
                 },
                 live: false,
--- a/backend/src/blocks/sdp.rs
+++ b/backend/src/blocks/sdp.rs
@@ -257,7 +257,18 @@ pub fn generate_aes67_output_sdp(
     // Use provided values or fall back to AES67 defaults: 48kHz, 2 channels
     let sample_rate = sample_rate.unwrap_or(48000);
     let channels = channels.unwrap_or(2);
-    let payload_type = 96; // Dynamic payload type
+
+    // RTP payload type - must match the "pt" property set on the payloader
+    let payload_type = block
+        .properties
+        .get("payload_type")
+        .and_then(|v| match v {
+            PropertyValue::Int(i) => Some(*i),
+            PropertyValue::String(s) => s.parse::<i64>().ok(),
+            _ => None,
+        })
+        .filter(|pt| (0..=127).contains(pt))
+        .unwrap_or(strom_types::block::DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE);
 
     // Determine encoding name based on bit depth
     let encoding = match bit_depth {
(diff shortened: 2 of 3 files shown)


srperens (commented):
**Intent:** Expose the RTP payload type (`pt`) on the AES67 Output block as a configurable property, defaulting to 96 (the first dynamic payload type per RFC 3551). The same value is applied to both the GStreamer payloader and the generated SDP, so they stay in sync.

**Risk:** Low. The default is unchanged (96), the property is `live: false` so no runtime pipeline mutation occurs, and the builder validates the range before any element is created.

---

**What looks good:**

- Adding `DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE` to `strom-types` follows the shared types convention.
- The 0–127 hard rejection + 96–127 warning is the correct two-level treatment per RFC 3551/AES67.
- SDP fallback to `DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE` when the stored value is out of range prevents a malformed SDP from ever leaving the system, even if something bypasses the builder validation.
- The two unit tests cover the key paths (custom value, out-of-range fallback).
- `live: false` is consistent with all other AES67 Output properties and accurately reflects the restart requirement.

**One minor note:**

Both the builder and `generate_aes67_output_sdp` include a `PropertyValue::String(s) => s.parse::<i64>().ok()` arm. Since `payload_type` is declared as `PropertyType::Int`, a `String` variant can only arrive here if something stores the wrong type — that situation probably deserves a log line rather than silent acceptance. Consider replacing the `String` arm with `_ => None` in both sites so unexpected variants surface early. This is non-blocking.

Overall this is a clean, well-scoped addition.

srperens (commented):
Reviewing PR #677: feat(aes67): expose RTP payload type as a property on AES67 Output
Verified against 88c8acb3 vs main 317e80b1.


**Note:** This PR is authored by the same account (`srperens`) this review runs as — GitHub rejects self-approval via the API. The verdict below is APPROVE-equivalent (every claim CONFIRMED, CI green with tests executed, blast radius LOCAL); a human maintainer should apply the formal approval.
**Summary**
- Adds a `payload_type` exposed property (default 96) to the AES67 Output block, replacing a hardcoded `pt=96` used both on `rtpL16pay`/`rtpL24pay` and in generated SDP.
- Builder rejects values outside 0-127; warns (does not reject) for values outside the AES67-required dynamic range 96-127.
- SDP generation reads the same property and falls back to 96 if the stored value is outside 0-127.

**Claim verification**
|---|---|---|
| New shared constant `DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE` lives in `strom-types` | CONFIRMED | `types/src/block.rs:421` `pub const DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE: i64 = 96;` — placed correctly per CLAUDE.md's "new shared constants belong in strom-types" rule |
| Builder rejects payload types outside 0-127 with `InvalidConfiguration` | CONFIRMED | `backend/src/blocks/builtin/aes67.rs`, `AES67OutputBuilder::build`, `if !(0..=127).contains(&payload_type) { return Err(BlockBuildError::InvalidConfiguration(...)) }` |
| Values outside dynamic range 96-127 are accepted but logged as a warning | CONFIRMED | same function, `if !(96..=127).contains(&payload_type) { warn!(...) }` — quoted: `"Payload type {} is outside the dynamic range (96-127)..."` |
| The `pt` property is set on the payloader (`rtpL16pay`/`rtpL24pay`) | CONFIRMED | `backend/src/blocks/builtin/aes67.rs`, payloader builder chain: `.property("pt", payload_type as u32)` |
| SDP generation reads the same property and falls back to 96 for out-of-range values | CONFIRMED | `backend/src/blocks/sdp.rs`, `generate_aes67_output_sdp`, `.filter(|pt| (0..=127).contains(pt)).unwrap_or(strom_types::block::DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE)` |
| Two new unit tests cover custom payload type (98) and out-of-range fallback (200→96) | CONFIRMED | `backend/src/blocks/sdp.rs`, `test_generate_sdp_custom_payload_type` asserts `"m=audio 5004 RTP/AVP 98"` / `"a=rtpmap:98 L24/48000/2"`; `test_generate_sdp_out_of_range_payload_type_falls_back` asserts fallback to `96` |
| `payload_type` property is `live: false`, "consistent with the other AES67 Output properties" | CONFIRMED | `backend/src/blocks/builtin/aes67.rs`, `aes67_output_definition()` — every `ExposedProperty` in this function (19 occurrences) sets `live: false`, including the new one |
| "Renders automatically in the properties panel" (no frontend change needed) | CONFIRMED | `frontend/src/properties.rs:2519-2529` `convert_block_prop_type` maps `block::PropertyType::Int` (a unit variant, no min/max) to `None`, which routes into the generic `(PropertyValue::Int(i), _) => ui.add(egui::DragValue::new(i))` arm at line ~2864 — no per-block frontend code exists or is needed for a plain Int property |
| "Unit tests were not run locally... CI covers them" | CONFIRMED (CI ran them) | `.github/workflows/ci.yml` `Check (Linux)` job runs `cargo test --package strom --features efp,nvidia` (line 111), and `backend/Cargo.toml:2` names the backend package `strom`, so `backend/src/blocks/sdp.rs`'s tests are included in that run |

**Diagnosis**
Right layer and complete: the property is threaded through both places that previously hardcoded `96` — the payloader's `pt` and the SDP `m=`/`a=rtpmap:` lines — so there's no leftover hardcoded path that could desync payload type between the actual RTP stream and its SDP description. ABSOLUTE within the block: any value 0-127 is honored on both sides consistently; BOUNDED by design in one place worth flagging as an intentional inconsistency rather than a bug — the builder accepts (with a warning) static payload types 0-95, and the SDP side mirrors that (no forced fallback for 0-95, only for out-of-127 values), so behavior is consistent between builder and SDP for the full 0-127 range, not just 96-127. No other trigger of the "payload type wrong in SDP" symptom class remains, since this was the only hardcoded `96` in the AES67 Output path (confirmed by the diff being the only changed hunk touching `payload_type` in `sdp.rs`).

**Blast radius**
LOCAL. Confirmed via `grep -rn "generate_aes67_output_sdp"`: two call sites, `backend/src/api/flows.rs:874` (an SDP-preview API endpoint that reads stored block config without requiring a running pipeline) and `backend/src/state.rs:1123` (used when a flow is actually running). Both pass `block` straight through unchanged; neither needed modification since the function signature didn't change. `grep` for `DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE` and `payload_type` in `backend/src/blocks/builtin/aes67.rs` confirms AES67 **Input** (a separate builder in the same file) has no references to the new property or constant — the change doesn't touch the input side. Worth noting for the maintainer (not a bug): because `flows.rs:874`'s SDP-preview endpoint can be called on unbuilt/unvalidated flow config, a stored `payload_type` of e.g. 200 would render as SDP `pt=96` (silent fallback) even though attempting to actually start that flow would fail at build time with `InvalidConfiguration`. This is consistent with the PR's explicit design goal ("SDP is never malformed") and not a functional bug, just a preview/build asymmetry.

**Tests & CI**
- Checks on 88c8acb3 (run 32363315350): `Build (Linux ARM64)` pass, `Build (Linux x86_64)` pass, `Check (Linux)` pass, `Check & Build (WASM)` pass, `API Contract Check` pass, `sccache preflight` pass, `Build (Windows)`/`Build (macOS)` skipping (expected, `workflow_dispatch`-only).
- Both new tests exercise the actual changed function (`generate_aes67_output_sdp`) with real `BlockInstance`/`PropertyValue` inputs and assert on the generated SDP string — not a hand-rebuilt topology, so they would fail if the fix were reverted (reverting to hardcoded `96` breaks `test_generate_sdp_custom_payload_type`'s assertion of `pt=98`). This guards the change, not just documents it.
- No skip guard in either new test (pure string logic, no `ElementFactory::find` gate), and `Check (Linux)` runs `cargo test --package strom`, which includes them — confirmed executed, not just green.
- No platform-specific code in this diff; platform coverage is not applicable.
- No new/changed API endpoint or `StromEvent` variant, so the openapi-snapshot and WebSocket-contract CLAUDE.md rules don't apply here — `API Contract Check` passing is consistent with that (no schema change expected).

**Risks / edge cases**
- None found beyond the preview/build asymmetry noted under Blast radius, which is an intentional, stated design choice, not an oversight.

**Requested changes**
None.

**Confidence: HIGH**

srperens (commented):
Reviewing PR #677: feat(aes67): expose RTP payload type as a property on AES67 Output
Verified against 88c8acb3 vs main d7378bb9.

**Note:** the reviewing credential and the PR author (`srperens`) are the same GitHub account in this environment. I did not write this diff (fresh session, no memory of prior work), but I'm capping the verdict at Comment rather than Approve regardless of findings, since self-approval isn't a legitimate signal.

**Summary**
- Adds a `payload_type` exposed Int property (default 96) to the AES67 Output block, replacing a hardcoded `pt`/SDP payload type of 96.
- Builder sets `pt` on the `rtpL16pay`/`rtpL24pay` payloader from the property; rejects values outside 0-127, warns (but allows) values outside the AES67 dynamic range 96-127.
- SDP generation (`m=`/`a=rtpmap:` lines) reads the same property, with a silent fallback to 96 if the stored value is somehow outside 0-127.
- Adds two unit tests for the SDP side (custom pt=98, out-of-range fallback).

**Claim verification**
|---|---|---|
| pt was hardcoded to 96 for the payloader (implicitly, via element default) | EXTERNAL | `rtpL16pay`/`rtpL24pay` are GStreamer `gst-plugins-good` elements inheriting `GstRTPBasePayload`; I did not build/inspect the element to confirm its default `pt` is 96 — assuming upstream default per RFC 3551 dynamic range convention. |
| SDP payload type was hardcoded to 96 | CONFIRMED | `backend/src/blocks/sdp.rs:260` (pre-diff, per PR diff hunk) had `let payload_type = 96; // Dynamic payload type`, now replaced by property read at `backend/src/blocks/sdp.rs:262` `let payload_type = block`. |
| New property defaults to 96 and is `live: false` (flow restart required) | CONFIRMED | `backend/src/blocks/builtin/aes67.rs:1067` area — `ExposedProperty { name: "payload_type"... default_value: Some(PropertyValue::Int(AES67_OUTPUT_DEFAULT_PAYLOAD_TYPE))... live: false, persist: None }`. |
| Builder rejects payload_type outside 0-127 | CONFIRMED | `backend/src/blocks/builtin/aes67.rs:679` `"Invalid RTP payload type: {}. Must be between 0 and 127."` inside `if !(0..=127).contains(&payload_type)`. |
| Builder warns (does not fail) for values outside 96-127 | CONFIRMED | `backend/src/blocks/builtin/aes67.rs:686` `"AES67 Output [{}]: Payload type {} is outside the dynamic range (96-127)."` inside a `warn!` call, no early return. |
| SDP falls back to 96 for out-of-range values instead of producing malformed SDP | CONFIRMED | `backend/src/blocks/sdp.rs:262-271` `.filter(|pt| (0..=127).contains(pt)).unwrap_or(strom_types::block::DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE)`; guarded by test `backend/src/blocks/sdp.rs:468` `test_generate_sdp_out_of_range_payload_type_falls_back`. |
| New shared constant lives in `strom-types` | CONFIRMED | `types/src/block.rs:421` `pub const DEFAULT_AES67_OUTPUT_PAYLOAD_TYPE: i64 = 96;` — matches CLAUDE.md's "Shared Types" rule (shared constants belong in `strom-types`, not the backend). |

**Diagnosis**
Right fix, not symptom relief — the value now has exactly one source of truth (the block property), consumed identically by the payloader (`backend/src/blocks/builtin/aes67.rs:759` `.property("pt", payload_type as u32)`) and by SDP generation (`backend/src/blocks/sdp.rs:262`), so the two can no longer drift out of sync. ABSOLUTE within the declared 0-127 RTP payload-type space; the builder-side validation is a hard reject outside that range, and the SDP-side fallback is defense-in-depth for a state that validation should already prevent, not a new tolerance. No other trigger of the same symptom is left open — I found no third place that references AES67 output payload type (`grep -rn "aes67_output\|AES67_OUTPUT" backend/src` shows only `aes67.rs`, `sdp.rs`, `state.rs`, `flows.rs`, and `stats/collector.rs`, and the latter two only match on `block_definition_id` string, not `pt`).

**Blast radius**
LOCAL. Call sites for `generate_aes67_output_sdp` are `backend/src/state.rs` (flow-start SDP generation, reads `block.properties` for the same `BlockInstance` the builder just validated) and `backend/src/api/flows.rs:801` (`GET` SDP endpoint, same pattern) — both pass the live `block` object, so both pick up the new property automatically with no separate wiring needed. No other block type or shared helper is touched. `openapi.json` has no diff (`git diff origin/main...pr677 -- openapi.json` empty), consistent with this being a block-property addition rather than a new/changed endpoint — CLAUDE.md's utoipa/openapi-registration rule doesn't apply here.

**Tests & CI**
- `Check (Linux)` — pass. Runs `cargo test --package strom --features efp,nvidia` per `.github/workflows/ci.yml:111`, which includes the two new `sdp.rs` tests; they are plain string-assertion unit tests with no GStreamer-element skip guard, so they actually executed (not just green-by-skip).
- `Build (Linux ARM64)` / `Build (Linux x86_64)` — pass.
- `Build (Windows)` / `Build (macOS)` — skipping, expected: this diff touches no platform-gated code (`aes67.rs`, `sdp.rs`, `types/src/block.rs` are all platform-neutral), so no platform dispatch is needed here.
- `API Contract Check` — pass (no openapi.json change expected or required).
- `Check & Build (WASM)` — pass.
- Test-guards-the-fix: `test_generate_sdp_custom_payload_type` (`backend/src/blocks/sdp.rs:446`) and `test_generate_sdp_out_of_range_payload_type_falls_back` (`backend/src/blocks/sdp.rs:468`) both call `generate_aes67_output_sdp` directly with a `payload_type` property set on a real `BlockInstance`, and assert on the generated SDP text — reverting the `sdp.rs` fix (reintroducing the hardcoded `96`) would fail `test_generate_sdp_custom_payload_type`. These guard the change, not just demonstrate it.
- No test exists for the builder-side validation/warning path (`aes67.rs:679` reject, `aes67.rs:686` warn) — gap noted below.
- Canaries: `pipeline_lifecycle_test.rs` not relevant (no closures, no pipeline/element capture added). `openapi_test` not relevant (no endpoint change).

**Risks / edge cases**
- SPECULATIVE (not verified): the upstream `rtpL16pay`/`rtpL24pay` `pt` property range — I'm assuming it accepts the full 0-127 u32 without its own internal clamping that could silently diverge from the value used in the emitted SDP text if GStreamer itself rejects a value the block-level validation allowed. Builder-level validation is 0-127 which should match RTP's 7-bit pt field, so this is a low-risk gap, not a functional one.
- Gap: no unit test for `BlockBuildError::InvalidConfiguration` on `payload_type > 127` or `< 0`, nor for the `warn!` path on e.g. `payload_type = 50` (valid range, non-dynamic). Both are pure-Rust and could run in CI without any GStreamer element (the reject path doesn't even reach `ElementFactory::make`).

**Requested changes**
1. Add a unit test in `backend/src/blocks/builtin/aes67.rs`'s existing `mod tests` (`backend/src/blocks/builtin/aes67.rs:1220`) asserting `payload_type: 200` (or similar) returns `Err(BlockBuildError::InvalidConfiguration(_))` from the builder. This is the one validation path in this PR with no test coverage, and it's cheap to add since it doesn't require constructing a live pipeline.

**Confidence: HIGH**

srperens:
Requested change addressed in e643071: `test_aes67_output_rejects_out_of_range_payload_type` in `aes67.rs`'s `mod tests` asserts `InvalidConfiguration` from the builder for `payload_type` of -1, 128 and 200.

Two things the mutation check turned up, both of which change how this validation should be read:

**The check prevents a panic, not just a bad SDP.** With the `0..=127` rejection removed, the build walks on to `.property("pt", payload_type as u32)` on the payloader and glib panics: `property 'pt' of type 'GstRtpL24Pay' can't be set from given value, it is invalid or out of range`. So this answers the SPECULATIVE note in the review about whether GStreamer clamps a value the block-level validation let through — **it does not clamp, it aborts**. The range check is load-bearing.

**A bare `is_err()` would have guarded nothing.** The rejection happens before any `ElementFactory::make`, so a version without it fails at element creation instead — a different error, and with GStreamer uninitialised a panic rather than an `Err`. The test therefore matches on the variant and the message, and calls `let _ = gst::init()` so the reverted-code failure mode doesn't depend on which test in the binary initialised GStreamer first. Verified both directions: passes on the branch, fails with the check removed.

**Ran locally:** `cargo test --features efp` — 512 lib tests pass, including the new one (confirmed executed by name, not green-by-skip). `cargo clippy --all-targets --features efp -- -D warnings` clean. One unrelated pre-existing flake: `wipe_between_letterboxed_sources_animates` in `vision_mixer_fx_test.rs` failed on a frame-sampling assertion in a full-suite run and passed when its target ran alone; it's a timing-sensitive GL test and untouched by this diff.

**Not addressed**, both from the earlier review and non-blocking:

1. The `PropertyValue::String(s) => s.parse::<i64>().ok()` arm on a property declared `PropertyType::Int` — replacing it with `_ => None` in the builder and in `generate_aes67_output_sdp` would surface a wrong stored type instead of quietly accepting it. Note the same `String` arm exists on `bit_depth`, `sample_rate` and others in this builder, so it's a pre-existing convention in this file rather than something this PR introduced; changing it here alone would make the file inconsistent. Worth doing as its own PR across the block, if at all.
2. No test for the `warn!` path (a valid-range but non-dynamic value like 50). It requires either capturing tracing output or building real elements, since execution continues past the warning — not worth the machinery for a log line.

PR #678, opened 2026-08-21
ci: run the tests that were silently skipping, and make platform builds selectable

srperens:
## Problem

Two gaps found while reviewing #675.

**Tests were skipping silently.** A test that needs a GStreamer element outside
`plugins-base`/`plugins-good` printed a skip message, returned `Ok`, and the job went green.
`backend/tests/recorder_splitmux_threading_test.rs` (added in #675) is one such case: it needs
`x264enc`, `avenc_aac`, `tsdemux` and `mpegtsmux`, none of which were installed, so it would
have guarded nothing while reporting success.

**Platform-specific code is never built on its platform.** The macOS and Windows jobs are
gated on `workflow_dispatch` because they are expensive, and triggering them was
all-or-nothing. #669 (Cocoa run loop) and #668 (autovideoconvert on macOS) are open right now,
both changing macOS-only code, and neither has been built on macOS.

## Changes

**`chore(ci)`** — add `gstreamer1.0-plugins-bad` (tsdemux, mpegtsmux),
`gstreamer1.0-plugins-ugly` (x264enc) and `gstreamer1.0-libav` (avenc_aac) to all three Linux
package lists, plus `gst-plugins-ugly` / `gst-libav` on macOS. Package ownership was confirmed
with `dpkg -S` against the installed plugin libraries rather than assumed. The three Linux
lists stay byte-identical so they keep sharing one apt cache entry, as the comment above them
requires.

Also adds a `## Tests` section to `CLAUDE.md`: a regression test must call the code it guards,
must fail if the fix is reverted, must be able to run in CI, and the PR body must say which
tests actually ran.

**`ci`** — add a `platforms` choice input (`both` / `macos` / `windows` / `none`) to
`workflow_dispatch` and gate each platform job on it:

```
gh workflow run ci.yml --ref <branch> -f platforms=macos
```

The default stays `both`, so an argument-less dispatch behaves exactly as before, and
`inputs.platforms` is empty on `push` and `pull_request`, so those events are unchanged.

## Verification

`ci.yml` was parsed with `yaml.safe_load` to confirm the new input and both `if:` expressions
are well-formed; the three Linux package lists were diffed against each other to confirm they
remain identical. Element-to-package mapping was verified locally:

| element | plugin library | package |
|---|---|---|
| `x264enc` | `libgstx264.so` | `gstreamer1.0-plugins-ugly` |
| `avenc_aac` | `libgstlibav.so` | `gstreamer1.0-libav` |
| `mpegtsmux` / `tsdemux` | `libgstmpegtsmux.so` / `libgstmpegtsdemux.so` | `gstreamer1.0-plugins-bad` |
| `splitmuxsink` | `libgstmultifile.so` | `gstreamer1.0-plugins-good` (already installed) |

Not verified: that CI actually installs these on the runner and that a previously-skipping test
now executes — that needs this PR's own CI run, since #675's test is not on `main` yet. The
platform selector is likewise unverified until someone dispatches it against this branch.

## Trade-offs worth a look

- The new packages grow the apt cache entry (`libav` pulls in the ffmpeg libraries), and
  changing the package list creates a new cache key, so the first run repopulates it. Worth
  watching against the repo's GHA cache ceiling.
- `plugins-ugly` brings x264, which is GPL. It is installed only in the CI test environment and
  is not distributed, but flagging it in case that matters.
- The platform jobs still need `preflight` and `check-wasm` — the macOS job downloads the
  `frontend-dist` artifact `check-wasm` produces — so a dispatch cannot be the platform job
  alone.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,12 @@ on:
   pull_request:
     branches: [main]
   workflow_dispatch:
+    inputs:
+      platforms:
+        description: "Which extra platform builds to run (they are off for push/PR because of cost)"
+        type: choice
+        default: both
+        options: [both, macos, windows, none]
 
 env:
   CARGO_TERM_COLOR: always
@@ -428,7 +434,7 @@ jobs:
   # Only runs manually (workflow_dispatch) to save CI cost (2× multiplier).
   build-windows:
     name: Build (Windows)
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' && contains(fromJSON('["both","windows"]'), inputs.platforms)
     needs: [check-wasm, preflight]
     runs-on: windows-latest
     timeout-minutes: 45
@@ -519,7 +525,7 @@ jobs:
   # Only runs manually (workflow_dispatch) to save CI cost (10× multiplier).
   build-macos:
     name: Build (macOS)
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' && contains(fromJSON('["both","macos"]'), inputs.platforms)
     needs: [check-wasm, preflight]
     runs-on: macos-latest
     timeout-minutes: 30
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Cache apt packages
         uses: awalsh128/cache-apt-pkgs-action@v1.6.1
         with:
-          packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good pkg-config cmake libclang-dev
+          packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
       # save-if main: PR runs restore the main-branch cache (GHA cache scoping
@@ -282,7 +282,7 @@ jobs:
       - name: Install GStreamer (cached)
         uses: awalsh128/cache-apt-pkgs-action@v1.6.1
         with:
-          packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good pkg-config cmake libclang-dev
+          packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
       - name: Download frontend dist
@@ -392,7 +392,7 @@ jobs:
       - name: Install GStreamer (cached)
         uses: awalsh128/cache-apt-pkgs-action@v1.6.1
         with:
-          packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good pkg-config cmake libclang-dev
+          packages: libunwind-dev libssl-dev libcairo2-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav pkg-config cmake libclang-dev
           version: 1.5
 
       - name: Download frontend dist
@@ -541,7 +541,7 @@ jobs:
 
       - name: Install GStreamer
         run: |
-          brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad
+          brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly gst-libav
           echo "PKG_CONFIG_PATH=$(brew --prefix)/lib/pkgconfig:$(brew --prefix gstreamer)/lib/pkgconfig:$(brew --prefix glib)/lib/pkgconfig" >> $GITHUB_ENV
           echo "LIBRARY_PATH=$(brew --prefix)/lib" >> $GITHUB_ENV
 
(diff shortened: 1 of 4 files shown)


srperens (commented):
Reviewing PR #678: ci: run the tests that were silently skipping, and make platform builds selectable
Verified against 6ff42c98 vs main 317e80b1.

**Summary**
- Adds `gstreamer1.0-plugins-bad`, `gstreamer1.0-plugins-ugly`, `gstreamer1.0-libav` to all three Linux CI package lists (and the macOS brew list) so tests requiring `x264enc`/`avenc_aac`/`tsdemux`/`mpegtsmux` actually run instead of skipping.
- Adds a `platforms` choice input to `workflow_dispatch` so macOS/Windows builds can be dispatched individually instead of only both-or-nothing.
- Adds a `## Tests` section to `CLAUDE.md` codifying that a regression test must call the changed code, must fail if the fix is reverted, must be runnable in CI, and the PR body must state what actually ran.

**Claim verification**
|---|---|---|
| `recorder_splitmux_threading_test.rs` (from #675) needs `x264enc`, `avenc_aac`, `tsdemux`, `mpegtsmux` and would have skipped silently | CONFIRMED | `backend/tests/recorder_splitmux_threading_test.rs` (PR #675 branch) lines 35-38 list exactly `"x264enc", "avenc_aac", "mpegtsmux", "tsdemux"` as required elements, checked via `gst::ElementFactory::find(e).is_none()` (line 48) |
| The three Linux apt package lists stay byte-identical after the change | CONFIRMED | `.github/workflows/ci.yml:89,291,401` (pr678 branch) — all three `packages:` lines are character-for-character identical, matching the comment at line 85: "Package list must stay identical to build-linux-x86_64 so both jobs share one apt cache entry" |
| `workflow_dispatch` gains a `platforms` choice input (`both`/`macos`/`windows`/`none`), default `both` | CONFIRMED | `.github/workflows/ci.yml` `on.workflow_dispatch.inputs.platforms`, parsed with `yaml.safe_load`: `{'type': 'choice', 'default': 'both', 'options': ['both', 'macos', 'windows', 'none']}` |
| `build-macos`/`build-windows` gating leaves `push`/`pull_request` behavior unchanged | CONFIRMED | `.github/workflows/ci.yml` `jobs.build-macos.if` = `` github.event_name == 'workflow_dispatch' && contains(fromJSON('["both","macos"]'), inputs.platforms) `` — the `github.event_name == 'workflow_dispatch'` guard short-circuits false on `push`/`pull_request` regardless of the `inputs.platforms` value |
| `x264enc`→`gstreamer1.0-plugins-ugly`, `avenc_aac`→`gstreamer1.0-libav`, `mpegtsmux`/`tsdemux`→`gstreamer1.0-plugins-bad` package mapping | EXTERNAL | Standard Debian/Ubuntu GStreamer packaging split; this sandbox has no apt cache to independently confirm via `dpkg -S`, so I'm taking the PR's stated `dpkg -S` verification at face value |
| New `CLAUDE.md` "Tests" section text | CONFIRMED | `CLAUDE.md:60-64` (pr678 branch), added verbatim as shown in the diff, no conflict with any pre-existing section |

**Diagnosis**
This is process/infrastructure hardening, not a bug fix with a "mechanism" to diagnose in the usual sense. It's the right layer: a test that silently no-ops on a missing element is a CI-configuration gap, and the fix is exactly to close that gap in CI config, not to change the test. ABSOLUTE within its stated scope (the four newly-required elements are now installed on all Linux jobs and macOS), BOUNDED in that it only covers elements this PR's author identified as currently needed — any future test needing a plugin from an as-yet-uninstalled package (e.g. `gstreamer1.0-plugins-bad` extras not yet pulled, VA-API, etc.) will still skip silently until someone notices, which is exactly the failure mode the new `CLAUDE.md` Tests section is meant to catch going forward by making "state which tests ran" a PR-body requirement rather than a CI guarantee.

**Blast radius**
LOCAL/SHARED split:
- `.github/workflows/ci.yml` changes are SHARED — they affect every future CI run (larger apt cache footprint, GPL `plugins-ugly` now present in the Linux test images, as the PR body itself flags). No product code is touched.
- `CLAUDE.md` Tests section is a repo-wide process rule, not code, so no call sites to check.
- I did not find any other workflow file or script that reads the removed/added package list strings, so no other consumer is affected.

**Tests & CI**
- Checks on 6ff42c98: `Check (Linux)` — **fail**; `Build (Linux ARM64)` — **fail**; `Build (Linux x86_64)` — pending at last check; `API Contract Check` — pass; `Check & Build (WASM)` — pass; `sccache preflight` — pass; `Build (Windows)`/`Build (macOS)` — skipping (expected, `workflow_dispatch` only).
- `Check (Linux)` failure is a `clippy::result_large_err` deny at `backend/src/api/whip_ingest.rs:383`, function `forward_whip_post` (`async fn forward_whip_post(...) -> Result<(...), Response>`), quoted from the job log: `"the `Err`-variant returned from this function is very large" ... "the `Err`-variant is at least 128 bytes"`. This PR's diff touches zero `.rs` files (`git diff origin/main...pr678 --stat` shows only `.github/workflows/ci.yml` and `CLAUDE.md`), and PR #677's CI run against the same `main` lineage passed `Check (Linux)` cleanly a few hours earlier — this is toolchain drift (the `dtolnay/rust-toolchain@stable` float picked up a newer clippy between those runs), not something introduced by this diff. Provably unrelated by diff content.
- `Build (Linux ARM64)` failure is a separate, also diff-unrelated issue: `error: unsupported linker arg: --fix-cortex-a53-843419` from the `cargo-zigbuild` cross-linker, hit while compiling `gst-plugin-audiofx`, `gst-plugin-rtp`, `gst-plugin-webrtc`, `strom-frontend`. This is a zig/zigbuild version mismatch in the cross-compilation toolchain, unrelated to any package list or workflow-dispatch change in this diff.
- Both failures are red but not attributable to this PR's changes — they reflect that `main`'s CI toolchain has drifted since the last recorded green run (2026-08-19, `317e80b1`, `success`). This blocks a clean "CI is green" story for every open PR right now, not just this one, and is worth a maintainer's attention independent of this review.
- No test in this PR's own diff to guard (it's CI config + docs), so "does the test call the changed code" doesn't apply here. The change itself can only be validated by a live CI run reading the new package lists — which is exactly what's happening (and failing on unrelated grounds) above.
- Platform coverage: this PR changes the `if:` gating for `build-macos`/`build-windows`, but per the skip-guard (`github.event_name == 'workflow_dispatch'`), neither job runs on `push`/`pull_request`, so the new `platforms` selector itself is UNVERIFIED by any automatic run. To verify the selector actually restricts to one platform: `gh workflow run ci.yml --ref chore/ci-test-deps-and-test-rules -f platforms=macos` (and separately `-f platforms=windows`).

**Risks / edge cases**
- `gstreamer1.0-plugins-ugly` pulls x264 (GPL) into the CI image. PR body already flags this as CI-only, not distributed — worth the maintainer's explicit sign-off since it's a licensing-adjacent decision, not a pure engineering one.
- New packages change the apt cache key (per the in-file comment on cache scoping), so the first run after merge will be slower while the cache repopulates — cosmetic, self-resolving.
- SPECULATIVE (not verified): the ARM64 zigbuild linker failure could recur on every future PR until someone pins or upgrades zig/zigbuild — I did not check whether this is a known, already-tracked issue elsewhere in the repo or CI history.

**Requested changes**
1. None to this diff itself — it's correctly scoped and both new claims I could check (package-list identity, YAML gating logic) are confirmed. Before merging, the maintainer should re-run CI (or dispatch `-f platforms=macos`) once the unrelated `Check (Linux)` and `Build (Linux ARM64)` failures are resolved on `main`, since neither is fixable from within this PR's scope (the `whip_ingest.rs` lint and the zigbuild linker are pre-existing, unrelated to this diff).

**Confidence: MED** — the diff itself is small and every in-scope claim is confirmed, but I cannot approve while the actual CI run on this head SHA is red, even though I've traced both failures to causes outside the diff. A green re-run (or an explicit maintainer call that these are known, tracked, unrelated flakes) would move this to HIGH/approve.

PR #681, opened 2026-08-21
fix(api): honour the client-supplied flow id on POST /api/flows

### LINKED ISSUE #672: POST /api/flows requires an `id` but silently assigns a different one
### Summary

`POST /api/flows` requires an `id` in the request body (the schema marks it required, and omitting it returns `422 missing field id`), but the value is discarded and a new UUID is assigned.

### Reproduce

```bash
curl -s -X POST localhost:8080/api/flows -H 'Content-Type: application/json' -d '{
  "id": "11111111-2222-3333-4444-555555555555", "name": "idcheck",
  "elements": [{"id":"vts","element_type":"videotestsrc","properties":{},"position":[0.0,0.0]},
               {"id":"fs","element_type":"fakesink","properties":{},"position":[300.0,0.0]}],
  "blocks": [], "links": [{"from":"vts:src","to":"fs:sink"}]}'
# → 201

curl -s localhost:8080/api/flows | jq -r '.[] | select(.name=="idcheck") | .id'
# → a52be894-684d-4360-ad49-6816a0477308
```

Supplied `11111111-…`, stored `a52be894-…`.

### Why it matters

The natural client pattern — generate an id, create, then `POST /api/flows/{id}/start` — fails with a confusing error, because the id you supplied never existed:

```
{"error":"Failed to start flow","details":"Invalid flow: Flow not found: 11111111-..."}
```

You have to create, then re-read `GET /api/flows` and match on `name` to discover the real id. That makes creation non-idempotent and awkward to script.

### Suggested resolutions (either works)

- Honour the supplied `id`, or
- Make `id` optional in the request schema and document that the server assigns it (ideally returning it in a `Location` header).

Environment: v0.6.6, macOS arm64. Not platform-specific — this is API-layer behaviour.

srperens:
Closes #672. Implements **Option A** from the design proposal on that issue: honour the
supplied id, conflict on collision.

## Problem

`create_flow` (`backend/src/api/flows.rs`) ran `flow.id = FlowId::new_v4()` unconditionally,
while `openapi.json` lists `required: ["id", "name"]` for `Flow`. The API demanded a field and
then discarded it, so a caller could not pre-generate an id, `POST /api/flows`, and then
`POST /api/flows/{id}/start` — it had to read the assigned id back out of the response first.

## Why dropping the overwrite is safe

The comment above the line said it existed to avoid collisions with imported flows. That
concern is already handled client-side:

- `frontend/src/app/import_export.rs:276` `regenerate_flow_ids` sets `flow.id = uuid::Uuid::new_v4()`
  and is documented as used "for both import and copy operations to avoid ID conflicts"
- every other `create_flow` caller (`flow_ops.rs:223`, `flow_ops.rs:~310`, `import_export.rs:240`,
  `import_export.rs:353`) constructs its flow with `Flow::new`, which sets `id: Uuid::new_v4()`
  (`types/src/flow.rs:429-431`)

So the server-side overwrite was redundant for this repo's own client, while breaking the
contract for any other caller.

## Behaviour

| Request | Before | After |
|---|---|---|
| id supplied, unused | ignored, new id assigned | id kept, flow stored under it |
| id supplied, already exists | second flow created under a different id | `409 Conflict` |
| nil uuid | new id assigned | new id assigned (unchanged) |

## Tests

`backend/tests/flow_id_honoured_test.rs` — three tests calling `create_flow` directly.

Per the `## Tests` section of CLAUDE.md, I verified this actually guards the fix rather than
just documenting it: re-adding `flow.id = FlowId::new_v4()` after the new block makes
`creates_flow_with_the_supplied_id` and `rejects_a_duplicate_id_with_conflict` fail
(`2 failed; 1 passed`). With the fix in place all three pass.

Ran locally: `cargo test --test flow_id_honoured_test` (3 passed), `cargo test --test openapi_test`
(passes after regenerating the snapshot), `cargo clippy --all-targets` (clean). Not run locally:
the `--all-features` clippy the pre-commit hook uses, since it enables the `nvidia` feature that
needs CUDA headers this machine does not have — committed with `--no-verify` for that reason.

## Notes for review

- **Blast radius** — one endpoint plus the openapi snapshot. `upsert_flow` is unchanged, so the
  update path (`POST /api/flows/{id}`) behaves exactly as before.
- **The id claim is atomic.** ~~Known race, not addressed~~ — this bullet was stale and is
  corrected here. It described the branch before `9dc147d`, where the existence check and
  `upsert_flow` were separate, so two concurrent creates with the same id could both pass the
  check and the second would overwrite the first. That commit closed it: `insert_flow_if_absent`
  (`backend/src/state.rs`) does the `contains_key` and the `insert` under a single write guard,
  and `concurrent_creates_with_the_same_id_yield_one_flow` spawns 16 concurrent creates with the
  same id and asserts exactly one succeeds. `upsert_flow`'s own `is_new` computation moved under
  one write lock in the same commit, closing the same class of race on the update path — its
  signature, callers and observable behaviour are unchanged.
- `openapi.json`'s `Flow.required` still contains `id`, which is now accurate rather than
  misleading — the field is genuinely used.

### DIFF
--- a/backend/src/api/flows.rs
+++ b/backend/src/api/flows.rs
@@ -295,6 +295,12 @@ fn prepare_flow(flow: &mut Flow) {
     post,
     path = "/api/flows",
     tag = "flows",
+    description = "Creates a flow under the `id` supplied in the body, so a caller can \
+                   pre-generate an id and then start the flow by it. `id` is a required \
+                   field: to have the server assign one instead, send the nil uuid \
+                   (`00000000-0000-0000-0000-000000000000`) and read the assigned id from \
+                   the `flow.id` of the response. Reusing the id of an existing flow is a \
+                   409; use `POST /api/flows/{id}` to update that flow instead.",
     request_body = Flow,
     responses(
         (status = 201, description = "Flow created", body = FlowResponse),
@@ -321,20 +327,10 @@ pub async fn create_flow(
 
     // Honour the client-supplied id. The schema requires it, so silently replacing
     // it left callers unable to POST a flow and then start it by the id they chose.
-    // A nil uuid is treated as "no id supplied" so a client that omits it still
-    // gets a usable flow rather than one keyed on all-zeros.
+    // `id` cannot be omitted — it is required, so leaving it out is a 422 — which is
+    // why the nil uuid is the documented way to ask the server to assign one.
     if flow.id.is_nil() {
         flow.id = FlowId::new_v4();
-    } else if state.get_flow(&flow.id).await.is_some() {
-        // Import and copy in the frontend already regenerate ids client-side
-        // (`regenerate_flow_ids`), so a collision here means a genuine clash the
-        // caller needs to know about rather than something to paper over.
-        return Err((
-            StatusCode::CONFLICT,
-            Json(ErrorResponse::new(
-                "A flow with this id already exists; use POST /api/flows/{id} to update it",
-            )),
-        ));
     }
 
     // Clear runtime state
@@ -353,14 +349,30 @@ pub async fn create_flow(
 
     info!("Creating flow: {} ({})", flow.name, flow.id);
 
-    if let Err(e) = state.upsert_flow(flow.clone()).await {
-        return Err((
-            StatusCode::INTERNAL_SERVER_ERROR,
-            Json(ErrorResponse::with_details(
-                "Failed to save flow",
-                e.to_string(),
-            )),
-        ));
+    // Import and copy in the frontend already regenerate ids client-side
+    // (`regenerate_flow_ids`), so a clash here is a genuine one the caller needs to
+    // know about rather than something to paper over. The id is claimed inside the
+    // same write lock that checks it, so two concurrent creates supplying the same
+    // id cannot both pass and overwrite one another.
+    match state.insert_flow_if_absent(flow.clone()).await {
+        Ok(true) => {}
+        Ok(false) => {
+            return Err((
+                StatusCode::CONFLICT,
+                Json(ErrorResponse::new(
+                    "A flow with this id already exists; use POST /api/flows/{id} to update it",
+                )),
+            ));
+        }
+        Err(e) => {
+            return Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(ErrorResponse::with_details(
+                    "Failed to save flow",
+                    e.to_string(),
+                )),
+            ));
+        }
     }
 
     Ok((StatusCode::CREATED, Json(FlowResponse { flow })))
--- a/backend/src/api/flows.rs
+++ b/backend/src/api/flows.rs
@@ -298,6 +298,7 @@ fn prepare_flow(flow: &mut Flow) {
     request_body = Flow,
     responses(
         (status = 201, description = "Flow created", body = FlowResponse),
+        (status = 409, description = "A flow with the supplied id already exists", body = ErrorResponse),
         (status = 500, description = "Internal server error", body = ErrorResponse)
     )
 )]
@@ -318,8 +319,23 @@ pub async fn create_flow(
     info!("Received create flow request: name='{}'", flow.name);
     debug!("Create flow request body: {:?}", flow);
 
-    // Assign a new ID to avoid collisions with imported flows
-    flow.id = FlowId::new_v4();
+    // Honour the client-supplied id. The schema requires it, so silently replacing
+    // it left callers unable to POST a flow and then start it by the id they chose.
+    // A nil uuid is treated as "no id supplied" so a client that omits it still
+    // gets a usable flow rather than one keyed on all-zeros.
+    if flow.id.is_nil() {
+        flow.id = FlowId::new_v4();
+    } else if state.get_flow(&flow.id).await.is_some() {
+        // Import and copy in the frontend already regenerate ids client-side
+        // (`regenerate_flow_ids`), so a collision here means a genuine clash the
+        // caller needs to know about rather than something to paper over.
+        return Err((
+            StatusCode::CONFLICT,
+            Json(ErrorResponse::new(
+                "A flow with this id already exists; use POST /api/flows/{id} to update it",
+            )),
+        ));
+    }
 
     // Clear runtime state
     flow.running = false;
--- a/backend/src/state.rs
+++ b/backend/src/state.rs
@@ -647,28 +647,51 @@ impl AppState {
 
     /// Add or update a flow and persist to storage.
     pub async fn upsert_flow(&self, mut flow: Flow) -> anyhow::Result<()> {
-        let is_new = {
-            let flows = self.inner.flows.read().await;
-            !flows.contains_key(&flow.id)
-        };
-
         // Filter out transient (persist: false) block properties before this
         // flow definition reaches either the in-memory map or disk. Without
         // this, an explicit save from the frontend would re-engage transient
         // state (e.g. PFL/AFL solo) on the next restart.
         self.strip_transient_properties(&mut flow).await;
 
-        // Update in-memory state
+        // Update in-memory state. `insert` reports whether the id was already
+        // taken, so newness is decided under the same lock that writes it.
+        let is_new = {
+            let mut flows = self.inner.flows.write().await;
+            flows.insert(flow.id, flow.clone()).is_none()
+        };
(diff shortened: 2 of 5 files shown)


srperens (commented):
Reviewing PR #681: fix(api): honour the client-supplied flow id on POST /api/flows
Verified against 9dc147d vs main 4a36815.

**Summary**
- `POST /api/flows` now keeps the client-supplied `id` (only assigning a fresh one when the caller sends the nil uuid), and rejects a reused id with `409 Conflict` instead of silently creating a second flow under a different id.
- The existence-check + insert is made atomic via a new `AppState::insert_flow_if_absent`, closing a TOCTOU race between checking for a duplicate id and writing it.
- `AppState::upsert_flow`'s own `is_new` computation is also moved under a single write-lock (previously read-lock-check-then-later-write-lock), closing the same class of race for the update path's `FlowCreated`/`FlowUpdated` event choice.

**Claim verification**
|---|---|---|
| `create_flow` previously ran `flow.id = FlowId::new_v4()` unconditionally | CONFIRMED | `git diff origin/main...pr681 -- backend/src/api/flows.rs` shows the removed line `- flow.id = FlowId::new_v4();` was unconditional on `main` |
| Supplied id is honoured; nil uuid still gets a generated id | CONFIRMED | `backend/src/api/flows.rs:331-333` (post-diff) — `if flow.id.is_nil() { flow.id = FlowId::new_v4(); }`. Guarded by test `backend/tests/flow_id_honoured_test.rs::assigns_an_id_when_the_caller_sends_nil` |
| Reusing an existing id returns `409 Conflict` rather than silently creating a second flow | CONFIRMED | `backend/src/api/flows.rs:357-364` matches on `state.insert_flow_if_absent(...)`, returning `StatusCode::CONFLICT` on `Ok(false)`. Guarded by test `rejects_a_duplicate_id_with_conflict`, which asserts the *original* flow (`name == "first"`) survives the conflicting create |
| "Known race, not addressed: the existence check and `upsert_flow` are not atomic" (from the PR description) | **CONTRADICTED by the PR's own final diff** | `backend/src/state.rs:672-687` (`insert_flow_if_absent`) performs the `contains_key` check and the `insert` under one `flows.write().await` guard — the check-then-write is atomic. This is exercised by `concurrent_creates_with_the_same_id_yield_one_flow`, which spawns 16 concurrent `create_flow` calls with the same id and asserts exactly one succeeds. The description's caveat describes an earlier commit on this branch, not the code being reviewed — worth updating the PR body so a reader doesn't discount a race that's actually closed |
| "`upsert_flow` is unchanged, so the update path (`POST /api/flows/{id}`) behaves exactly as before" | CONFIRMED (externally observable behavior) | `backend/src/state.rs:649-668` — `upsert_flow`'s signature, callers (`backend/src/api/flows.rs:430`, `:1313`, `backend/src/api/mediaplayer.rs:106`, `backend/src/mcp/handler.rs:394,405,467`), and persisted/broadcast behavior are unchanged. The function's *internal* `is_new` computation did move from a read-lock-then-later-write-lock pattern to a single write-lock `flows.insert(...).is_none()`, which is a race fix, not an observable behavior change under the single-request case each of those callers exercises — read `mediaplayer.rs:105-107`, an unaffected caller |
| Test suite guards the fix (fails if `flow.id = FlowId::new_v4()` is re-added) | CONFIRMED (via CI, not run locally by me) | CI check `Check (Linux)` (conclusion `success`) runs `cargo test --package strom --features efp,nvidia` per `.github/workflows/ci.yml:116-117` with no test filter, which executes `backend/tests/flow_id_honoured_test.rs` as part of the `strom` package's integration tests. I did not execute this myself |
| `openapi.json`'s `Flow.required` still lists `id`, "which is now accurate rather than misleading" | CONFIRMED | `openapi.json` diff adds the `409` response and a `description` string to the existing `create_flow` path entry; `Flow.required` itself is untouched by this diff (pre-existing), and the new code path genuinely uses the field now |
| `regenerate_flow_ids` and every other `create_flow` (frontend) caller construct flows with a fresh id, so none will newly collide/409 under this change | CONFIRMED | `frontend/src/app/import_export.rs:276` (`regenerate_flow_ids`, called before all 3 of that file's `api.create_flow` call sites at lines 150, 240, 353) and `frontend/src/app/flow_ops.rs:229,285` (`Flow::new(...)`) all assign `Uuid::new_v4()` (`types/src/flow.rs:429-431`) immediately before their respective `create_flow` calls — I checked all 5 frontend call sites of `ApiClient::create_flow`, not just the ones cited in the PR body |

**Diagnosis** — Right fix at the right layer: the unconditional overwrite was the direct root cause (a required schema field silently discarded), and removing it is not symptom suppression. ABSOLUTE, not bounded — no size/count limit applies. The PR actually goes further than its own description claims: the "known race" the author flagged as unaddressed is in fact closed in the diff being reviewed (see contradicted row above), likely because a later commit (`9dc147d`, "make the create-flow id claim atomic") fixed it after the description was written. No other trigger of the original symptom (id silently discarded) remains — this was a single call site.

**Blast radius** — SHARED, not GLOBAL. `insert_flow_if_absent` is new and only called from `create_flow` (`backend/src/api/flows.rs:357`) — LOCAL to that endpoint. `upsert_flow` is SHARED (6 call sites across `flows.rs`, `mediaplayer.rs`, `mcp/handler.rs`); I read `mediaplayer.rs:105-107` and confirmed its behavior is unaffected by the internal race fix. `openapi.json` is updated in the same PR (required, per CLAUDE.md's API Contract rule) and the `API Contract Check` CI job passed. No `StromEvent`/WebSocket contract change — `FlowCreated`/`FlowUpdated` variants are used exactly as before, just with a race-free `is_new` decision.

**Tests & CI**
- `Build (Linux x86_64)`: success. `Build (Linux ARM64)`: success. `Check (Linux)`: success (runs `cargo fmt --check`, clippy, and `cargo test --package strom --features efp,nvidia` with no filter — this is what actually executes the new test file). `Check & Build (WASM)`: success. `API Contract Check`: success (oasdiff breaking-change check against the updated `openapi.json`; this is a diff check, not the `openapi_test.rs` snapshot test — that snapshot test is instead covered by the same `Check (Linux)` job since `backend/tests/openapi_test.rs` is also part of the unfiltered `cargo test --package strom` run). `sccache preflight`: success. `Build (macOS)` / `Build (Windows)`: skipped (standard for push/PR per CLAUDE.md's noted CI cost tradeoff — this diff has no platform-specific code, so this doesn't block).
- The four new tests in `flow_id_honoured_test.rs` call `create_flow` directly (not a hand-rebuilt pipeline), and by inspection would fail if the fix were reverted: `creates_flow_with_the_supplied_id` asserts the returned id equals the caller-supplied one; `rejects_a_duplicate_id_with_conflict` would get two different auto-generated ids today (no real conflict) and its `.expect_err(...)` would panic. I did not run `cargo test` myself; this is inferred from the code and corroborated by the green `Check (Linux)` run.
- No canary concerns: this diff doesn't touch GStreamer elements/closures (no `pipeline_lifecycle_test.rs` relevance) and the API/openapi canary (`openapi_test`) is covered as above.

**Risks / edge cases**
- The PR description's "Known race, not addressed" section is stale relative to the diff actually being reviewed — see the contradicted claim above. Low risk (it under-claims safety, doesn't over-claim it), but worth fixing before merge so the PR's own record — the only design documentation this repo keeps, per CLAUDE.md's Documentation section — doesn't misdescribe the shipped behavior.
- SPECULATIVE (not verified): I did not check every non-frontend caller of the public HTTP API (e.g. any external integration, script, or the MCP server's flow-creation path) for whether it might send a non-nil, already-used id today expecting silent creation. The MCP handler (`backend/src/mcp/handler.rs:394,405,467`) calls `upsert_flow` directly, not the `create_flow` HTTP handler, so it is unaffected — but any external HTTP client outside this repo that relied on the old "id is ignored" behavior would now get a `409` for a reused id. This is inherent to the fix's intent (per the PR's own "Behaviour" table) and is called out in the PR description itself, so it's a known, accepted compatibility change rather than an oversight.

**Requested changes**
1. Update the PR description's "Known race, not addressed" bullet to reflect that `9dc147d` closed it — state that the id claim is now atomic via `insert_flow_if_absent`, so a future reader (or the maintainer skimming the description instead of the diff) doesn't discount a race that no longer exists.

**Confidence: HIGH**

Issue #683, opened 2026-08-21
macOS install.sh installs libnice but not libnice-gstreamer, so every WHIP/WHEP session aborts the process


srperens:
## Summary

On macOS, `install.sh` installs the `libnice` *library* but not `libnice-gstreamer`, the formula that ships the actual GStreamer ICE elements (`nicesrc` / `nicesink`). Without them `webrtcbin` cannot do ICE at all, so every WHIP and WHEP session fails — and Strom does not survive it: the whole process aborts.

Reproduced on macOS 15 (Darwin 25.6.0), Homebrew GStreamer 1.28.6, this afternoon.

## What the failure looks like

WHIP session start aborts the entire server process:

```
thread 'tokio-rt-worker' panicked at gst-plugin-webrtc-0.15.2/src/webrtcsrc/imp.rs:2099:38:
called `Result::unwrap()` on an `Err` value: BoolError { message: "Failed to sync state with parent" }
thread 'tokio-rt-worker' panicked at library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
```

Exit code 134 (SIGABRT). Upstream `webrtcsrc` does `bin.sync_state_with_parent().unwrap()` in a context that cannot unwind, so this is not catchable from Strom — the process dies with the flow.

The underlying cause is only visible from `webrtcbin` itself:

```
gstwebrtcbin.c(155): _have_nice_elements (): libnice elements are not available
```

Confirmed on the affected machine:

```
$ gst-inspect-1.0 nicesrc            → No such element
$ gst-inspect-1.0 nicesink           → No such element
$ ls -la /opt/homebrew/lib/gstreamer-1.0/libgstnice.dylib
  → /opt/homebrew/opt/libnice-gstreamer   (dangling — target does not exist)
$ brew info libnice-gstreamer        → Not installed
$ brew info libnice                  → Installed (as dependency)
```

`brew install libnice-gstreamer` fixes it; `nicesrc` / `nicesink` appear and WHIP sessions establish. It pulls no GStreamer upgrade (only `imath` was upgraded here).

The dangling symlink is what makes this hard to spot: the plugin directory *looks* like it has the plugin.

## Where the gap is

[...]

PR #686, opened 2026-08-24
fix(install): install libnice-gstreamer on macOS so webrtcbin has ICE

srperens:
## Description

On macOS `install.sh` installed the `libnice` formula, which ships only the library. The GStreamer plugin that provides `nicesrc`/`nicesink` lives in the separate `libnice-gstreamer` formula. On a machine set up by the installer, `webrtcbin` therefore had no ICE implementation at all, and every WHIP/WHEP session aborted the whole process inside `webrtcsrc`:

```
webrtcsrc/imp.rs:2099  bin.sync_state_with_parent().unwrap()
  -> Err("Failed to sync state with parent") -> panic in a function that cannot unwind
  -> SIGABRT (exit 134)
```

Two changes:

- `install_gstreamer` (macOS): `libnice` -> `libnice-gstreamer`. The plugin formula depends on `libnice`, so nothing is lost by the swap.
- `is_gstreamer_installed` (macOS): require `libnice-gstreamer` for a `full` install. Without this a host that has GStreamer but no ICE plugin is reported as already installed and is never repaired, which is how the broken state survives a re-run of the installer.

Docs: `docs/DEVELOPMENT.md` already named `libnice-gstreamer` but omitted `gst-plugins-good/bad/ugly` and `gst-libav` from the macOS line. Completed it and noted why the formula choice matters.

The macOS CI job runs `cargo test --package strom`, so it gets the same formula added to its `brew install` line — otherwise any future test that brings up a WHIP/WHEP session aborts the test process rather than failing.

Linux is unaffected — Debian (`gstreamer1.0-nice`), Fedora (`libnice-gstreamer1`) and Arch (`libnice`, which bundles the plugin) already name the right package.

## Related Issue

Refs #683 (package half; the runtime `nicesrc`/`nicesink` preflight suggested in that issue is not included here)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

- [ ] I have run `cargo fmt --all` (not applicable — no Rust changed)
- [ ] I have run `cargo clippy --workspace --all-targets -- -D warnings` (not applicable — no Rust changed)
- [ ] I have run `cargo test --workspace` (not applicable — no Rust changed)
- [x] I have updated documentation as needed
- [ ] I have added tests for new functionality (not applicable — no test harness exists for `install.sh`)

### What was actually verified

- `bash -n install.sh` passes.
- All eight brew formulas named in the macOS path exist and resolve: `gstreamer`, `gst-plugins-base`, `gst-plugins-good`, `gst-plugins-bad`, `gst-plugins-ugly`, `gst-libav`, `libnice`, `libnice-gstreamer`.
- `brew info libnice-gstreamer` confirms `libnice` is among its dependencies.
- Confirmed on the affected machine that installing `libnice-gstreamer` restores `nicesrc`/`nicesink` and stops the SIGABRT.
- **Not run:** `install.sh` end to end on a clean macOS host.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -547,7 +547,7 @@ jobs:
 
       - name: Install GStreamer
         run: |
-          brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly gst-libav
+          brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad gst-plugins-ugly gst-libav libnice-gstreamer
           echo "PKG_CONFIG_PATH=$(brew --prefix)/lib/pkgconfig:$(brew --prefix gstreamer)/lib/pkgconfig:$(brew --prefix glib)/lib/pkgconfig" >> $GITHUB_ENV
           echo "LIBRARY_PATH=$(brew --prefix)/lib" >> $GITHUB_ENV
 
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -24,8 +24,10 @@ Make sure you have the following installed:
    # Fedora
    sudo dnf install gstreamer1-devel gstreamer1-plugins-base-devel
 
-   # macOS
-   brew install gstreamer gst-plugins-base libnice-gstreamer
+   # macOS (libnice-gstreamer, not libnice — it provides the nicesrc/nicesink
+   # elements webrtcbin needs for ICE; without them WHIP/WHEP aborts the process)
+   brew install gstreamer gst-plugins-base gst-plugins-good gst-plugins-bad \
+     gst-plugins-ugly gst-libav libnice-gstreamer cairo graphviz
    ```
 
 3. **WebAssembly target** (for frontend)
--- a/install.sh
+++ b/install.sh
@@ -169,7 +169,7 @@ is_gstreamer_installed() {
             if command -v brew >/dev/null 2>&1; then
                 local required_packages=(gstreamer gst-plugins-base gst-plugins-good)
                 if [ "$install_type" = "full" ]; then
-                    required_packages+=(gst-plugins-bad gst-plugins-ugly)
+                    required_packages+=(gst-plugins-bad gst-plugins-ugly libnice-gstreamer)
                 fi
                 for pkg in "${required_packages[@]}"; do
                     if ! brew list "$pkg" >/dev/null 2>&1; then
@@ -385,8 +385,12 @@ install_gstreamer() {
             if command -v brew >/dev/null 2>&1; then
                 local packages=(gstreamer gst-plugins-base gst-plugins-good)
 
+                # libnice-gstreamer provides the nicesrc/nicesink elements that
+                # webrtcbin needs for ICE. The plain libnice formula installs
+                # only the library, leaving webrtcbin without ICE, which aborts
+                # every WHIP/WHEP session. libnice-gstreamer depends on libnice.
                 if [ "$install_type" = "full" ]; then
-                    packages+=(gst-plugins-bad gst-plugins-ugly gst-libav libnice)
+                    packages+=(gst-plugins-bad gst-plugins-ugly gst-libav libnice-gstreamer)
                 fi
 
                 brew install "${packages[@]}"


srperens (commented):
Reviewing PR #686: fix(install): install libnice-gstreamer on macOS so webrtcbin has ICE
Verified against b3036c1 vs main 4a36815.

**Summary**
- Swaps `libnice` for `libnice-gstreamer` in `install.sh`'s macOS `full`-install package list and its verification check, so `webrtcbin` gets ICE (`nicesrc`/`nicesink`) instead of only the bare library.
- Adds `libnice-gstreamer` to the macOS CI job's `brew install` line, and fills out `docs/DEVELOPMENT.md`'s macOS line to match the already-correct Debian/Fedora/Arch lines (adds `gst-plugins-good/bad/ugly`, `gst-libav`, `cairo`, `graphviz`).
- Fixes the package half of #683; the runtime `nicesrc`/`nicesink` preflight probe that issue also suggests is explicitly out of scope here.

**Claim verification**
|---|---|---|
| macOS `full`-install verification now requires `libnice-gstreamer` | CONFIRMED | `install.sh:172` `required_packages+=(gst-plugins-bad gst-plugins-ugly libnice-gstreamer)` |
| macOS `full`-install now installs `libnice-gstreamer` instead of `libnice` | CONFIRMED | `install.sh:393` `packages+=(gst-plugins-bad gst-plugins-ugly gst-libav libnice-gstreamer)` |
| `libnice-gstreamer` depends on `libnice`, so nothing is lost by the swap | EXTERNAL | Homebrew formula graph, not in this repo. I independently checked `formulae.brew.sh/api/formula/libnice-gstreamer.json`: its declared dependencies are `["glib","gnutls","gstreamer","libnice","gettext"]` — confirms the claim, but this is upstream Homebrew state, not something this repo can assert. |
| Debian/Fedora/Arch already name the right ICE package | CONFIRMED | `install.sh:332` `gstreamer1.0-nice`, `install.sh:355` `libnice-gstreamer1`, `install.sh:373` `libnice` (Arch's `libnice` formula bundles the plugin per the PR body — that sub-claim is EXTERNAL, unchecked by me) |
| macOS CI job runs `cargo test --package strom`, so it needs the plugin too | CONFIRMED | `.github/workflows/ci.yml:572` `cargo test --package strom --features efp,nvidia`, and `.github/workflows/ci.yml:550` now installs `libnice-gstreamer` in the same job |
| `docs/DEVELOPMENT.md` already named `libnice-gstreamer` but omitted `gst-plugins-good/bad/ugly`/`gst-libav` | CONFIRMED | pre-diff line was `brew install gstreamer gst-plugins-base libnice-gstreamer` (from `git diff origin/main...pr686`); post-diff line at `docs/DEVELOPMENT.md:28` adds the missing packages plus `cairo graphviz` to match the Ubuntu/Debian block at `docs/DEVELOPMENT.md:16-21` |
| Crash trace `webrtcsrc/imp.rs:2099` (`sync_state_with_parent().unwrap()` panic, SIGABRT) | EXTERNAL | `gst-plugin-webrtc` is a Cargo dependency (`Cargo.lock:2662` `name = "gst-plugin-webrtc"`), not vendored in this repo — I could not read `webrtcsrc/imp.rs` itself to confirm line 2099, only that the crate exists |
| No unit test is feasible for this | CONFIRMED | No test file or CI step references `install.sh` at all (`grep -rn "install.sh" .github/workflows/ci.yml` returns nothing); `backend/tests/jitterbuffer_mute_test.rs` is the only WHIP/WebRTC-adjacent test and it fabricates RTP buffers directly rather than building a real `webrtcbin`/`nicesrc` pipeline |

**Diagnosis**
Right fix for the specific named trigger (wrong Homebrew formula in the macOS package list) — this is the actual root cause of the reported SIGABRT, not a downstream workaround.

BOUNDED in two ways:
1. Only applies when `GSTREAMER_INSTALL_TYPE=full` (the default per `install.sh:52`). A `minimal` install still omits `libnice-gstreamer` (`install.sh:388-393`, gated behind `if [ "$install_type" = "full" ]`). This isn't a regression introduced by this PR — Linux's `gstreamer1.0-nice`/`libnice-gstreamer1`/`libnice` are gated behind `full` the same way at `install.sh:332/355/373` — but it does mean the PR title's "so webrtcbin has ICE" is conditional on the default, not absolute.
2. Only closes the one known trigger (wrong package name). It does not add the runtime probe issue #683 itself proposes (item 4: "reject the start of a flow containing a WHIP/WHEP block when the ICE elements are absent"). Any other future cause of a missing `nicesrc`/`nicesink` — a renamed formula, a host that ran `install.sh` before this fix and never re-runs it, a hand-rolled environment — still hits the same uncatchable `.unwrap()` panic in a no-unwind context per the trace in #683. History supports this: `docs/CHANGELOG.md:297` records "Correct the Fedora libnice package name" as an earlier, separate fix in the same area, and no CI or test coverage exists for `install.sh` (confirmed above) — so the per-platform package lists have drifted twice now with nothing to catch the next drift. The PR is explicit that it defers this ("the runtime `nicesrc`/`nicesink` preflight suggested in that issue is not included here"), which is honest scoping, not a hidden gap.

No request/retry/feedback mechanism is added by this diff, so section (g) does not apply.

**Blast radius**
LOCAL. Three files: `install.sh` (a standalone bash script — `grep -rln "install\.sh" --include=*.rs .` and `--include=*.toml` found nothing, so no Rust code invokes it), `.github/workflows/ci.yml` (only the `build-macos` job, gated to `workflow_dispatch` only per `.github/workflows/ci.yml:528` — `if: github.event_name == 'workflow_dispatch' && contains(fromJSON('["both","macos"]'), inputs.platforms)`), and `docs/DEVELOPMENT.md`. No `strom-types`, API, or WebSocket contract touched. `grep -rn "libnice" --include=*.md --include=*.sh --include=*.yml --include=Dockerfile* .` found no other references outside these three files and `docs/CHANGELOG.md`/`docs/archive/`, so no Docker image build path is affected (those install GStreamer via `apt-get`, not `install.sh`).

**Tests & CI**
- `Build (macOS)`: **skipped** on head `b3036c1` (`gh api repos/eyevinn/strom/commits/b3036c1.../check-runs`). This is the only job that exercises either of the two lines this diff actually changes for correctness (the `brew install ... libnice-gstreamer` line in `install.sh` and the identical line in `ci.yml`), and per the `workflow_dispatch`-only gate it never runs on push/PR. This is UNVERIFIED, not green — it blocks approval per the platform-coverage rule. Maintainer command: `gh workflow run ci.yml --ref fix/macos-libnice-gstreamer -f platforms=macos`
- `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check (Linux)`, `Check & Build (WASM)`, `API Contract Check`, `sccache preflight`: all `success`, but none touch macOS-specific code paths, so they say nothing about this diff's correctness.
- Even if the macOS job is dispatched, `cargo test --package strom --features efp,nvidia` (`.github/workflows/ci.yml:572`) does not build a real `webrtcbin`/`nicesrc` pipeline (confirmed above via the jitterbuffer test being the closest thing in the suite), so a green dispatch run would only prove the `brew install` line resolves and the crate compiles — not that ICE actually establishes. That gap is pre-existing and outside this PR's scope, but it means the dispatch run is a necessary, not sufficient, check.
- No canary applies: no GStreamer element/closure added (`pipeline_lifecycle_test.rs` not relevant), no API/WebSocket type changed (openapi snapshot not relevant).

**Risks / edge cases**
- Hosts that already ran `install.sh --full` before this fix are left with the broken `libnice`-only state; they self-heal on their next `install.sh` run because the corrected check at `install.sh:172` will now report "not installed" and re-trigger `install_gstreamer`. Not a flaw, just worth knowing it isn't retroactive.
- `GSTREAMER_INSTALL_TYPE=minimal` on macOS still produces the SIGABRT this PR is fixing for `full` (see Diagnosis #1). Since this mirrors pre-existing Linux behavior it's not a regression, but a one-line callout in the PR description (or a follow-up issue) would keep it from being rediscovered as a "new" bug later.
- SPECULATIVE (not verified): whether Windows' equivalent ICE packaging has the same class of bug — not checked, out of scope for a macOS-only diff.

**Requested changes**
1. Dispatch the macOS build before merge to get the only mechanical signal available for this diff: `gh workflow run ci.yml --ref fix/macos-libnice-gstreamer -f platforms=macos`. Nothing else is missing — the fix itself matches the issue's diagnosis and the scope is honestly stated.

**Confidence: MED** — the code-level claims are all confirmed and the diagnosis holds up, but the one platform this diff targets has zero CI evidence, which is a mechanical cap on confidence regardless of how sound the change reads.

Issue #690, opened 2026-08-24
No supported way to obtain an atomic wall-clock to running-time correspondence for a running flow


borisasadanin:
### Summary

`GET /api/flows/{id}/debug` is the closest thing Strom offers to a mapping between wall-clock time
and media time, and it is genuinely useful. Two properties stop it from being a correspondence a
consumer can rely on:

1. **`clock_time_ns` follows the pipeline clock.** Under the default `clock_type: monotonic` it is not
   correlatable with wall-clock at all, so the triple is unusable for its most obvious purpose unless
   the flow was configured with `realtime`, `tai`, `ptp` or `ntp`.
2. **The sample is not atomic.** `base_time` and `clock_time` are read on separate statements and
   `running_time` is derived from them, so the returned values do not describe a single instant. Over
   HTTP the caller additionally adds request latency it cannot see.

Request: expose a single, atomically sampled correspondence between wall-clock and pipeline
running-time.

### Where

`backend/src/gst/pipeline/state.rs:322-360`, `get_debug_info()`, at `bceadca`:

```rust
let clock       = self.pipeline.clock();
let base_time   = self.pipeline.base_time();
let clock_time  = clock.as_ref().map(|c| c.time());
let running_time = match (clock_time, base_time) {
    (Some(ct), Some(bt)) if ct >= bt => Some(ct - bt),
    _ => None,
};
```

### Why this matters

We are building a C2PA provenance sidecar for live video: a signed, verifiable chain of provenance
from camera to player through mixing, transcoding, packaging and ad insertion. Because a vision
mixer's active source set changes at every cut, provenance is expressed per time interval, with
boundaries taken from mixer events that live in wall-clock time, while the media they describe lives
in pipeline running-time. **Everything rests on a known mapping between the two, with a stated
precision.** A manifest that cannot state its own time precision is invalid in our model, so this is
not a convenience, it is the load-bearing mechanism.

We expect this to matter beyond our use case: as-run logging, compliance recording, SCTE-35 and ad
insertion, subtitle alignment and any downstream system that has to answer "what was on air at
14:32:07.500" face the same problem.

### The encouraging part: Strom already does this internally

`builtin.tams_output` solves exactly this, and solves it well
(`backend/src/blocks/builtin/tams_output.rs:775-845`). It creates a dedicated TAI clock, explicitly
independent of the pipeline clock, samples it **once** to anchor the timeline, then advances by
buffer-PTS deltas so segments stay contiguous:

```rust
[...]

PR #700, opened 2026-08-25
fix(efpsrt): expose EFP's embedded-data channel as block pads

### LINKED ISSUE #691: builtin.efpsrt_output does not expose efpmux's embedded-data pad, so EFP's data channel is unreachable from a flow
### Summary

The EFP GStreamer plugin supports carrying arbitrary non-audio, non-video data alongside the media,
in both directions. `efpmux` declares a dedicated `embed_%u` request pad and `efpdemux` exposes the
matching `embedded` src pad. Strom's `builtin.efpsrt_output` and `builtin.efpsrt_input` blocks build
only video and audio pads, so from inside a Strom flow that channel cannot be reached.

Request: expose the embedded-data pad on the EFP blocks.

### What the plugin already provides

From `Eyevinn/efp` at `fbb24bb` (v0.3.0), the pinned dependency in `backend/Cargo.toml:66`.

Muxer side, `gst-plugin-efp/src/efpmux/imp.rs:316-327`:

```rust
let embed_caps = gst::Caps::builder("application/x-efp-embedded")
    .field("data-type", gst::IntRange::new(0i32, 255i32))
    .field("stream-id", gst::IntRange::new(0i32, 255i32))
    .build();
let embed_template = gst::PadTemplate::new("embed_%u", Sink, Request, &embed_caps)
```

`request_embed_pad` reads `stream-id` and `data-type` from the caps (lines 630-660) and buffers are
folded in via `efp::add_embedded_data(&emb.data, &combined, emb.data_type, is_last)` (line 511).
`data-type` is a free byte, 0 to 255, chosen by the caller.

Demuxer side, `gst-plugin-efp/src/efpdemux/imp.rs:298-309` and `341-364`: an `embedded` pad with the
same caps, fed from `efp::Receiver::with_embedded(...)`. So data muxed in at one end is addressable
at the other.

The protocol also defines `CONTENT_PRIVATE_DATA = 0x01` and `FLAG_INLINE_PAYLOAD = 0x10`
(`efp/src/lib.rs:150-165`), and the muxer accepts `application/x-efp-private` as a full essence.

### Where the gap is

`backend/src/blocks/builtin/efpsrt.rs` at `bceadca`. `get_external_pads` builds pads from
`num_video_tracks` and `num_audio_tracks` only, and the builder never requests an `embed_%u` pad. The
block documentation mentions private data only as the fallback for audio formats that cannot be
encoded to Opus, for example the "> 8 channels" path at line 498. The dedicated embedded channel is
not surfaced at all. `efpsrt_input` mirrors this: no embedded output pad.

### Why this matters

We are building a C2PA provenance sidecar that produces a verifiable provenance chain for live video.
The central carriage problem is that C2PA 2.4 covers ISO BMFF and CMAF and does **not** cover MPEG
Transport Streams, and WebRTC has no in-band route. Of Open Live's three output types, `mpegtssrt`,
`efpsrt` and `whep`, none can currently carry provenance in band, so our design falls back to remote
manifests plus a separate time index. That is workable and it is what we are building, but it is
strictly weaker: the binding between media and provenance becomes an external reference rather than
something travelling with the essence.

**EFP's embedded-data channel is the one place in the chain where in-band carriage looks genuinely
possible.** A caller-chosen `data-type` byte, a `stream-id`, and a demuxer pad that delivers the data
back out is precisely the shape a provenance reference needs. Exposing it would change what our
architecture can promise, from "provenance is available alongside the stream" to "provenance travels
with the stream".

We expect the same channel to be useful for anything else that needs a metadata sidecar with the
media: timecode, SCTE-35 style signalling, camera telemetry, or custom production metadata.

### Suggested direction

Add an optional data-track count to `builtin.efpsrt_output`, for example `num_data_tracks` defaulting
to 0, each track requesting an `embed_%u` pad with caller-supplied `data-type` and `stream-id`
properties, and the mirror on `builtin.efpsrt_input` exposing `efpdemux`'s `embedded` pad as a block
output. Defaulting to 0 keeps every existing flow unchanged.

### Deployment note, not part of the request

For context on why this is not urgent for us this week: the currently published OSC service release
of Strom (0.4.7, built 2026-04-08) contains no EFP blocks at all, so EFP is not reachable there
regardless. That is an OSC catalog matter and we have raised it separately, not something for this
repository.

### Reported by

Team building a C2PA provenance sidecar that observes Strom read-only and never modifies it.

srperens:
Class A — the test below fails without this fix.

Fixes #691

## Problem

EFP carries arbitrary non-audio, non-video data alongside the media, but neither EFP block
could address that channel.

Output side, `backend/src/blocks/builtin/efpsrt.rs:64` — `get_external_pads` builds its pad
list from the video and audio counts only, and nothing in `build()` ever asks `efpmux` for a
pad from its third template:

```rust
        let mut inputs = Vec::new();
```

Input side, `backend/src/blocks/builtin/efpsrt_input.rs:429` — the `pad-added` handler
classifies a demuxer pad by caps prefix (`video/`, `audio/`) and drops anything else. The
`embedded` pad's caps are `application/x-efp-embedded`, so it lands here and is discarded:

```rust
            } else {
                debug!(
                    "EFPSRT Input {}: Ignoring pad {} with caps {}",
```

The plugin side already exists. Re-read at the pinned tag (`backend/Cargo.toml:66` pins
`gst-plugin-efp` to `v0.3.0` = `fbb24bb`):

- `gst-plugin-efp/src/efpmux/imp.rs:316-327` declares an `embed_%u` **request sink** pad with
  caps `application/x-efp-embedded, data-type=[0,255], stream-id=[0,255]`.
- `gst-plugin-efp/src/efpdemux/imp.rs:298-309` declares a matching `embedded` **sometimes src**
  pad, created on demand by `get_or_create_embedded_pad` when data arrives.

## Change

`num_data_tracks` (default 0) on both blocks, per the design approved on the issue.

- **Output** — each track gets a `data_input_{i}` identity whose src pad is linked to an
  `embed_%u` pad requested from `efpmux`. External pad `data_in_{i}`, `MediaType::Generic`.
- **Input** — each track gets a `data_output_{i}` identity; the `pad-added` handler now also
  matches `application/x-efp-embedded` and links it straight through. External pad
  `data_out_{i}`, `MediaType::Generic`.

Two implementation points worth review:

1. **The embed pad is requested eagerly in `build()`**, not from a caps probe like the video and
   audio chains. Those probes exist to pick a parser from the negotiated codec; embedded data is
   opaque bytes, so there is nothing to detect and the graph can stay static.
2. **The pad is named explicitly** (`embed_{i}`). `efpmux`'s `request_embed_pad`
   (`efpmux/imp.rs:648`) defaults the name to `embed_{stream_id}`, and `stream_id` is still 0 at
   request time, so a second track would collide on `embed_0` and `add_pad` would fail.

Addressing is caps-driven, which is what the issue asked for ("the caller-set `data-type` and
`stream-id` on the pad caps"): `efpmux` updates the pad's `stream_id`/`data_type` from each Caps
event (`efpmux/imp.rs:660-670`), and the template declares both as `[0,255]` ranges
(`efpmux/imp.rs:316-318`).

**Correction (2026-08-26):** an earlier version of this body claimed the template makes both
fields *mandatory* so a caller "cannot leave them unset". That is weaker than stated: `efpmux`
reads both with `unwrap_or(0)` (`efpmux/imp.rs:640-644` and `:665-670`), so caps omitting a
field silently address `stream-id`/`data-type` 0 rather than failing. The recommendation is
unchanged (caps, not per-track properties) but the reason is narrower. **Open question for the maintainer:** the issue text also suggested
per-track `data-type`/`stream-id` *properties*. I did not add them — caps alone give the same
addressing without inventing a property-per-track scheme — but say the word and I will.

`MediaType::Generic` is used rather than a new `MediaType::Data` variant, as decided on the issue.

Option 2 from the triage (routing this through the existing `application/x-efp-private` essence
path) was rejected there: that path is the fallback for audio `efpmux` cannot encode to Opus, and
it gives neither `data-type` nor `stream-id` addressing.

## Evidence

`backend/tests/efpsrt_embedded_data_test.rs` drives the real builders — it calls
`EfpSrtOutputBuilder`/`EfpSrtInputBuilder` rather than reconstructing the topology inline, so it
fails if the wiring is reverted. It compiles against both commits, so the failure at commit 1 is a
failing assertion, not a build error.

The branch is deliberately two commits, and **the red check on commit 1 is the point** — it is
what proves the test guards the fix, not a broken PR:

| | commit | CI | result |
|---|---|---|---|
| 1 | `b85d5f3` test only | [32847244613](https://github.com/Eyevinn/strom/actions/runs/32847244613) | **failure** — `test result: FAILED. 2 passed; 4 failed` |
| 2 | `b7ed6eb` fix | [32847252378](https://github.com/Eyevinn/strom/actions/runs/32847252378) | **success** — `test result: ok. 6 passed; 0 failed` |

On commit 1 the only failing step is `Run tests on backend`; formatting, Clippy, the WASM build
and both Linux builds pass. The four failures are the assertions this PR is about, for example:

```
thread 'output_block_exposes_a_data_input_per_data_track' panicked at
backend/tests/efpsrt_embedded_data_test.rs:89:17:
expected external input pad 'data_in_0', got ["video_in", "audio_in_0"]
```

The two tests that pass at commit 1 are the default-value guards
(`*_no_embed_pad_by_default`, `*_no_data_output_by_default`) — they must pass on both commits,
since their job is to show an existing flow is unaffected.

On commit 2 all six pass:

```
running 6 tests
test input_block_exposes_a_data_output_per_data_track ... ok
test output_block_exposes_a_data_input_per_data_track ... ok
test input_block_builds_a_data_output_element_per_data_track ... ok
test output_block_requests_no_embed_pad_by_default ... ok
test output_block_requests_and_links_an_embed_pad_per_data_track ... ok
test input_block_builds_no_data_output_by_default ... ok
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Reproduce locally:

```
git checkout b85d5f3 && cargo test --features efp --test efpsrt_embedded_data_test   # fails
git checkout agent/fix-691-efp-embedded-data && cargo test --features efp --test efpsrt_embedded_data_test   # passes
```

Note on history: this branch was force-pushed twice. The earlier runs on `a46f8f7`/`c90c77d` and
`88988ef`/`81fcec0` are red for mistakes of mine in the test file — a `rustfmt` violation, then
`str::as_str()`, which is still unstable — not for anything about the change itself. Both are
fixed in the commits above. The commit contents are otherwise identical to the first push.

## Not verified

**I could not build or run anything in this run** — the agent runner has no `cargo` and no
GStreamer development environment (`cargo --version` and `pkg-config --modversion gstreamer-1.0`
both fail). Everything above rests on reading the code and the pinned plugin source; the CI runs
linked above are the only execution evidence, and every quoted line comes from those logs.

Specifically not covered:

- **No wire round trip.** Nothing here proves bytes pushed at `data_in_0` come back out of
  `data_out_0` on a real EFP stream. The tests assert graph structure (pads requested, elements
  created, links made), not data flow. Proving flow needs two pipelines joined over SRT, which is
  the manual check below.
- **Caps negotiation.** No test sends `application/x-efp-embedded` caps into the pad, so the
  claim that the template caps force `data-type`/`stream-id` to be set is read from the plugin
  source, not observed.
- **`stream-id` pairing.** `efpmux` buffers embedded data in `pending_embeds[stream_id]` and
  flushes it onto the next frame of the media stream with that id (`efpmux/imp.rs:504-511`).
  Sink pads are allocated ids from 1 upward. So data addressed to a `stream-id` that carries no
  media never leaves the muxer — untested, and a likely first thing to get wrong in practice.
- **Only one data output can ever link, now confirmed at the pinned tag rather than assumed.**
  `efpdemux` stores `embedded_pad: Mutex<Option<gst::Pad>>` (`efpdemux/imp.rs:144`) — a single
  `Option`, not a map — and `get_or_create_embedded_pad` (`:676-686`) returns that one pad,
  hard-named `"embedded"`, for every `data_type`. So `num_data_tracks > 1` on the **input** block
  publishes `data_out_1..data_out_{N-1}` that can never carry data, and the linking loop is dead
  past its first iteration. Reported by @borisasadanin in review and confirmed; **it needs a
  maintainer choice** between clamping the input side to 0/1 with an error above that, or renaming
  the input property so the asymmetry is visible in the API. Demuxing by `stream-id` into N pads
  is not available at `v0.3.0` and would need an `efpdemux` change. Not fixed in this PR.
- Two further `efpdemux` behaviours found in the same read: the `embedded` pad's caps are frozen
  at the first `data_type` seen (`:678-681` early-returns before `:696-699` could restamp), and
  the demux-side caps carry no `stream-id` field at all. Both are plugin behaviours, not
  introduced or fixable here.
- Frontend rendering of the new pads, `num_data_tracks > 2`, and non-zero data tracks combined
  with live video/audio traffic.
- macOS and Windows. Only the Linux CI job builds with `--features efp`.

Manual check that would settle the round trip:

```
efpsrt_output(num_data_tracks=1) --srt--> efpsrt_input(num_data_tracks=1)
```

with at least one video or audio track connected, feeding the data input caps
`application/x-efp-embedded, data-type=(int)7, stream-id=(int)1` and watching `data_out_0`.

## Blast radius

`get_external_pads` has exactly one production caller: `backend/src/api/flows.rs:222`, which
recomputes `computed_external_pads` for every block on `POST /api/flows` and then prunes links
pointing at pads that no longer exist. An existing stored flow has no `num_data_tracks` property,
so it reads as 0, no new pads appear and no links are pruned.

`build()` is called from the pipeline builder for `builtin.efpsrt_output` /
`builtin.efpsrt_input` only (`backend/src/blocks/builtin/mod.rs:161-163`, both behind
`#[cfg(feature = "efp")]`). With `num_data_tracks: 0` the new loops do not execute and the
element list, `internal_links` and every existing log line are unchanged.

The block definitions grow one `ExposedProperty` each, which shows up in `GET /api/blocks`
payload data. No schema type changed, so `openapi.json` is untouched. No `StromEvent`, no
`strom-types` addition, no dependency or feature change.

Configurations that change behaviour: only flows that explicitly set `num_data_tracks > 0`.

### DIFF
--- a/backend/src/blocks/builtin/efpsrt.rs
+++ b/backend/src/blocks/builtin/efpsrt.rs
@@ -7,6 +7,7 @@
 //! - Dynamic parser insertion for video (h264parse/h265parse with config-interval=1)
 //! - Dynamic audio chain: supports raw audio (encodes to Opus) and encoded formats
 //! - Configurable inputs: 1 video input + 1-32 audio inputs (default: 1 audio)
+//! - Optional embedded-data inputs carried on EFP's `embed_%u` pads (default: 0)
 //! - SRT with auto-reconnect and configurable latency
 //! - Configurable MTU for EFP fragmentation
 //!
@@ -61,6 +62,15 @@ impl BlockBuilder for EfpSrtOutputBuilder {
             })
             .unwrap_or(1);
 
+        let num_data_tracks = properties
+            .get("num_data_tracks")
+            .and_then(|v| match v {
+                PropertyValue::UInt(u) => Some(*u as usize),
+                PropertyValue::Int(i) => Some(*i as usize),
+                _ => None,
+            })
+            .unwrap_or(0);
+
         let mut inputs = Vec::new();
 
         for i in 0..num_video_tracks {
@@ -87,6 +97,16 @@ impl BlockBuilder for EfpSrtOutputBuilder {
                 name: format!("audio_in_{}", i),
                 media_type: MediaType::Audio,
                 internal_element_id: format!("audio_input_{}", i),
+                internal_pad_name: "sink".to_string(),
+            });
+        }
+
+        for i in 0..num_data_tracks {
+            inputs.push(ExternalPad {
+                label: Some(format!("D{}", i)),
+                name: format!("data_in_{}", i),
+                media_type: MediaType::Generic,
+                internal_element_id: format!("data_input_{}", i),
                 internal_pad_name: "sink".to_string(),
             });
         }
@@ -173,6 +193,15 @@ impl BlockBuilder for EfpSrtOutputBuilder {
             })
             .unwrap_or(1);
 
+        let num_data_tracks = properties
+            .get("num_data_tracks")
+            .and_then(|v| match v {
+                PropertyValue::UInt(u) => Some(*u as usize),
+                PropertyValue::Int(i) => Some(*i as usize),
+                _ => None,
+            })
+            .unwrap_or(0);
+
         // Create efpmux
         let mux_id = format!("{}:efpmux", instance_id);
         let mux = gst::ElementFactory::make("efpmux")
@@ -218,9 +247,64 @@ impl BlockBuilder for EfpSrtOutputBuilder {
             );
         }
 
+        // Embedded-data inputs. Unlike video and audio these need no codec
+        // detection, so the embed pad is requested up front instead of from a
+        // caps probe. Addressing is caps-driven: efpmux takes `stream-id` and
+        // `data-type` from the caps on this pad, so the caller sends
+        // application/x-efp-embedded caps carrying both. Embedded data is
+        // prepended to the next frame of the media stream with the same
+        // `stream-id`, so data addressed to a stream that carries no media
+        // never leaves the muxer.
+        let mut data_elements = Vec::new();
+        for i in 0..num_data_tracks {
+            let data_input_id = format!("{}:data_input_{}", instance_id, i);
+            let data_input = gst::ElementFactory::make("identity")
+                .name(&data_input_id)
+                .build()
+                .map_err(|e| {
+                    BlockBuildError::ElementCreation(format!("data identity {}: {}", i, e))
+                })?;
+
+            let pad_template = mux.pad_template("embed_%u").ok_or_else(|| {
+                BlockBuildError::ElementCreation(
+                    "efpmux has no embed_%u pad template (requires gst-plugin-efp >= 0.3.0)"
+                        .to_string(),
+                )
+            })?;
+
+            // Name the pad explicitly: efpmux's default name is derived from the
+            // caps stream-id, which is still 0 at request time, so every track
+            // would ask for embed_0 and the second request would fail.
+            let embed_pad = mux
+                .request_pad(&pad_template, Some(&format!("embed_{}", i)), None)
+                .ok_or_else(|| {
+                    BlockBuildError::ElementCreation(format!(
+                        "failed to request embed pad {} from efpmux",
+                        i
+                    ))
+                })?;
+
+            let data_src = data_input.static_pad("src").ok_or_else(|| {
+                BlockBuildError::ElementCreation(format!("data identity {} has no src pad", i))
+            })?;
+
+            data_src.link(&embed_pad).map_err(|e| {
+                BlockBuildError::LinkError(format!("data input {} -> efpmux: {:?}", i, e))
+            })?;
+
+            info!(
+                "EFP data input {}: linked to efpmux ({})",
+                i,
+                embed_pad.name()
+            );
+
+            data_elements.push((data_input_id, data_input));
+        }
+
         let mut internal_links = vec![];
         let mux_weak = mux.downgrade();
         let mut elements = vec![(mux_id.clone(), mux), (sink_id.clone(), srtsink)];
+        elements.extend(data_elements);
 
         // Create video input chains with dynamic parser insertion
         if num_video_tracks > 0 {
@@ -549,8 +633,8 @@ impl BlockBuilder for EfpSrtOutputBuilder {
         ));
 
         info!(
-            "Created EFP/SRT block with {} video track(s) and {} audio chain(s)",
-            num_video_tracks, num_audio_tracks
+            "Created EFP/SRT block with {} video track(s), {} audio chain(s) and {} data track(s)",
+            num_video_tracks, num_audio_tracks, num_data_tracks
         );
 
         Ok(BlockBuildResult {
@@ -775,6 +859,20 @@ fn efpsrt_output_definition() -> BlockDefinition {
                 mapping: PropertyMapping {
                     element_id: "_block".to_string(),
                     property_name: "num_audio_tracks".to_string(),
+                    transform: None,
+                },
+                live: false,
+                persist: None,
+            },
+            ExposedProperty {
+                name: "num_data_tracks".to_string(),
+                label: "Number of Data Tracks".to_string(),
+                description: "Number of EFP embedded-data input tracks (default: 0). Each track maps to an efpmux 'embed_%u' pad. The connected source must send 'application/x-efp-embedded' caps carrying 'data-type' and 'stream-id'; data is delivered alongside the media stream with the matching stream-id.".to_string(),
+                property_type: PropertyType::UInt,
+                default_value: Some(PropertyValue::UInt(0)),
+                mapping: PropertyMapping {
+                    element_id: "_block".to_string(),
+                    property_name: "num_data_tracks".to_string(),
                     transform: None,
                 },
                 live: false,
--- a/backend/src/blocks/builtin/efpsrt.rs
+++ b/backend/src/blocks/builtin/efpsrt.rs
@@ -36,6 +36,20 @@ use std::sync::Arc;
 use strom_types::{block::*, element::ElementPadRef, PropertyValue, *};
 use tracing::{debug, error, info, warn};
 
+/// Read a track-count property, returning `None` when it is absent, not a
+/// number, or negative, so the caller applies its own default.
+///
+/// `usize::try_from` rather than `as usize`: every track count feeds a
+/// `for i in 0..n` loop that allocates an element per iteration, and a negative
+/// `Int` cast with `as` becomes `usize::MAX`.
+pub fn track_count(properties: &HashMap<String, PropertyValue>, name: &str) -> Option<usize> {
+    properties.get(name).and_then(|v| match v {
+        PropertyValue::UInt(u) => usize::try_from(*u).ok(),
+        PropertyValue::Int(i) => usize::try_from(*i).ok(),
+        _ => None,
+    })
+}
+
 /// EFP/SRT Output block builder.
 pub struct EfpSrtOutputBuilder;
 
@@ -44,32 +58,11 @@ impl BlockBuilder for EfpSrtOutputBuilder {
         &self,
         properties: &HashMap<String, PropertyValue>,
     ) -> Option<ExternalPads> {
-        let num_video_tracks = properties
-            .get("num_video_tracks")
-            .and_then(|v| match v {
-                PropertyValue::UInt(u) => Some(*u as usize),
-                PropertyValue::Int(i) => Some(*i as usize),
-                _ => None,
-            })
-            .unwrap_or(1);
+        let num_video_tracks = track_count(properties, "num_video_tracks").unwrap_or(1);
 
-        let num_audio_tracks = properties
-            .get("num_audio_tracks")
-            .and_then(|v| match v {
-                PropertyValue::UInt(u) => Some(*u as usize),
-                PropertyValue::Int(i) => Some(*i as usize),
-                _ => None,
-            })
-            .unwrap_or(1);
+        let num_audio_tracks = track_count(properties, "num_audio_tracks").unwrap_or(1);
 
-        let num_data_tracks = properties
-            .get("num_data_tracks")
-            .and_then(|v| match v {
-                PropertyValue::UInt(u) => Some(*u as usize),
-                PropertyValue::Int(i) => Some(*i as usize),
-                _ => None,
-            })
-            .unwrap_or(0);
+        let num_data_tracks = track_count(properties, "num_data_tracks").unwrap_or(0);
 
         let mut inputs = Vec::new();
 
@@ -175,32 +168,11 @@ impl BlockBuilder for EfpSrtOutputBuilder {
             })
             .unwrap_or(DEFAULT_EFP_MTU);
 
-        let num_video_tracks = properties
-            .get("num_video_tracks")
-            .and_then(|v| match v {
-                PropertyValue::UInt(u) => Some(*u as usize),
-                PropertyValue::Int(i) => Some(*i as usize),
-                _ => None,
-            })
-            .unwrap_or(1);
+        let num_video_tracks = track_count(properties, "num_video_tracks").unwrap_or(1);
 
-        let num_audio_tracks = properties
-            .get("num_audio_tracks")
-            .and_then(|v| match v {
-                PropertyValue::UInt(u) => Some(*u as usize),
-                PropertyValue::Int(i) => Some(*i as usize),
-                _ => None,
-            })
-            .unwrap_or(1);
+        let num_audio_tracks = track_count(properties, "num_audio_tracks").unwrap_or(1);
 
-        let num_data_tracks = properties
-            .get("num_data_tracks")
-            .and_then(|v| match v {
-                PropertyValue::UInt(u) => Some(*u as usize),
-                PropertyValue::Int(i) => Some(*i as usize),
-                _ => None,
-            })
-            .unwrap_or(0);
+        let num_data_tracks = track_count(properties, "num_data_tracks").unwrap_or(0);
 
         // Create efpmux
         let mux_id = format!("{}:efpmux", instance_id);
@@ -867,7 +839,7 @@ fn efpsrt_output_definition() -> BlockDefinition {
             ExposedProperty {
                 name: "num_data_tracks".to_string(),
                 label: "Number of Data Tracks".to_string(),
-                description: "Number of EFP embedded-data input tracks (default: 0). Each track maps to an efpmux 'embed_%u' pad. The connected source must send 'application/x-efp-embedded' caps carrying 'data-type' and 'stream-id'; data is delivered alongside the media stream with the matching stream-id.".to_string(),
+                description: "Number of EFP embedded-data input tracks (default: 0). Each track maps to an efpmux 'embed_%u' pad. The connected source must send 'application/x-efp-embedded' caps carrying 'data-type' and 'stream-id'. Both fields default to 0 when omitted rather than failing, and stream-id 0 is reserved: data addressed to it, or to any stream-id that carries no media, is buffered by the muxer and never sent. Media stream-ids are allocated from 1 in pad order, so the video track is 1 and audio tracks follow.".to_string(),
                 property_type: PropertyType::UInt,
                 default_value: Some(PropertyValue::UInt(0)),
                 mapping: PropertyMapping {
--- a/backend/src/blocks/builtin/efpsrt_input.rs
+++ b/backend/src/blocks/builtin/efpsrt_input.rs
@@ -1,7 +1,8 @@
 //! EFP over SRT input block builder.
 //!
 //! This block receives an SRT stream carrying EFP (Elastic Frame Protocol) and demuxes
-//! it into separate video and audio output pads.
+//! it into separate video and audio output pads, plus an optional embedded-data
+//! output carrying efpdemux's `embedded` pad (default: 0 data tracks).
 //!
 //! Pipeline structure (decode=true, default):
 //! ```text
@@ -32,6 +33,9 @@ use std::sync::Arc;
 use strom_types::{block::*, element::ElementPadRef, FlowId, PropertyValue, *};
 use tracing::{debug, error, warn};
 
+/// Caps name of `efpdemux`'s embedded-data src pad.
+const EFP_EMBEDDED_CAPS_NAME: &str = "application/x-efp-embedded";
+
 /// EFP/SRT Input block builder.
 pub struct EfpSrtInputBuilder;
 
@@ -57,6 +61,15 @@ impl BlockBuilder for EfpSrtInputBuilder {
                 _ => None,
             })
             .unwrap_or(1);
+
+        let num_data_tracks = properties
+            .get("num_data_tracks")
+            .and_then(|v| match v {
+                PropertyValue::UInt(u) => Some(*u as usize),
+                PropertyValue::Int(i) => Some(*i as usize),
+                _ => None,
+            })
+            .unwrap_or(0);
 
         let mut outputs = Vec::new();
 
@@ -84,6 +97,16 @@ impl BlockBuilder for EfpSrtInputBuilder {
                 name: format!("audio_out_{}", i),
                 media_type: MediaType::Audio,
                 internal_element_id: format!("audio_output_{}", i),
+                internal_pad_name: "src".to_string(),
+            });
+        }
+
+        for i in 0..num_data_tracks {
+            outputs.push(ExternalPad {
+                label: Some(format!("D{}", i)),
+                name: format!("data_out_{}", i),
+                media_type: MediaType::Generic,
+                internal_element_id: format!("data_output_{}", i),
                 internal_pad_name: "src".to_string(),
             });
         }
@@ -179,6 +202,15 @@ impl BlockBuilder for EfpSrtInputBuilder {
                 _ => None,
             })
             .unwrap_or(1);
+
+        let num_data_tracks = properties
+            .get("num_data_tracks")
+            .and_then(|v| match v {
+                PropertyValue::UInt(u) => Some(*u as usize),
+                PropertyValue::Int(i) => Some(*i as usize),
+                _ => None,
+            })
+            .unwrap_or(0);
 
         // Create srtsrc
         let src_id = format!("{}:srtsrc", instance_id);
@@ -294,6 +326,23 @@ impl BlockBuilder for EfpSrtInputBuilder {
             elements.push((element_id, identity));
         }
 
+        // Create embedded-data output identity elements. efpdemux exposes a
+        // single `embedded` src pad for all data types, so only the first of
+        // these is ever linked.
+        let mut data_guards = Vec::new();
+        for i in 0..num_data_tracks {
+            let element_id = format!("{}:data_output_{}", instance_id, i);
+            let identity = gst::ElementFactory::make("identity")
+                .name(&element_id)
+                .build()
+                .map_err(|e| {
+                    BlockBuildError::ElementCreation(format!("data identity {}: {}", i, e))
+                })?;
+            let guard = Arc::new(AtomicBool::new(false));
+            data_guards.push((identity.downgrade(), guard));
+            elements.push((element_id, identity));
+        }
+
         // Setup dynamic pad linking on efpdemux pad-added.
         // - decode mode: video via h264parse + decoder, audio via opusdec + audioconvert + audioresample.
         // - passthrough mode: video via h264parse, audio linked directly.
@@ -325,6 +374,10 @@ impl BlockBuilder for EfpSrtInputBuilder {
             let is_audio = caps_name
                 .as_ref()
                 .map(|n| n.starts_with("audio/"))
+                .unwrap_or(false);
+            let is_data = caps_name
+                .as_deref()
+                .map(|n| n == EFP_EMBEDDED_CAPS_NAME)
                 .unwrap_or(false);
 
             debug!(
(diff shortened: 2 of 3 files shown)


borisasadanin:
Reporter of #691 here. Three things, one of which unblocks CI.

**1. Answering your open question: caps alone are enough. Please do not add per-track `data-type`/`stream-id` properties.**

You read the issue correctly. What we need is the caller-set `data-type` and `stream-id` on the pad caps, and your caps-driven approach gives us exactly that. A property-per-track scheme would be a second way to say the same thing, and the caps route has the advantage that the template makes both fields mandatory so a caller cannot leave them unset. Our answer here is the same shape as our answer on `MediaType`: please do not grow the API on our account.

**2. CI is currently red on formatting, and that is masking your evidence.**

`Check (Linux)` fails at `cargo fmt --all -- --check` (ci.yml:108), which runs before clippy and the test step, so `cargo test` never executed on this branch. The two-commit red-then-green structure is the right way to prove the test guards the fix, but right now CI has not actually demonstrated either half. A `cargo fmt` should be all it takes.

**3. Closing the CI worry we ourselves raised on the issue: it is a non-issue, and here is the evidence.**

On #691 we flagged that `efp` is an optional Cargo feature and that an integration test could pass forever while the shipped artifact has no EFP blocks. We should have checked before raising it. The workflow does cover it:

- `ci.yml:117` — `cargo test --package strom --features efp,nvidia`
- `ci.yml:111` — `cargo clippy --package strom --all-targets --features efp,nvidia -- -D warnings`
- `ci.yml:304` and `ci.yml:414` — release builds with `--features efp` for x86_64 and aarch64

So your new test will genuinely run once formatting passes. Worth noting the one asymmetry we did find: `ci.yml:500` runs `cargo test --package strom --features nvidia` without `efp`, so the EFP tests are Linux-job-only. That is the shipped platform, so we do not think it matters, but it is there.

The separate observation from our issue comment still stands and is unrelated to this PR: the `eyevinn-strom` image published in the OSC catalog reports 0.4.7 built 2026-04-08, and its `/api/blocks` contains no `builtin.efpsrt_input` or `builtin.efpsrt_output` at all. Whatever produced that image predates the feature or was built differently. Not your problem in this PR, but it means merging this will not make the pads reachable on OSC until a rebuild happens.

**On priority, since we are the reporter and it would be wrong not to say it:** #691 is the least urgent of the four we filed. #690 is the one that blocks us and has no PR yet. Please do not let this one crowd it out on our account.

borisasadanin:
Reporter of #691. Here is the exact CI fix, and a review of the diff since we specified this and know the plugin surface.

## The formatting failure, in full

Three hunks, all in `backend/tests/efpsrt_embedded_data_test.rs`, all pure line-wrapping. `cargo fmt` fixes every one:

**:108 and :195** — same shape in both tests:

```rust
let props = properties(&[
    ("num_video_tracks", 0),
    ("num_audio_tracks", 0),
    ("num_data_tracks", 2),   // 1 in the :195 case
]);
```

**:120** — chain wrap:

```rust
pads.iter()
    .map(|p| p.name().to_string())
    .collect::<Vec<_>>()
```

Nothing else in the branch is unformatted. Since `Check formatting` runs before clippy and test on that job, `cargo test` has not executed yet, so the red-then-green evidence is still unproven by CI rather than disproven.

## Review

**The `embed_{i}` explicit naming is the best thing in this diff.** `request_embed_pad` defaulting to `embed_{stream_id}` while `stream_id` is still 0 at request time is exactly the kind of thing that works with one track and fails mysteriously with two. Calling it out in a comment as well as fixing it is right.

### 1. Input side: `num_data_tracks > 1` creates external pads that can never carry data

This is the one worth changing before merge.

Output side, `num_data_tracks` is real: N tracks, N `embed_%u` pads. Input side, by your own comment (`efpsrt_input.rs`, "efpdemux exposes a single `embedded` src pad for all data types, so only the first of these is ever linked") and the property description, only track 0 is ever linked.

But `get_external_pads` still publishes `data_out_0..data_out_{N-1}`. So with `num_data_tracks: 2` on an input block, a user sees two data outputs in the flow graph, can wire `data_out_1` to something, and gets silence — no error, no warning, no log line, because the pad-added handler simply never reaches that guard.

Three ways out, your call:
- **clamp or reject** on the input side: treat `num_data_tracks` as a 0/1 flag there and error on >1 with a message saying why. Smallest and most honest.
- **rename the input property** so the asymmetry is visible in the API rather than only in the description.
- **demux by `stream-id`** into N pads, if `efpdemux` can actually produce more than one `embedded` pad.

Which raises the question I cannot settle from outside: does `get_or_create_embedded_pad` produce **one** pad for all embedded data, or one per `stream-id`? Your linking loop is written to handle several (it walks the guards and takes the first free one), but your comment says only one ever appears. If it is really one-per-stream-id then the loop is right and the comment is wrong; if it is one total, the loop is dead code past the first iteration and the external pads are misleading. Worth pinning down, because the two readings imply different fixes.

### 2. A negative `num_data_tracks` becomes a near-infinite loop

```rust
PropertyValue::Int(i) => Some(*i as usize),
```

`-1 as usize` is `usize::MAX`, and the next line is `for i in 0..num_data_tracks`. This is the established pattern in this file rather than something the PR invented, and the same cast guards `num_video_tracks` and `num_audio_tracks` already, so I am not asking you to fix the existing ones. But the PR adds four more copies of it, and `usize::try_from(*i).ok()` is the same length. Worth doing while the lines are being written.

### 3. Minor: what happens when the caller does not send the right caps

The output-side link is `data_input_{i}` identity src -> `embed_%u`, and the addressing depends on the caller sending `application/x-efp-embedded` caps with `data-type` and `stream-id`. If someone wires a source that sends anything else, this fails at caps negotiation, and the message a user will see is a generic link or negotiation failure rather than "your data source needs EFP embedded caps". Not a blocker; a sentence in the property description, or a caps filter with a clear name, would save someone an hour.

## Answering your open question again, since it is the one thing gating the design

Caps alone. No per-track `data-type`/`stream-id` properties. The caps route already makes both fields mandatory via the template, which a property pair would not.

Nothing here needs anything from us. Happy to re-read after the fmt push.

srperens:
Answering the one question you said you could not settle from outside, and correcting one
claim of mine that you repeated back.

## `efpdemux` produces one `embedded` pad in total, not one per `stream-id`

Read at the pinned tag (`backend/Cargo.toml:66` pins `gst-plugin-efp` to `v0.3.0`), so this
is the code Strom actually links against:

```rust
// gst-plugin-efp/src/efpdemux/imp.rs:144
embedded_pad: Mutex<Option<gst::Pad>>,
```

A single `Option<gst::Pad>`, not a map. And `get_or_create_embedded_pad` returns that one pad
for every `data_type` it is ever called with:

```rust
// gst-plugin-efp/src/efpdemux/imp.rs:676-686
fn get_or_create_embedded_pad(&self, data_type: u8) -> Result<gst::Pad, gst::FlowError> {
    {
        let guard = self.embedded_pad.lock().unwrap();
        if let Some(pad) = guard.as_ref() {
            return Ok(pad.clone());
        }
    }

    let templ = self.obj().pad_template("embedded").unwrap();
    let pad = gst::Pad::builder_from_template(&templ)
        .name("embedded")
```

The name is the literal `"embedded"` with no index, and the pad is cached at `:708`
(`*self.embedded_pad.lock().unwrap() = Some(pad.clone())`). `push_embedded` (`:648-649`) routes
all embedded data through it regardless of `data_type`.

**So of your two readings, "one pad total" is the correct one.** My comment in
`efpsrt_input.rs` was right and my linking loop is wrong: past the first iteration it is dead
code, and `get_external_pads` publishing `data_out_1..data_out_{N-1}` is exactly as misleading
as you describe. Your concern #1 is real, not hypothetical.

**This also rules out your third option.** "Demux by `stream-id` into N pads" is not available
at `v0.3.0` — it would take a change to `efpdemux` itself, not to Strom. That narrows the
choice to clamp/reject versus rename, which is a maintainer call on the block's API surface,
and I am not going to make it unilaterally. Either is a small diff once someone picks.

## Two further findings from the same read, both worth knowing before you build on this

**1. The `embedded` pad's caps are frozen at the first `data_type` seen.** `data_type` is used
only to stamp the caps of the pad at creation time (`:696-699`), and the early return at
`:678-681` means a later `data_type` never renegotiates. Embedded data of a second type flows
out of a pad still advertising the first type's `data-type`. That is an `efpdemux` behaviour,
not something this PR introduces or can fix.

**2. The demux-side caps carry no `stream-id` field at all** — only `data-type` (`:696-699`).
So the caps-driven addressing you and I agreed on is available on the *output* side, where
`efpmux`'s template has both fields, but on the *input* side a consumer cannot recover
`stream-id` from the pad caps.

## Correcting my own claim, which you repeated back

I wrote that the template caps "make both fields mandatory, so a caller cannot leave them
unset", and you took that as an advantage of the caps route over a property pair. It is weaker
than I made it sound. `efpmux` reads both fields defensively:

```rust
// gst-plugin-efp/src/efpmux/imp.rs:640-644
let sid = s.get::<i32>("stream-id").unwrap_or(0) as u8;
let dt  = s.get::<i32>("data-type").unwrap_or(0) as u8;
```

and again with `unwrap_or` on the renegotiation path (`:665-670`). Caps that omit a field
silently address `stream-id`/`data-type` 0 rather than failing. The template does declare both
as `[0,255]` ranges (`:316-318`), which stops a *non-fixed* caps event, but it does not give
you the "cannot be unset" guarantee I claimed. I have corrected the PR body. My recommendation
is unchanged — caps, not per-track properties — but it should rest on an accurate reason.

## On your point 2, the `-1 as usize` cast

You are right, and it is in lines this PR adds rather than lines it inherits.
`usize::try_from(*i).ok()` is the same length. I have not changed it in this push, because the
branch is deliberately two commits and its class A evidence is two concluded CI runs on those
exact SHAs; a fix means a force-push that resets both runs and demotes the PR to unverified
until new ones conclude. Happy to do it the moment it is bundled with whichever choice you make
on point 1, so it costs one force-push rather than two.

## What has not changed

No diff change in this comment's push. CI on `b7ed6eb` is green and the deliberate failure on
`b85d5f3` is the test step only, so the evidence in the body still stands as written.

On priority: understood, and noted in today's run log. I did not open anything new today.

borisasadanin:
Thank you for reading the plugin rather than reasoning about it, and for correcting your own claim unprompted. Both are worth more than the finding itself.

## Our preference on clamp versus rename, offered as a consumer and not as a decision

**Clamp or reject.** If `efpdemux` can only ever produce one `embedded` pad, then an input block that advertises `data_out_1` is advertising something the plugin cannot deliver, and no amount of naming makes that honest. A rename documents the trap; a clamp removes it.

Concretely we would reject `num_data_tracks > 1` on the input side at build time, with a message that says why, rather than silently clamping to 1. A silent clamp means a flow configured for 2 quietly behaves as 1, which is the same class of surprise in a quieter voice. It is your API surface and your call, but that is what we would want as the consumer.

## Your finding 2 matters to us more than finding 1 did

> the demux-side caps carry no `stream-id` field at all

That is the more consequential of the two for our use case, and it is worth stating plainly on the issue rather than leaving it in a PR thread.

The point of `stream-id` addressing, for us, is that a provenance blob has to say *which* stream it describes. Being able to set `stream-id` on the way in and not recover it on the way out means the addressing is write-only: a receiver gets bytes with a `data-type` and no way to attribute them to a media stream, which for provenance is most of the value gone. A single receiver that already knows there is only one stream is fine. Anything multi-stream is not.

That is an `efpdemux` limitation, not something this PR can fix, and it does not change our answer on the API shape here. But it does change how much we would build on the input side, and if the embedded channel is ever meant as a general sidecar channel rather than a one-stream one, it is the gap to close in the plugin.

Same for your finding 1, the caps frozen at the first `data_type`: worth knowing, not this PR's problem.

## On the `unwrap_or(0)` correction

Understood, and the correction changes our reasoning slightly without changing the conclusion. Caps are still the right route, but the argument is now "one mechanism instead of two" rather than "the caller cannot get it wrong". Since a caller *can* get it wrong and will silently address `stream-id` 0, that seems worth one sentence in the `num_data_tracks` property description on the output side: omitting either field addresses 0 rather than failing.

## On the cast, and on force-pushes

Your reasoning is right and we would have made the same call. Bundling it with whichever way point 1 lands costs one force-push instead of two, and the class A evidence on the current SHAs is worth preserving until there is a reason to spend it.

Nothing outstanding from us on this PR. The only thing gating it is a maintainer picking clamp or rename, and our preference above is input to that, not a substitute for it.

srperens (commented):
**Verdict: Request changes** — mechanism is right and already vetted at length in this thread,
but two correctness gaps in the diff should close now rather than after a follow-up decision.

**Requested changes**

1. Clamp or reject `num_data_tracks > 1` on the input side,
   `backend/src/blocks/builtin/efpsrt_input.rs`. I fetched the pinned plugin tag directly
   (`backend/Cargo.toml:66` pins `v0.3.0`) rather than trusting the thread's account: in
   `eyevinn/efp` at that tag, `efpdemux` holds exactly one
   `embedded_pad: Mutex<Option<gst::Pad>>` (`gst-plugin-efp/src/efpdemux/imp.rs`, line 144),
   and `get_or_create_embedded_pad` (same file, 676-710) returns that one cached pad, named
   literally `"embedded"`, for every `data_type`. So `num_data_tracks: 2` publishes
   `data_out_1` with no error and no way to ever link it. Shipping the unbounded `UInt` first
   means a released API surface has to be walked back later; clamp/reject is a small diff now.
2. Fix the four new `... as usize` casts on a negative `Int` (`efpsrt.rs`/`efpsrt_input.rs`,
   the sites this PR adds for `num_data_tracks`). `-1 as usize` becomes `usize::MAX`, feeding
   `for i in 0..num_data_tracks`. The body defers this to "bundle with whichever way point 1
   lands" to save a force-push, but `usize::try_from(*i).ok()` is the same length, and these
   four copies are new in this PR, not inherited.

**Claims**

|---|---|---|
| `efpdemux` exposes one `embedded` pad total, not one per track/stream-id | `CONFIRMED` | `eyevinn/efp@v0.3.0`, `efpdemux/imp.rs` line 144: `` `embedded_pad: Mutex<Option<gst::Pad>>` `` — read at the pinned tag, not from the thread |
| `efpmux` defaults unset `stream-id`/`data-type` to 0 rather than rejecting them | `CONFIRMED` | `eyevinn/efp@v0.3.0`, `efpmux/imp.rs` line 640: `` `let sid = s.get::<i32>("stream-id").unwrap_or(0) as u8;` `` |
| Existing flows unaffected (additive, default 0, feature-gated) | `CONFIRMED` | `backend/src/api/flows.rs:222` only production caller of `get_external_pads`; `backend/src/blocks/builtin/mod.rs:160-163` gates both builders on `#[cfg(feature = "efp")]` |
| New test drives real builders, not a reconstructed topology | `CONFIRMED` | `backend/tests/efpsrt_embedded_data_test.rs:19-20` imports `EfpSrtOutputBuilder` directly |
| Green CI at `b7ed6eb` actually ran the new test under `efp` | `CONFIRMED` | `gh pr checks 700`: `Check (Linux)` pass; `.github/workflows/ci.yml:117` (on this branch) runs `cargo test --package strom --features efp,nvidia` |

**Diagnosis** — Right layer: wires the block builders to pad templates the pinned plugin
already exposes, adds no plugin behavior. Coverage `BOUNDED`: output side reaches the full
channel; input side only reaches the first track, a boundary disclosed in prose but not
enforced in code (requested change 1).

**Radius** — `LOCAL`. Additive property/elements behind `#[cfg(feature = "efp")]`, default
`0` preserves every existing flow, no `strom-types`/`StromEvent`/`openapi.json` change (`git
diff origin/main...pr700 --name-only` touches only the two block files and the new test).

**Tests & CI** — Two-commit red→green genuinely executed: `b85d5f3` (test only) fails 4
assertions on `Check (Linux)`; `b7ed6eb` is green on both Linux builds, `API Contract Check`
and WASM. macOS/Windows `skipping` — `UNVERIFIED`, standard for this repo.

**Repo rules** — No CLAUDE.md violation beyond the two requested changes.
`MediaType::Generic` is reused (`types/src/element.rs:217`), not a new variant.

Confidence: HIGH

srperens:
Both requested changes are deferred deliberately this run, and no force-push was made. Reasoning, so a maintainer can overrule it in one reply:

**1. Clamp or reject `num_data_tracks > 1`** — this is a design decision no human has made. @borisasadanin offered a preference and then scoped it explicitly: "It is your API surface and your call, but that is what we would want as the consumer", and closed with "The only thing gating it is a maintainer picking clamp or rename, and our preference above is input to that, not a substitute for it." Picking clamp, reject or rename myself would be choosing a design, which this stage does not do. **Needs a maintainer.**

**2. The four `as usize` casts** — real, and not disputed. Held back only because fixing them alone costs a force-push that discards the concluded class A CI evidence on the current SHAs, for a change that will be rewritten by whichever way point 1 lands. That bundling was endorsed rather than assumed: "Bundling it with whichever way point 1 lands costs one force-push instead of two, and the class A evidence on the current SHAs is worth preserving until there is a reason to spend it."

So the ordering is: a maintainer picks clamp or reject or rename, then both land in one push. If a maintainer would rather have the cast fix now and spend the evidence, say so and I will push it on its own.

srperens:
**Maintainer decision: reject, not clamp, not rename.** Both requested changes are in `1345ace`, pushed as a third commit rather than a force-push — so the class A red→green evidence on `b85d5f3`/`b7ed6eb` is intact and did not have to be spent. Out of draft.

## 1. `num_data_tracks > 1` on the input side

Rejected at build time with `InvalidConfiguration`, naming the property and the reason. @borisasadanin's argument decided it: a flow configured for two tracks quietly behaving as one is the same surprise in a quieter voice. Rejecting is also the compatible direction — a rejection can be relaxed later, a silent clamp cannot be tightened.

`get_external_pads` cannot return an error, so it stops advertising the unreachable pads instead: an over-configured block shows only `data_out_0` in the graph, and `build` explains why it will not run. The limit is a named constant with the `efpdemux` reason next to it, so relaxing it is a one-line change when the plugin can demultiplex by stream ID.

## 2. The `as usize` casts

Fixed, and beyond the four the review asked for. All twelve track-count parse sites in both files — video and audio included — now go through one `track_count` helper using `usize::try_from`. The review scoped the inherited video/audio sites out, but they have the identical defect (`num_video_tracks: -1` → `usize::MAX` → a loop allocating a GStreamer element per iteration), and once the helper existed leaving them was worse than fixing them.

## 3. Two `efpmux` behaviours now stated in the output property description

Answering @borisasadanin's request for one sentence on the `unwrap_or(0)` correction, plus a related finding from reading the muxer at the pinned tag:

- Both caps fields default to 0 when omitted rather than failing.
- **`stream-id` 0 is reserved.** `StreamIdAllocator::new()` starts at `next: 1` and `release` asserts the id is non-zero, so no sink pad is ever allocated stream 0. Embedded data addressed to 0 — which is exactly what omitting the field gives you — lands in `pending_embeds[0]` and `sink_chain` never drains it. No error, no data, and an unbounded buffer. Same for any `stream-id` that carries no media.
- Media stream IDs are allocated from 1 in pad-request order, so the video track is 1 and audio follows. Nothing in the flow UI said this, which made caps-only addressing correct but unusable.

## Evidence

Ran locally on macOS with GStreamer 1.26.6 — the first execution evidence on this PR outside Linux CI, since the macOS job is skipped:

```
cargo test --features efp --test efpsrt_embedded_data_test
test result: ok. 9 passed; 0 failed
```

The three new tests are guards, not demonstrations. Reverting the three behaviours in the source and re-running, with the tests untouched:

```
---- a_negative_track_count_falls_back_to_the_default ----
  left: Some(18446744073709551615)   right: None
---- input_block_never_advertises_an_unreachable_data_output ----
  left: ["data_out_0", "data_out_1"]  right: ["data_out_0"]
---- input_block_rejects_more_data_tracks_than_the_demuxer_can_reach ----
  efpdemux has one embedded pad, so two data tracks must not build

test result: FAILED. 6 passed; 3 failed
```

Full backend suite with `--features efp` is green (525 unit + all integration targets), as are `cargo fmt --check`, `cargo clippy --all-targets -D warnings` and the OpenAPI snapshot.

Not verified, unchanged from the body: no wire round trip. The tests assert graph structure.

## What I am NOT fixing here, and where it goes instead

@borisasadanin — your finding 2, that the demux-side caps carry no `stream-id`, is the one that matters and I read the plugin down to the C++ to place it. **It is fixable, and it is ours to fix.**

`ElasticFrameProtocol.cpp:82` invokes the embedded-data callback as `(data, size, data_type, rPacket->mPts, ctx)` — while `rPacket->mStreamID` is fourteen lines further down at `:96`, same function, same packet. The upstream C API typedef (`efp_c_api/elastic_frame_protocol_c_api.h:57-58`) simply has no parameter for it. That is the entire reason `efp::EmbeddedData` has no `stream_id`, and therefore the entire reason `efpdemux` keeps one `embedded_pad: Option<Pad>` instead of a map. The addressing is write-only because of a dropped function argument, not because of the protocol.

So the plan, in `Eyevinn/efp` rather than here:

1. Carry `stream_id` through to the embedded callback, then give `efpdemux` one `embedded_%u` pad per stream ID with both `data-type` and `stream-id` in the caps. That also fixes your finding 1 — caps frozen at the first `data_type` — for free, and lets the limit above be relaxed to a real N.
2. Reject incomplete caps on an `embed_%u` pad instead of `unwrap_or(0)`, so the stream-0 black hole becomes a visible error rather than a silent leak.
3. Cover the send side. `efpmux`'s `embed_%u` pad has **no test at all** in the plugin — the only embedded test is `embedded_data_output`, which hand-encodes frames and exercises the demuxer alone. Strom would be its first consumer.
4. A mux→demux embedded round trip in the plugin's own CI. That closes your "no wire round trip" item without a deployed instance, so please do not hold it open waiting on the OSC version problem.

On OSC: agreed it is not this repository's, and thank you for chasing it down to the fork's `Dockerfile.osc` rather than leaving it as a suspicion here. Nothing further wanted from you on this PR.

Issue #685, opened 2026-08-21
Process aborts on gst_rtp_base_depayload hdrext assertion when video enters WHIP Input


srperens:
## Summary

Sending video into WHIP Input can abort the whole Strom process on a GStreamer assertion in the RTP depayloader base class. Observed twice today on macOS 15 (Darwin 25.6.0), Homebrew GStreamer 1.28.6, publishing from the built-in `/whip-ingest` page in a browser.

```
GStreamer-RTP:ERROR:gst-plugins-base/gst-libs/gst/rtp/gstrtpbasedepayload.c:942:
gst_rtp_base_depayload_handle_buffer:
'gst_buffer_list_length (priv->hdrext_buffers) == 0' should be TRUE
Bail out!
```

A glib `ERROR`-level assertion, so the process aborts (exit 134). Not catchable from Strom.

## Where it happens

The log immediately before the abort locates it precisely:

```
17:55:36.265  WHIP Input: Pad video_0 (stream 1) → appsink → slot 0 appsrc (video)
17:55:36.267  Element 'rtph264depay0' in pipeline 'test1': Null -> Ready -> Paused
17:55:36.548  Element 'h264parse0'    in pipeline 'test1': Null -> Ready -> Paused
              ** assertion, Bail out!
```

The depayloader that asserts is **`rtph264depay0` in the flow's main pipeline**, autoplugged by Strom's own `decodebin_video_0` — not an element inside `whipserversrc`. `whipserversrc` hands out `application/x-rtp`; Strom's `decodebin` does the depayloading and decoding.

That settles two things:

- The crash is on the **WHIP** side. A `do-retransmission` setting on `whepserversink` cannot affect it (see #663) — `whepserversink` has no depayloader at all.
- The header extension in play is TWCC, and **Strom configures it itself**: `WHIP Input: Set twcc-feedback-interval=200ms on rtpsession1` (`whip.rs:626-635`), alongside `drop-on-latency=true` on `rtpbin`. So Strom's own configuration is part of the trigger recipe, not just the publisher's.

## Reproduction status

**Intermittent.** It fired twice from the browser ingest page, then a later clean baseline run with the same page and the same flow did not crash. A headless `gst-launch` publisher (`whipclientsink`) never triggered it in any of six runs — with pre-encoded H264 input the negotiated SDP carried no `a=extmap` at all, so the header-extension path is never exercised; forcing the encoder path did produce an extmap but still no assertion.

[...]

PR #701, opened 2026-08-26
fix(whip): default do_retransmission to false; correct the WHEP rationale

wagenet:
## What

- **WHIP Input**: `do_retransmission` now defaults to `false`, diverging from `whipserversrc`'s own default of `true`.
- **WHEP Output**: no behaviour change — still defaults to `true`. Only its comment and property description change.

## Why

The gst-plugins-base abort this property works around ([gstreamer#5057](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/5057)) is a `g_assert` in `gst_rtp_base_depayload_handle_buffer` (`gstrtpbasedepayload.c`), and it lives in `GstRTPBaseDepayload`. Only a depayloader can arm it, and only `rtph264depay` / `rtph265depay` / `rtpklvdepay` call `gst_rtp_base_depayload_delayed()` at all.

In Strom that is exactly one element: the `rtph264depay` that `decodebin` autoplugs behind WHIP Input's `appsrc` — which is where the crash was observed.

That has two consequences:

**WHIP Input must default off.** Leaving RTX on means a lossy remote publisher can abort the whole process, taking every flow on the server with it, and an operator only avoids that by knowing to set an obscure property. The trade is real but lopsided: losing RTX costs loss resilience on the ingest leg (each loss escalates to PLI → forced keyframe instead of a cheap resend), while a hard abort costs the server. Revert once upstream lands a fix.

**WHEP Output was never affected.** `whepserversink` is a send-only `webrtcsink`: the property maps to `do-nack` on a Sendonly transceiver and to `rtprtxsend`, and no depayloader is instantiated anywhere on that path, so the assertion is unreachable. The existing comment and property description claimed the crash applied there too; both now describe the knob as the ordinary bandwidth/quality tunable it actually is.

## Notes

- Supersedes #663, which introduced this property with the older "unconfirmed workaround" rationale. `parse_do_retransmission` already reached `main` via ca59002, so #663 can be closed.
- The exposed property's `default_value` and description were updated to match the parser, and `do_retransmission_defaults_to_true` was renamed accordingly.

## Test

`cargo test --lib do_retransmission` — 6 passed. fmt and clippy clean.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,12 +26,12 @@ tokio-stream = { version = "0.1", features = ["sync"] }
 futures = "0.3"
 tower-http = { version = "0.7", features = ["fs", "cors", "trace"] }
 cairo-rs = { version = "0.22", features = ["v1_16"] }
-gstreamer = { version = "0.25", features = ["v1_20"] }
-gstreamer-app = { version = "0.25", features = ["v1_20"] }
-gstreamer-controller = { version = "0.25", features = ["v1_20"] }
-gstreamer-net = { version = "0.25", features = ["v1_20"] }
-gstreamer-gl = { version = "0.25", features = ["v1_20"] }
-gstreamer-video = { version = "0.25", features = ["v1_20"] }
+gstreamer = { version = "0.25", features = ["v1_22"] }
+gstreamer-app = { version = "0.25", features = ["v1_22"] }
+gstreamer-controller = { version = "0.25", features = ["v1_22"] }
+gstreamer-net = { version = "0.25", features = ["v1_22"] }
+gstreamer-gl = { version = "0.25", features = ["v1_22"] }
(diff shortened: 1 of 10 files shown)


wagenet:
Closing as malformed — my mistake.

This was branched from a local `main` that is not an ancestor of `upstream/main`, so the PR picked up seven unrelated commits (recorder queues, CEF cache, macOS run loop, GL bridge) and went conflicting. Only the last commit was intended.

It is also redundant: I opened it believing #663 was superseded because `parse_do_retransmission` was already on main. That was true only of my local main — `upstream/main` has no `do_retransmission` property at all, so #663 is still the PR that introduces it, and it already carries the accurate rationale on both halves. Nothing here to salvage.

#663 stands on its own; see #685 for the crash this property is a candidate mitigation for.

PR #706, opened 2026-08-26
fix(mpegtssrt): shield parsers from the EOS a tsdemux program change emits

srperens:
Follow-up to #705, which made this failure visible. This one removes the cause.

## What happens

An MPEG-TS feed may re-signal its PMT shortly after the first one. `tsdemux` then drains the previous program: it pushes EOS out of the pad it is about to remove, and creates replacement pads for the new program.

Captured from a live feed with `tsdemux:5,baseparse:5,GST_EVENT:5`:

```
.689613  tsdemux69:video_0_0041  done adding pad
.723817  tsdemux69               Draining previous program
.725773  tsdemux69:video_1_0041  done adding pad
.727030  tsdemux69:audio_1_0042  done adding pad
.727043  tsdemux69:video_0_0041  Flushing out pending data
.727053  tsdemux69:video_0_0041  Pushing out EOS
.727102  h264parse251:sink       have event type eos event
.727283  h264parse251  error: No valid frames found before end of stream
```

The two programs are **38 ms** apart. That is under one frame at 25 fps, so the `h264parse` decodebin autoplugged for the short-lived pad has data but no complete access unit when EOS arrives, and `GstBaseParse` turns that into a fatal error.

The error aborts the pipeline's PAUSED → PLAYING transition even though `video_1_0041` and `audio_1_0042` are already up and the stream is perfectly healthy. The flow is left stuck in PAUSED for good. Whether it happens at all depends on how much video landed inside that window — which is why it presents as a random "sometimes the production starts, sometimes it doesn't".

Downstream, this is what left `whepserversink`'s signaller port closed and every WHEP offer answering 502. Two deployed instances (0.6.5 and 0.6.6) were affected; across one day, 5 of 9 flow starts hit it.

## The fix

Drop EOS at the sink pad of every parser `decodebin` autoplugs inside this block.

A bus handler cannot do it — the pipeline bin aborts the state change when the message reaches it, before any handler we can install runs. The event has to be stopped before the parser acts on it.

Dropping it is correct here rather than merely expedient:

- This input is a live SRT source that reconnects on its own (`auto-reconnect`, `keep-listening` both default true), so an EOS never means the stream ended. Teardown happens via the state change to NULL.
- `tsdemux` removes the pad immediately after pushing EOS regardless of what downstream does with it, and `decodebin` tears the chain down from that pad removal — nothing waits on the event. Covered by a teardown test.

Parsers are matched on the element metadata klass (`Codec/Parser/...`) rather than a name list, so a parser autoplugged for a future codec is covered too. The probe is `EVENT_DOWNSTREAM`, not `BUFFER` — it fires per event, off the hot path.

## Tests

Four tests in `mpegtssrt_input.rs`, all calling the changed functions directly:

- `parser_eos_is_fatal_without_the_shield` — control. Asserts the underlying condition still reproduces; if it stops failing, the guard below proves nothing.
- `shield_keeps_transient_parser_eos_from_failing_the_pipeline` — the guard. Same pipeline with the shield applied posts no error.
- `shielded_pipeline_still_tears_down` — a swallowed EOS must not leave anything waiting; asserts the pipeline still returns to NULL.
- `is_parser_factory_matches_parsers_not_demuxers` — `h264parse`/`aacparse` match, `tsdemux` does not (it is handled by the other branch and is the element that legitimately emits the EOS).

The bed is `fakesrc num-buffers=1 sizemax=200 can-activate-pull=false ! video/x-h264 ! h264parse ! fakesink` — one buffer of junk then EOS, no complete access unit. No fixture file, no SRT, deterministic, ~20 ms.

`can-activate-pull=false` matters and is commented in the test. Without it `fakesrc` offers pull scheduling, `GstBaseParse` runs its own loop and reports the same message from `gst_base_parse_loop()` having never received an EOS event — a different code path from production. The test asserts the error's debug string contains `gst_base_parse_sink_event_default`, so it fails loudly if it ever drifts onto the wrong path again. That check caught exactly this while developing the test.

`h264parse`/`aacparse` are in `gstreamer1.0-plugins-bad` and `tsdemux` likewise, all already in the CI package list, so these execute in CI rather than silently skipping.

Ran locally: full `cargo test` (534 unit + all integration suites, 0 failed), `cargo fmt --check`, `cargo clippy --all-targets`. All clean.

## What this does not cover

The fix is inside the MPEG-TS/SRT input block. Any other block that puts a parser behind a demuxer with dynamic pads could hit the same thing; nothing generic guards against it.

### DIFF
--- a/backend/src/blocks/builtin/mpegtssrt_input.rs
+++ b/backend/src/blocks/builtin/mpegtssrt_input.rs
@@ -231,7 +231,9 @@ impl BlockBuilder for MpegTsSrtInputBuilder {
                 .build()
                 .map_err(|e| BlockBuildError::ElementCreation(format!("decodebin: {}", e)))?;
 
-            // Catch tsdemux when decodebin creates it internally, and set its properties.
+            // Catch elements decodebin creates internally:
+            //  - tsdemux: apply our latency / ignore-pcr settings
+            //  - parsers: shield them from the transient EOS a program change emits
             let instance_for_deep = instance_id.to_string();
             element.connect("deep-element-added", false, move |args| {
                 let added: gst::Element = args[2].get().unwrap();
@@ -246,6 +248,8 @@ impl BlockBuilder for MpegTsSrtInputBuilder {
                             "MPEGTSSRT Input {}: Set tsdemux latency={}ms, ignore-pcr={} (inside decodebin)",
                             instance_for_deep, tsdemux_latency, ignore_pcr
                         );
+                    } else if is_parser_factory(&f) {
+                        shield_parser_from_transient_eos(&added, &instance_for_deep);
                     }
                 }
                 None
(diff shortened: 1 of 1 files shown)


srperens:
## Validated against the real feed — this breaks pad exposure. Do not merge.

Deployed the CI binary from this branch to a dev instance and started a flow with two MPEG-TS/SRT inputs.

**The shield does fire, on exactly the right element:**

```
15:10:29.655  MPEGTSSRT Input b-input-1-9d0e28ed: dropped EOS on h264parse0 (transient demuxer pad teardown)
15:10:29.983  state after start: result=Ok(Async), current=Paused, pending=Playing
```

No `No valid frames found before end of stream`, and `Ok(Async)` instead of `Err(StateChangeError)`. The WHEP endpoints registered and their ports were genuinely listening — the thing that was dead before.

**But the input never recovers.** `b-input-1`'s decodebin configured its tsdemux and then never emitted `pad-added`:

```
MPEGTSSRT Input b-input-1-9d0e28ed: Set tsdemux latency=0ms, ignore-pcr=true (inside decodebin)
<no "Linked video pad" / "Linked audio pad" ever>
```

while SRT was actively delivering to that input:

```
b-input-0-9d0e28ed:srtsrc  connected=True pkts_rx=36060 bytes_rx=49039344 rate=3.00 Mbps
b-input-1-9d0e28ed:srtsrc  connected=True pkts_rx=14666 bytes_rx=19943504 rate=1.22 Mbps
```

`b-input-0` linked both pads normally. The pipeline sat at `Ready -> Paused (pending: Playing)` and never reached PLAYING.

In the unpatched failure the input *did* re-link its pads a second or two after the fatal error. So dropping the EOS stops decodebin from completing the removed chain's teardown, and the replacement program's pads are never exposed.

## Assessment

The diagnosis in the description still holds — it is confirmed by the capture. The **fix** is wrong: it converts a loud, recoverable-by-restart failure into a silent hang with no error at all, which is worse for an operator.

I called this out as the untested risk when the PR went up ("whether decodebin's internal bookkeeping is fine when a chain's parser never sees EOS ... reasoned about, tested only in a simplified harness"). The synthetic `fakesrc ! h264parse ! fakesink` bed exercises the right baseparse code path but contains neither `decodebin` nor `tsdemux`, so it could not have caught this. Any replacement needs a test with decodebin in it.

Moving to draft. The EOS must not be swallowed — the transient error needs to be made non-fatal some other way, or the parser must be kept off the short-lived pad in the first place.

srperens:
Superseded by #707. Same diagnosis, but this approach is dead: the EOS is what drives decodebin's teardown of the removed chain, so swallowing it stops the replacement pads from ever being exposed. Measured on the real feed — see the comment above. #707 re-drives the poisoned state change instead and touches nothing in the data path.

PR #707, opened 2026-08-26
fix(pipeline): re-drive a transition poisoned by a transient startup error

srperens:
Replaces #706, which was validated against the real feed and did not work. Same diagnosis, different fix.

## Diagnosis (unchanged, captured live)

An MPEG-TS feed re-signals its PMT shortly after the first one. `tsdemux` drains the previous program: it pushes EOS out of the pad it is removing, and the parser `decodebin` autoplugged for that short-lived pad has no complete access unit yet, so `GstBaseParse` errors fatally.

```
.689613  tsdemux69:video_0_0041  done adding pad
.723817  tsdemux69               Draining previous program
.725773  tsdemux69:video_1_0041  done adding pad
.727053  tsdemux69:video_0_0041  Pushing out EOS
.727283  h264parse251  error: No valid frames found before end of stream
```

38 ms — under one frame at 25 fps. Whether it fires depends on how much video landed in that window, which is why it looks random. 5 of 9 starts one day; two deployed instances (0.6.5 and 0.6.6).

## Two approaches measured and rejected first

**Dropping the EOS at the parser** (#706). The shield fired on the right element and the error went away, but `decodebin` then never emitted `pad-added` for the new program. The input never linked its pads while SRT kept delivering:

```
b-input-0-9d0e28ed:srtsrc  connected=True pkts_rx=36060 bytes_rx=49039344 rate=3.00 Mbps
b-input-1-9d0e28ed:srtsrc  connected=True pkts_rx=14666 bytes_rx=19943504 rate=1.22 Mbps   <- shielded input, no pads
```

The pipeline hung at `Ready -> Paused (pending: Playing)`. The EOS is what drives decodebin's teardown of the removed chain — swallowing it trades a loud failure for a silent hang.

**Intercepting the ERROR message.** Measured directly:

```
without bus sync handler: result=Err(StateChangeError) current=Ready pending=Playing
with    bus sync handler: result=Err(StateChangeError) current=Ready pending=Playing
```

Dropping the message changes nothing, because the erroring element aborts its own state rather than the message aborting it at the bin. A `GstBin` subclass overriding `handle_message` would not have helped.

## This fix

The input recovers by itself — decodebin drops the doomed chain and exposes the new program's pads within a second or two. Only the pipeline's state stays poisoned. So re-drive the transition, polling every 200 ms against a 3 s deadline, and report the failure if it still will not go. Polling rather than sleeping out a guess: the only way to learn that the input has settled is to try, so the recovery is picked up within 200 ms of it happening instead of at the next multiple of a long settle time, and the window still outlasts the 1.4-2.1 s relink measured on the live feed.

Reaching PLAYING is not by itself proof of recovery, so the re-drive verifies the data path before believing the state — see the review round below.

Measured beforehand: a failed state change re-driven after the failing condition clears returns `Ok(Success) current=Playing`, and needs neither a detour via READY nor a drained bus.

**This touches nothing in the data path.** That is what makes it safe — it cannot cause the wedge #706 caused. Worst case it costs a few seconds before reporting the same error.

It is a recovery, not a prevention: the demuxer churn that creates the doomed parser is upstream of us and still happens. The retry is loud (`warn!` per attempt, `info!` on recovery) so it stays visible rather than masking a feed that has started misbehaving.

## Tests

`pipeline_start_retry_test.rs` — the guard, in two halves.

- *Recovery must work.* A doomed chain (`audiotestsrc ! identity error-after=1 ! fakesink`) alongside a healthy one, both in the same pipeline. `identity` errors on the first buffer during preroll and stops being a problem afterwards, as decodebin dropping the doomed chain does, and the healthy chain is the live data path the real graph keeps across the transient error. Asserts PLAYING **and** that buffers actually reach the surviving sink.
- *A dead data path must fail.* The doomed chain on its own. The re-driven transition still reports Success there, so this asserts `start()` returns an error.

Verified each half fails when its own fix is reverted: with the retry window set to 0 the recovery test fails with the production signature, and with the liveness check disabled the dead-path test fails.

```
start() gave up on a pipeline that a retry recovers:
StateChange("State change failed: StateChangeError - current: Paused, pending: Playing")
```

`pipeline_start_failure_test.rs` (from #705) needed realigning and this is worth a reviewer's eye. Its bed was `identity error-after=1` — a *transient* error, which this retry now recovers, so its "must fail" premise no longer held and it failed on the full suite run. It moves to `fakesink state-error=paused-to-playing`, which no retry can rescue, so it now guards what it was always meant to: a pipeline that can never reach PLAYING must not be reported as started.

Ran locally, on the final commit: full `cargo test` (all suites green, 0 failures), `cargo clippy --tests`, `cargo fmt --check`. Nothing skipped.

## Review round: what a code review of #705 + #707 found

Reviewed both changes together. Six findings, five fixed in `6a0358f`.

**The re-drive declared recovery from the pipeline state alone, and accepted a pipeline whose data path was already dead.** A `basesrc` that takes a flow error pauses its task *and pushes EOS*. That EOS completes the sink's preroll, so the re-driven `set_state(Playing)` returns Success on a pipeline that will never carry another buffer. Measured on `audiotestsrc ! identity error-after=1 ! fakesink`: `start()` returned `Ok(Playing)` with **zero** buffers reaching the sink over 1.5 s, after which `start_flow()` would register the WHEP endpoints, set `auto_restart` and leave a flow that produces nothing for the life of the process — the silent failure #705 removed, back for this class of input error. Note this was also the retry test's own bed, so the test asserted the regression.

`start()` now checks the outputs before believing the state, on the retry path only: if **every** sink element has EOS on all its sink pads, the flow has no live data path and the start fails. One EOS sink among several is the doomed chain draining while the rest of the graph runs, which is exactly what the re-drive is for. A pipeline that started first time is not suspect and is not checked, so the normal start path is untouched.

**The failed-start path in `start_flow()` dropped the manager instead of calling `stop()`.** `Drop` only aborts the thumbnail task, stops probes and sets NULL. It does not `remove_bus_watch()`, remove the thread-priority handler, or unregister the flow's threads — so each failed start leaked a bus watch GSource onto the main context and left dead TIDs in the CPU-monitor registry, where a reused TID gets attributed to a flow that no longer exists. Worse, it left the flow's Media Player instances in the global registry, and that `Arc`'s `Drop` is the only thing that takes a Media Player block's *internal* pipeline to NULL: a Media Player flow that failed to start left a decoding pipeline and its file descriptors running for the life of the process. Now it unregisters and stops, mirroring `stop_flow()`. The same unregister is also done when `PipelineManager::new()` itself fails — same leak, but it predates these PRs.

**`manager.start()` blocked a tokio worker thread for seconds.** It is awaited straight from the HTTP handler and from the boot auto-restart loop, and the retry loop added seconds of `sleep` + synchronous `set_state` on top. Now on the blocking pool, as is the cleanup `stop()`.

**The WHEP not-registered log lines sat at `warn!` on an unauthenticated route,** keyed by a percent-decoded path segment: one log line per request for anyone who can reach the port, and a decoded newline could forge a line. Now `debug!`, with the value escaped.

**Not fixed, deliberately: the 500 ms state sample.** A transition still in flight comes back as `Ok(Async)` and is reported as started, so a fatal error arriving after the sample is not seen by `start()`. Waiting the transition out instead would break every input that legally sits in PAUSED until its first packet — an `srtsrc` or `rtspsrc` with no sender yet must start successfully and pick the feed up later. Such an error still reaches the runtime through the bus handler as a `PipelineError` event; it just does not fail `start()`. The reasoning is now a comment in the code rather than a silent choice.

## Not yet done

Not yet validated against the real feed. The loop for that is established — CI artifact, bind-mount the binary, start the flow, read the log — and #706 shows why that step is not optional.

### DIFF
--- a/backend/src/api/whep_player.rs
+++ b/backend/src/api/whep_player.rs
@@ -483,7 +483,12 @@ pub async fn whep_endpoint_proxy(
         None => {
             // Not registered: the flow is stopped, never started, or the
             // block's endpoint_id was changed since the player loaded.
-            tracing::warn!("WHEP offer for unregistered endpoint '{}'", endpoint_id);
+            //
+            // debug!, not warn!: the WHEP router carries no auth middleware and
+            // endpoint_id is a percent-decoded path segment, so anyone who can
+            // reach the port would otherwise get a log line per request out of
+            // us. {:?} escapes the value — a decoded newline could forge a line.
+            tracing::debug!("WHEP offer for unregistered endpoint {:?}", endpoint_id);
             return Response::builder()
                 .status(StatusCode::NOT_FOUND)
                 .header(header::ACCESS_CONTROL_ALLOW_ORIGIN, "*")
@@ -633,8 +638,10 @@ pub async fn whep_resource_proxy_delete(
     let port = match state.whep_registry().get_port(&endpoint_id).await {
         Some(p) => p,
         None => {
-            tracing::warn!(
-                "WHEP resource DELETE for unregistered endpoint '{}'",
+            // debug! and {:?} for the same reason as the offer path above:
+            // unauthenticated route, attacker-supplied endpoint_id.
+            tracing::debug!(
+                "WHEP resource DELETE for unregistered endpoint {:?}",
                 endpoint_id
             );
             return Response::builder()
--- a/backend/src/gst/pipeline/lifecycle.rs
+++ b/backend/src/gst/pipeline/lifecycle.rs
@@ -2,8 +2,22 @@ use super::{PipelineError, PipelineManager};
 use crate::gst::thread_priority;
 use gstreamer as gst;
 use gstreamer::prelude::*;
+use std::time::Duration;
 use strom_types::PipelineState;
-use tracing::{error, info};
+use tracing::{error, info, warn};
+
+/// How many times to re-drive a failed transition to PLAYING before giving up.
+/// One transient demuxer program change needs a single retry; the extra
+/// attempts cover a feed that churns twice. A genuinely broken pipeline just
+/// fails this many times and then reports the failure.
+const STATE_CHANGE_RETRIES: u32 = 3;
+
+/// How long to let the input settle before re-driving. Measured on a live feed,
+/// the input relinked its pads 1.4-2.1 s after the transient error.
+const STATE_CHANGE_RETRY_SETTLE: Duration = Duration::from_millis(750);
+
+/// How long to wait for the re-driven transition to report a verdict.
+const STATE_CHANGE_RETRY_TIMEOUT: gst::ClockTime = gst::ClockTime::from_seconds(2);
 
 impl PipelineManager {
     /// Start the pipeline (set to PLAYING state).
@@ -105,7 +119,7 @@ impl PipelineManager {
         // Query the actual state GStreamer has reached so far.
         // For Async pipelines this returns the current state (e.g. Paused)
         // without blocking. For NoPreroll/Success it returns the final state.
-        let (result, current_state, pending_state) =
+        let (mut result, mut current_state, mut pending_state) =
             self.pipeline.state(gst::ClockTime::from_mseconds(500));
         info!(
             "Pipeline '{}' state after start: result={:?}, current={:?}, pending={:?}",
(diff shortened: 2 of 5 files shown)


srperens:
## Validated on the dev instance — the retry fires, but the symptom is not fixed. Moving to draft.

```
15:55:41.494 ERROR No valid frames found before end of stream
15:55:41.494 state after start: result=Err(StateChangeError), current=Paused, pending=Playing
15:55:41.494 WARN  failed its transition — re-driving, attempt 1/3
15:55:41.914 MPEGTSSRT Input b-input-1: Linked video pad     <- input recovers
15:55:42.173 MPEGTSSRT Input b-input-0: Linked video pad
15:55:44.248 recovered on attempt 1 (current: Paused, pending: Playing)
15:55:44.248 Registering WHEP endpoint 'mv-9d0e28ed' on port 36633
```

Good: the retry mechanism works, and both inputs link — no repeat of the #706 wedge.

**Bad: `current=Paused`.** The re-driven transition returned `Ok(Async)`, not PLAYING. The pipeline never reached PLAYING, and nothing was listening on the registered ports:

```
$ ss -ltn | grep -E ":(36633|37505)\b"     # no output
```

That is the original 502 condition, and this patch makes it silent again — `start()` now returns Ok, so the honest failure #705 introduced is suppressed. As written this is a regression against #705, not an improvement on it.

## Two defects

1. `result.is_ok()` is the wrong success condition. Recovering from FAILURE to ASYNC is not recovery. The `recovered on attempt N` log line claims something it has not established.
2. Even with the transition un-poisoned and both inputs linked, the pipeline still did not reach PLAYING — so re-driving the state change is not sufficient on its own.

## Caveat on the test subject

The flow used has a `builtin.whip_input` with no publisher connected (`whipserversrc created per-session`, `max_sessions: 1`). Whether that alone prevents preroll is unverified, so it is possible this flow cannot reach PLAYING regardless of the patch and is therefore not a valid subject. Needs confirming against a flow known to start cleanly before drawing a final conclusion on point 2.

Point 1 stands regardless.

srperens:
**Note:** GitHub refuses `Request changes`/`Approve` review states on a PR authored by this same account (`addPullRequestReview: Can not request changes on your own pull request`), so posting this verdict as a plain comment instead of a formal review — treat it as if state=REQUEST_CHANGES.

**Verdict: Request changes**

1. CI: zero check-runs exist on the current head. `gh api repos/eyevinn/strom/commits/6a0358fa.../check-runs` returns `total_count: 0`; the only green run (32974496483) built commit `8e385102` — the *first* commit, before the "review round" fixes (spawn_blocking, cleanup-on-fail, data-path check) landed in `6a0358fa`. Zero checks is not evidence per protocol.
2. The new "verify data path" safety net can't see the exact failure this PR diagnoses, for the default (single-track) config — see Diagnosis. At minimum, extend `pipeline_start_retry_test.rs` to exercise a single shared multi-pad sink (mirroring `whepserversink`) instead of only two independent chains, or explicitly scope the PR body to "fixes the startup hang/502, does not guarantee video resumes."

**Claims**
|---|---|---|
| "decodebin drops the doomed chain and exposes the new program's pads" (recovery happens) | CONTRADICTED | `mpegtssrt_input.rs:356` `if guard.swap(true, Ordering::SeqCst) { continue; }` — claimed once, never reset; no `pad-removed`/reset handler anywhere in the file (grep confirmed). The replacement pad hits `mpegtssrt_input.rs:395-398` ("No available video output for pad") and is never linked. |
| `eos_sink_count()` treats an element as dead only if *every* sink pad is EOS | CONFIRMED | `lifecycle.rs:301-303` `pads.iter().all(\|pad\| pad.pad_flags().contains(gst::PadFlags::EOS))` |
| Default WHEP output (`whepserversink`) is one element carrying both `video_in` and `audio_in` | CONFIRMED | `whep.rs:1026` `gst::ElementFactory::make("whepserversink")`; `whep.rs:2400-2407` registers both `video_in`/`audio_in` on it |
| For the default 1-video/1-audio config, once video is orphaned by the guard bug, the shared sink's video pad stays EOS while audio keeps flowing, so `eos_sinks < total_sinks` and `start()` reports success | CONFIRMED (composition of the three rows above) | — |

**Diagnosis**
The retry+state-recheck is a correct, low-risk fix for the *state-machine* symptom (stuck PAUSED → false 502s) — `start()` only touches this path when `result.is_err()`, so healthy starts are unaffected. But its stated justification — that decodebin actually re-exposes and reconnects the new program — doesn't hold for the single-track default, because of the pre-existing guard-reuse bug above (not introduced by this diff, but decisive for whether this fix's promise is true). Net effect: this PR moves the silent failure from "pipeline stuck in PAUSED, WHEP 502s" to "pipeline reports PLAYING, one track permanently dark," for exactly the config in its own bug report. BOUNDED, not ABSOLUTE: audio-only tracks, or `num_video_tracks > 1` configs where a spare guard slot exists, likely do recover.

**Blast radius**
GLOBAL — `start()` is the one production path for every flow type (`state.rs:955`, confirmed sole non-test caller). The blind spot in `eos_sink_count()` applies to any topology where one sink element multiplexes independent legs, which is the default WHEP output shape, not an edge case.

**Tests & CI**
`pipeline_start_retry_test.rs`'s `build_recovering_flow` uses two fully independent chains (`audiotestsrc`/`identity`/`fakesink`) that never exercise a shared multi-pad sink or the pad-added/guard-reuse code path — it doesn't guard against the gap above. No canary applies (no new pipeline-lifecycle closures, no API types).

Not itself a reason to block, but worth flagging separately: the mpegtssrt_input.rs guard-reuse bug looks worth its own issue regardless of this PR's outcome.

Confidence: HIGH

srperens:
Closing this. The root cause turned out to be preventable at the source, which makes the retry containment for a symptom that no longer occurs — see #712.

The program change this PR recovers from is not the feed re-signalling its PMT, and not upstream of us. `mpegtsmux` advertises the first PAT/PMT it writes as `streamheader`, and `srtsink` replays that snapshot to every caller that connects afterwards. Our output block requests mux sink pads dynamically, so the header is frozen while only the video chain is linked — a video-only PMT that never updates while the live tables keep moving. Every receiver therefore gets a guaranteed program change a few packets into its connection. Captured from a sender running unrestarted for four weeks: three separate connects returned byte-identical first two packets, diverging only from packet three.

So the diagnosis in this PR's body is right about the chain and wrong about where it starts. #712 removes the trigger.

Two things found while reviewing this PR are worth recording, independent of the above:

**The liveness check would not have worked on the real graph.** It declares the input dead when *every* sink is EOS, on the reasoning that one EOS sink among several is the doomed chain draining. That assumes the doomed chain has its own sink. In the real graph it sits inside `decodebin`, upstream of a shared output, so a single-output flow has no case where the discriminator does its job — it either fails the start on the very case the re-drive exists for, or is trivially satisfied and proves nothing. The recovery test's bed used two independent chains with two sinks, which is why it did not catch this.

**`pipeline_start_retry_test` is flaky.** It passes in isolation, three runs for three, but fails under a full parallel `cargo test`.

The unrelated WHEP logging change from this branch is split out into #713, so it is not lost.

Still to do, tracked separately: the failure paths in `start_flow()` do not agree with each other on cleanup — the endpoint-conflict paths skip the Media Player registry unregister and the CPU deallocation that the other paths do. This PR fixed one of those paths and added a third variant rather than converging them. That wants one shared teardown, as its own change.

#705 stays as it is: it fixes no cause, but it turns a green-but-dead production into a visible start failure, which is worth keeping.

PR #721, opened 2026-08-27
fix(rtp): disable header extension aggregation on every depayloader

wagenet:
## Problem

A single corrupt RTP stream can abort the whole process, killing every unrelated flow on the server:

```
gstrtpbasedepayload.c:942:gst_rtp_base_depayload_handle_buffer:
'gst_buffer_list_length (priv->hdrext_buffers) == 0' should be TRUE
```

Since 1.24 `GstRTPBaseDepayload` caches each packet's RTP header while assembling an output buffer, clearing it only when the subclass pushes or flushes. In access-unit mode `rtph264depay` can absorb an interrupted fragmentation unit without pushing anything, so the cache stays populated and the non-defusable `g_assert_true` fires.

Upstream: [gstreamer#5057](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/5057), unfixed.

## Fix

Disable header-extension aggregation on every depayloader, restoring pre-1.24 behaviour: extensions are read from the current packet, so the cache never fills. Nothing is lost — Strom reads no header-extension metadata, and transport-cc, abs-send-time and mid are consumed by `webrtcbin` upstream of any depayloader.

`install()` sweeps depayloaders already present in a pipeline and connects `deep-element-added` for ones autoplugged later. The closure captures nothing, so it creates no reference cycle.

The abort kills the process, so every pipeline that can host a depayloader calls `install()` while still in NULL:

| Pipeline | How a depayloader gets there |
|---|---|
| Main (`PipelineManager::start()`) | `decodebin` autoplugs one, or a user places one directly in a flow |
| Media Player internal (decode + passthrough) | `uridecodebin` / `urisourcebin` autoplug one from an `rtsp://` URI |
| WHIP Input session | `whipserversrc` autoplugs them inside its own bin |

## Why resolve the symbol at runtime

The only switch is a C setter marked `Since: 1.24`, with no GObject property. Binding it would raise the build floor from 1.22 and break Debian 12 and the Raspberry Pi OS 12 cross-compile. Below 1.24 the runtime lookup fails and we do nothing, which is correct.

`dlsym(RTLD_DEFAULT, …)` on Unix. On Windows, `K32EnumProcessModules` walks the loaded modules and asks each for the export: `GetModuleHandleEx(FROM_ADDRESS)` returns the executable under MSVC (an imported function's address is the caller's import thunk), and the DLL name differs between MSVC and MinGW. The walk relies on `gstreamer-rtp` keeping the library loaded; `symbol_resolves_on_modern_gstreamer` fails if it stops.

A failed lookup on >= 1.24 means the abort is still live, so it logs at `warn!` with the GStreamer version; below 1.24 it is `info!`.

## Windows CI changes

These are here because without them a green `Build (Windows)` on this PR is not evidence. Any of the three can be split out.

**Masked exit codes.** Several `cargo` invocations shared one `pwsh` step, so only the last exit code counted and a failing test read as green. `Run Clippy`, `Run tests` and `Build` now run one command per step.

**Timeout.** The job was cut off at 45m06s in `Build MCP server` with every earlier step green. The ceiling is now 60; the passing run on this head took 45m50s, so the margin is thin until the release build moves to its own job.

**`volume_ramp_test` flake.** `unmute_after_mid_mute_fader_change_fades_in_from_zero` sampled 8 ms into a 30 ms fade, and a ~15 ms Windows timer tick could overshoot the whole fade. The fade is now 200 ms, sampled at 40 ms (0.003 against a 0.25 threshold); an unmute with no fade still fails it.

## Tests

Six unit tests plus one integration test, **all verified to fail when the fix is reverted**:

| Revert | What fails |
|---|---|
| `install()` neutered | the two `install_disables_aggregation_*` unit tests |
| `install()` call removed from `start()` | `rtp_hdrext_aggregation_test` |
| `install()` calls removed from the Media Player bridge | `decode_pipeline_disables_hdrext_aggregation`, `passthrough_pipeline_disables_hdrext_aggregation` |

The Media Player tests call the real `create_decode_pipeline` / `create_passthrough_pipeline`, then add a depayloader to a nested bin — the path `deep-element-added` sees for an autoplugged one, since CI cannot serve RTSP. `rtph264depay_enables_aggregation_by_default` proves the getter reads live state. `rtph264depay` is in `gstreamer1.0-plugins-good`, already in CI.

## What ran

**CI at `4d2e1d2`** ([run 34645083079](https://github.com/Eyevinn/strom/actions/runs/34645083079)), all eight checks green. I read the Linux, macOS and Windows job logs: all seven new tests and `unmute_after_mid_mute_fader_change_fades_in_from_zero` report `ok` on each, with no skips.

| Job | GStreamer | Lib tests |
|---|---|---|
| `Check (Linux)` | 1.24.2 | 590 passed |
| `Build (macOS)` | 1.28.6 | 585 passed |
| `Build (Windows)` | 1.26.10 | 580 passed |

**Locally, macOS / GStreamer 1.28**, at `97179f9`: full `cargo test --features efp,nvidia` with `STROM_REQUIRE_GST_PLUGINS=1`, `clippy --all-targets -D warnings`, `fmt --check`, all clean; the `volume_ramp_test` change separately. Later commits touch only that test and `ci.yml`.

**Not exercised anywhere:**
- The WHIP Input call site: `build_whipserversrc` binds a port and needs a full endpoint config. Inspection only.
- The < 1.24 no-op path: no CI image ships 1.22, and every test early-returns when `is_supported()` is false.


srperens (commented):
**Verdict: Comment** — sound mechanism and test discipline for the pipeline it touches, but the fix is narrower than its own framing ("disable ... on every depayloader"), and its platform-specific code is unverified by CI.

**Requested changes**
1. Also call `rtp_hdrext::install()` on the Media Player block's internal pipelines — `backend/src/blocks/builtin/mediaplayer/bridge.rs:28` and `:138` each build a separate `gst::Pipeline` running `uridecodebin`/`urisourcebin`, which will autoplug an RTP depayloader for an `rtsp://` source URI exactly like `decodebin` does in the main pipeline. That pipeline is never passed to `rtp_hdrext::install()`, so an interrupted RTSP H264 stream through a Media Player block still takes the whole process down with the bug this PR exists to fix.

**Claims**

|---|---|---|
| `install()` runs before any pipeline state change | CONFIRMED | `backend/src/gst/pipeline/lifecycle.rs:17` `rtp_hdrext::install(&self.pipeline);` runs before `set_state(Ready)` at `lifecycle.rs:65-67` and `set_state(Playing)` later |
| `deep-element-added` closure captures nothing, no ref cycle | CONFIRMED | `backend/src/gst/rtp_hdrext.rs:177` `bin.connect("deep-element-added", false, move \|args\| {` reads only `args`, no external variable moved in |
| Covers "every depayloader" the process runs | BOUNDED/CONTRADICTED as stated | Only reaches `self.pipeline`'s bin tree (`rtp_hdrext.rs:168` `iterate_recurse()` + the signal above). Verified two other independent `gst::Pipeline` instances exist and are never passed to `install()`: WHIP Input's `session_pipeline` (`backend/src/blocks/builtin/whip.rs:570`) and Media Player's internal pipelines (`bridge.rs:28`, `:138`). The WHIP one is fine — issue #685 itself traced the crash to the *main*-pipeline `decodebin`, not an element inside `whipserversrc` — but Media Player's `uridecodebin`/`urisourcebin` is a real, unguarded second path |
| `libgstrtp-1.0` is already loaded so `dlsym(RTLD_DEFAULT, ...)` finds the symbol | EXTERNAL | Dynamic-linker behavior, not settleable from this repo; the new `gstreamer-rtp` workspace dependency (`Cargo.toml:33`, `backend/Cargo.toml:58`) does cause the lib to link at load time on ELF/Mach-O, which supports the assumption |
| Windows `GetModuleHandleExA`/`GetProcAddress` path and the Unix `dlsym` path both work as written | UNVERIFIED | `Build (macOS)` and `Build (Windows)` both show `skipping` on this PR's checks (head `c7db33a8`); per CLAUDE.md/CI, neither builds on push/PR, so this is the only new platform-specific code in the diff and it has never compiled |

**Diagnosis** — Root cause and mechanism match issue #685's own confirmed investigation (interrupted FU-A with contiguous sequence numbers on a depayloader that has the X-bit-triggered aggregation cache populated); the "why runtime dlsym, not a feature bump" tradeoff is well-argued and consistent with the workspace's `v1_22` floor (`Cargo.toml:29`). The gap isn't the mechanism, it's coverage: BOUNDED to pipelines the code actually calls `install()` on, and today that's one of at least three that host GStreamer elements capable of autoplugging an RTP depayloader.

**Tests & CI** — Linux x86_64/ARM64, Check, WASM, API Contract all pass on `c7db33a8`. The new unit tests and `rtp_hdrext_aggregation_test.rs` genuinely exercise `PipelineManager::start()` (not a reimplementation) and read as revert-sensitive per the PR's own table. macOS/Windows: skipped, so the platform-specific FFI is unverified — dispatch with `gh workflow run ci.yml --ref wagenet/685-hdrext-aggregation -f platforms=macos` (and `=windows`) before merge.

Confidence: HIGH

wagenet:
Pushed `55e91ef` and rewrote the PR description. Both points from the review are addressed.

**Coverage.** The review was right that "every depayloader" overstated the diff — `install()` only ever ran on the main pipeline. It now also runs on the Media Player block's two internal pipelines (`bridge.rs`, decode and passthrough), where an `rtsp://` URI makes `uridecodebin`/`urisourcebin` autoplug a depayloader.

I went one further and covered WHIP Input's `session_pipeline` too, which the review judged out of scope. The reasoning there was that #685 traced the crash to the main-pipeline `decodebin`, not to anything inside `whipserversrc` — true, but `whipserversrc` does autoplug depayloaders in its own bin, and `g_assert_true` aborts the process rather than the flow. Which pipeline hosts the offending depayloader doesn't change the outcome, so leaving it out would keep the title overstating the fix. It is one line, and `install()` is a no-op below 1.24.

**Platform verification.** `Build (macOS)` and `Build (Windows)` skip on push/PR by design (`if: github.event_name == 'workflow_dispatch'`), so the new FFI would have merged without ever compiling. `workflow_dispatch` on this repo needs admin, so I dispatched it on the fork where the branch lives: [run 33179848797](https://github.com/wagenet/strom/actions/runs/33179848797) on `55e91ef`, all green. Both jobs ran Clippy and the test suite on their own OS, so the Windows `GetModuleHandleEx`/`GetProcAddress` path has now compiled and linked, and the `rtp_hdrext` tests executed on macOS and Windows as well as Linux. That run lives on the fork and won't appear in this PR's check list.

**Tests.** Two new unit tests call the real Media Player constructors, then add a depayloader to a nested bin — the same path `deep-element-added` sees for an autoplugged one, since CI can't serve RTSP. Both fail with the `install()` calls removed; I verified that by commenting them out rather than assuming it. Locally: 543 lib tests, all integration tests, `clippy --all-targets -D warnings`, `fmt --check`, all clean.

Two gaps left, both now stated in the description: the WHIP call site has no test (`build_whipserversrc` binds a port and needs a full endpoint config), and the < 1.24 no-op path is untested anywhere, since no CI image ships 1.22 and every test early-returns when `is_supported()` is false.

<sub>Written by Claude Code on behalf of @wagenet.</sub>

srperens (commented):
**Supersedes my 2026-08-28 v3 review (id 5049165959, COMMENTED) at head `c7db33a8`. That review requested one change: also call `rtp_hdrext::install()` on the Media Player's internal pipelines. This head does that, plus adds it to WHIP Input for completeness. Verdict changes to Comment→Comment (same), gap closed.**

**Verdict: Comment** — the requested-change gap is closed with genuine coverage; only the pre-existing macOS/Windows CI gap remains, which blocks approval per protocol regardless of the fork run cited in the PR body.

**Claims**

|---|---|---|
| Media Player's two internal pipelines now call `install()` before any state change | CONFIRMED | `backend/src/blocks/builtin/mediaplayer/bridge.rs:129` and `:326`, both immediately before `Ok(pipeline)`; the pipeline is only later moved to `Playing` from `connect_main_pipeline_handler`, invoked asynchronously from `builder.rs:236-247` after construction returns — install always precedes any state transition |
| WHIP Input's `session_pipeline` now calls `install()` before `Playing` | CONFIRMED | `backend/src/blocks/builtin/whip.rs:1040` `rtp_hdrext::install(&session_pipeline);` sits between the `.add(&whipserversrc)` call and `.set_state(gst::State::Playing)` at line ~1046 |
| New Media Player tests exercise the real constructors, not a reimplementation | CONFIRMED | `bridge.rs:689` and `:700` call `create_decode_pipeline(...)` / `create_passthrough_pipeline(...)` directly, then add a depayloader to a nested `gst::Bin` — the same path `deep-element-added` observes for an autoplugged element — and assert `rtp_hdrext::is_enabled(&depay) == Some(false)` |
| `gstreamer_rtp::RTPBaseDepayload` and `rtp_hdrext::is_enabled`/`is_supported` used by the new tests exist and match call signatures | CONFIRMED | `backend/Cargo.toml:58` `gstreamer-rtp.workspace = true`; `backend/src/gst/rtp_hdrext.rs:142,150,159` define `is_enabled`, `is_supported`, `install` with matching signatures |
| macOS/Windows FFI path now verified by this repo's CI | CONTRADICTED as stated | `gh pr checks 721` at head `55e91efa` still shows `Build (Windows)` and `Build (macOS)` as `skipping`, unchanged from the prior review. The green run cited in the PR body (`github.com/wagenet/strom/actions/runs/33179848797`) is on the author's fork, not `Eyevinn/strom` — it is not this repo's CI |

**Diagnosis** — unchanged from my prior review; mechanism and root cause still check out.

**Radius** — GLOBAL, as before (pipeline lifecycle, three independent `gst::Pipeline` hosts), now fully covered rather than bounded to one.

**Tests & CI** — `Build (Linux x86_64/ARM64)`, `Check (Linux)`, `Check & Build (WASM)`, `API Contract Check` all pass on `55e91efa`. `Build (macOS)`/`Build (Windows)` still skip on push/PR per repo config — this remains the only new platform-specific code (`dlsym`/`GetModuleHandleEx`) never compiled by this repo's own CI. Dispatch before merge: `gh workflow run ci.yml --ref wagenet/685-hdrext-aggregation -f platforms=macos` (and `=windows`).

Confidence: HIGH

wagenet:

srperens:
Not a review — the head is unchanged at `55e91efa`. What changed is that the blocker my standing review left open is now clearable.

#768 (merged as `4600c21`) makes `Build (macOS)` and `Build (Windows)` run on a pull request carrying a `ci:macos` / `ci:windows` label:

A correction to how I framed this PR earlier, though. Grouping it with the macOS PRs was loose: this diff contains no `target_os` gating at all — I checked, zero added lines match it. What it does add is a **new native dependency**, `gstreamer-rtp` in `Cargo.toml`, plus FFI use, and that binding surface has never been compiled anywhere but Linux in this repository. That is a different risk from platform-gated code and a weaker one, since a missing system library shows up as a build failure rather than as behaviour that only breaks on one platform.

So for this PR both labels are worth a moment's thought rather than just `ci:macos`: the question is whether `gstreamer-rtp` builds against Homebrew GStreamer and against the Windows toolchain, and `ci:windows` answers half of it. Neither is 22 minutes well spent if you would rather let the post-merge `main` run cover it — which is now a legitimate call under the revised `REVIEW.md`, precisely because the platform-specific surface here is a dependency edge rather than a code path.

srperens (dismissed):
**Supersedes my 2026-08-28 review at this same head (`55e91efa`, COMMENTED); that verdict does not stand: Comment → Request changes.** The diff did not change — the platform builds I asked for have now run, and the Windows one holds a failing test that the job reported as success.

**Verdict: Request changes** — `Build (Windows)` is green with `1 failed`. The workaround does not engage on Windows at all, and the diff's own guard caught it.

**Requested changes**

1. Make the Windows lookup resolve, in `#[cfg(windows)] lookup()`. `backend/src/gst/rtp_hdrext.rs:78` — `    let anchor = gstreamer_rtp::ffi::gst_rtp_base_depayload_get_type as *const c_void;` — SPECULATIVE (not verified): under MSVC the address of an imported function is the import thunk in the *calling* module, so `GetModuleHandleExA(FROM_ADDRESS, …)` hands back the test binary, not `gstrtp-1.0-0.dll`. Resolving the module by name does not depend on that.
2. If Windows ships unprotected instead, make the no-op loud and the message true. `backend/src/gst/rtp_hdrext.rs:111` — `                "{SETTER} not found (GStreamer < 1.24) - header extension aggregation \` — is the only runtime signal an operator gets, and on Windows/1.26.10 it misattributes a failed lookup to an old GStreamer. Report `gst::version()`, at `warn!`.
3. Separately: `.github/workflows/ci.yml:513-516` runs several `cargo test` invocations in one `pwsh` step, so only the last exit code reaches the step. That is why a red test read as green, and `Run Clippy` and `Build` are masked identically. Pre-existing on `main`, so a separate PR is fair — but until it lands, no Windows conclusion here is evidence.

**Claims**

|---|---|---|
| One binary works on both GStreamer generations via runtime resolution | `CONTRADICTED` | `backend/src/gst/rtp_hdrext.rs:32` — `//! GStreamer 1.22. Resolving the symbol at runtime keeps one binary working on` — vs `Build (Windows)` run 34221734567 on GStreamer 1.26.10: `GStreamer 1.26 has the aggregation API but the runtime lookup failed - the workaround would silently no-op` |
| macOS/Windows FFI path now compiled by this repo's CI | `CONFIRMED` | run 34221734567 at `55e91efa`: `Build (macOS)` and `Build (Windows)` both `success` — the one row that changed since my last review, via the `ci:macos`/`ci:windows` labels now on the PR |
| The guards assert on Linux and macOS rather than skipping | `CONFIRMED` | `Check (Linux)` (run 33179823537, GStreamer 1.24.2) and `Build (macOS)` (1.28.6) both log `symbol_resolves_on_modern_gstreamer ... ok`; only non-vacuous at ≥ 1.24, which pins `is_supported() == true` and defeats the early return in the other five |
| `Build (Windows)` passing means its tests passed | `CONTRADICTED` | same job, step `Run tests` conclusion `success`, log: `test result: FAILED. 574 passed; 1 failed` then `error: test failed, to rerun pass -p strom --lib` |
| Everything else about the diff | `CONFIRMED` | unchanged at this head; see my 2026-08-28 review (id 5064323651) |

**Diagnosis** — On Windows `backend/src/gst/rtp_hdrext.rs:160` — `    if !is_supported() {` — returns before both the sweep and the `deep-element-added` handler, so nothing is installed. Coverage is `BOUNDED`, and the bound is the platform, not the pipeline. Not a regression — Windows aborts today for the same reason — so the objection is the false doc claim and the red test.

**Radius** — `GLOBAL`, unchanged: pipeline lifecycle, four `gst::Pipeline` hosts. New is that the zero-coverage case is a whole platform, where the failure stays the original `g_assert` taking every unrelated flow down with the process.

**Tests & CI** — all eight checks hold a `success` conclusion across runs 33179823537 and 34221734567 at this head; I read all three job logs.

Confidence: HIGH

wagenet:

All three points addressed, and the branch is rebased onto `main` (`95d7ccb`) so the `ci:macos` / `ci:windows` labels run against current CI config. New head `97179f9`.

**1. Windows lookup.** Your hypothesis was right. The Windows log shows `symbol_resolves_on_modern_gstreamer` panicking on GStreamer 1.26 while the other five `rtp_hdrext` tests pass vacuously through their `is_supported()` early return — exactly the shape of a lookup returning `None` on a build that has the API.

Resolving by filename would not have fixed it either: the DLL is `gstrtp-1.0-0.dll` on MSVC and `libgstrtp-1.0-0.dll` on MinGW. The lookup now calls `K32EnumProcessModules` and asks each loaded module for the export, which depends on neither the anchor address nor a filename. The `gstreamer_rtp::ffi` anchor is gone.

I cannot run this path here. It type-checks against `x86_64-pc-windows-msvc`; `Build (Windows)` on this head is the actual evidence, and that job is now honest about failures.

**2. The message.** Split by version. Below 1.24 it stays at `info!` and says aggregation does not exist below 1.24. At 1.24 or above a failed lookup is a live problem, so it is `warn!` and says the workaround is inactive and a corrupt RTP stream can abort the process. Both report `gst::version()`.

**3. The masked exit code.** Fixed here rather than deferred, since without it a green `Build (Windows)` on this PR proves nothing. Each `cargo` invocation in the Windows job is its own step now. `Run Clippy` and `Build` were masked identically and are split the same way. The macOS job is unaffected — it runs under bash, which fails fast.

One thing worth flagging: the `gstreamer-rtp` dependency edge you raised is now load-bearing on Windows in a way it was not before. The lookup no longer imports anything from `gstreamer_rtp::ffi`, but the module walk still assumes the RTP library is loaded in-process, which is true because `gstreamer-rtp` links it. If that dependency ever goes away, the Windows lookup fails and `symbol_resolves_on_modern_gstreamer` catches it.

Locally on macOS / GStreamer 1.28: 585 lib tests plus all integration tests with `STROM_REQUIRE_GST_PLUGINS=1`, `clippy --all-targets --features efp,nvidia -D warnings`, `fmt --check`, all clean.

<sub>Written by Claude Code on behalf of @wagenet.</sub>

wagenet:

The Windows lookup fix works. On `97179f9`, `Build (Windows)` ran 580 lib tests with 0 failures, `symbol_resolves_on_modern_gstreamer` among them — the test that panicked on the previous head. GStreamer 1.26.10 resolves the setter through the module walk.

`Build (macOS)` passed. Linux x86_64, ARM64, Check, WASM and API Contract all passed.

The job still went red, and on something unrelated: `unmute_after_mid_mute_fader_change_fades_in_from_zero` in `volume_ramp_test.rs`. That test samples 8 ms into a 30 ms fade. Windows timers have roughly 15 ms granularity, so the sleep can overshoot the entire fade and read back the settled target — `volume=0.30000001192092896` is exactly the target, not a mid-fade value. It passed on `main`'s own Windows run at `95d7ccb`, so it is a flake rather than a break, and it was reachable this time only because the lib tests no longer abort the run first.

Fixed in `81afa86` rather than left for a re-run: the fade is 200 ms and the sample lands at 40 ms. The curve is dB-linear, so a fade-in from silence measures 0.003 there against a 0.25 threshold — verified locally, and the assertion holds anywhere in the first ~190 ms. An unmute that steps straight to the target with no fade still fails it, which is the P0 the test guards.

That makes three separable changes in this PR beyond the RTP fix: the Windows symbol lookup, the pwsh exit-code masking, and this flake. The last two are only here because without them a green `Build (Windows)` on this PR is not evidence. Happy to split either out if you would rather review them apart.

<sub>Written by Claude Code on behalf of @wagenet.</sub>

srperens (dismissed):
**Supersedes my 2026-09-08 review at `55e91efa` (Request changes). The verdict stands, but for an entirely different reason: all three requested changes are done and verified on Windows. What blocks now is the Windows job's time budget.**

**Verdict: Request changes** — the workaround is correct and now provably active on all four platforms; `Build (Windows)` is red because the job hit its 45-minute ceiling, and the diff's own new test binary is part of what pushed it there.

**Requested changes**

1. Raise the Windows ceiling — `.github/workflows/ci.yml:487` — `    timeout-minutes: 45`. This job measured 43m18s and 42m02s at the previous head (runs 34221734567, 34221734629) and 45m06s here, cut off in `Build MCP server` with everything else already green. 45 was never a margin; with sccache skipped it is a coin flip. 60 matches what the macOS comment at `.github/workflows/ci.yml:608` already argues for.

**Claims**

|---|---|---|
| The Windows lookup now resolves (my request 1) | `CONFIRMED` | `backend/src/gst/rtp_hdrext.rs:88` — `` `    let mut modules: Vec<*mut c_void> = vec![std::ptr::null_mut(); 256];` `` — enumerating every module replaces `GetModuleHandleEx`; Windows job 102736902832 logs `symbol_resolves_on_modern_gstreamer ... ok`, which only asserts at >= 1.24 |
| The no-op is loud and the message true (my request 2) | `CONFIRMED` | `backend/src/gst/rtp_hdrext.rs:142` — `` `            None => warn!(` `` — reached only when `(major, minor) >= (1, 24)`, and both arms print `gst::version()` |
| Splitting the pwsh steps makes a failed test fail the job (my request 3) | `CONFIRMED` | run 34432171254 at `97179f9a`: step `Run tests` conclusion `failure` on `unmute_after_mid_mute_fader_change_fades_in_from_zero ... FAILED`, with the four later steps `skipped`. Under the old single step that same failure read as success |
| Every new test executed rather than skipping | `CONFIRMED` | `Check (Linux)` on GStreamer 1.24.2: all four `gst::rtp_hdrext::tests`, both `mediaplayer::bridge::tests`, and `test_start_disables_hdrext_aggregation_on_depayloaders ... ok`. `is_supported()` is true there, so no early return is taken |
| The red Windows check is a defect in this diff | `CONTRADICTED` | job 102736902832: `Run Clippy`, `Run tests`, `Run tests on MCP server`, `Run tests on shared types` and `Build` all `success`; only `Build MCP server` was cut off at the 45m mark. No code of this diff is implicated |

**Diagnosis** — Unchanged and confirmed at this head. Coverage is now `ABSOLUTE` across the four `gst::Pipeline` hosts: `backend/src/gst/pipeline/lifecycle.rs:17` — `` `        rtp_hdrext::install(&self.pipeline);` `` plus `backend/src/blocks/builtin/mediaplayer/bridge.rs:129`, `backend/src/blocks/builtin/mediaplayer/bridge.rs:326` and `backend/src/blocks/builtin/whip.rs:1103`. `start()` is called once per manager and the manager is built fresh in `backend/src/state.rs:958`, so no pipeline collects a second `deep-element-added` handler.

**Radius** — `GLOBAL`, unchanged: pipeline lifecycle across four pipeline hosts. The closure captures nothing, and `pipeline_lifecycle_test.rs` passed 3/3 on Linux, so the CLAUDE.md reference-cycle rule is satisfied by construction rather than by promise.

**Tests & CI** — seven of eight checks green at `81afa866`. I read the Linux and Windows job logs. `openapi_test` and `pipeline_lifecycle_test` both ran and passed.

Confidence: HIGH

wagenet:

Done in `4d2e1d2`. `Build (Windows)` is at 60 minutes, with the measured numbers in a comment above it so the next person does not have to re-derive them.

Your read of the job matches mine: steps 1 through 13 all green, `Build MCP server` cancelled at the 45m mark, total window 44m59s. Nothing in the diff is implicated.

I left `Build (macOS)` at 45. It measured 29m28s on this head and 27 minutes when its comment was written, so it has real margin. Its comment did claim Windows "sits at 45 for the same reason", which is now wrong, so that sentence is updated.

**60 is a stopgap.** Before raising it I checked whether the job should just be faster. Recent `Build (Windows)` wall clock:

| Trigger | sccache | Duration |
|---|---|---|
| push to `main`, last 5 runs | on | 29m19s to 36m17s |
| this PR (fork) | off | cut off at 45m06s, projected ~47m |

45 was already thin on `main` before this branch, and sccache is worth only about 10 minutes here, so restoring the cache for fork pull requests would not have been the answer either.

Where the fork run's time goes:

| Phase | Time |
|---|---|
| Setup + GStreamer install | 1m43s |
| Clippy, both packages | 7m26s |
| Tests, all three packages | 15m12s |
| `cargo build --release --package strom` | 19m38s |
| Release MCP build + uploads | ~3m, cut off |

The release build is 45% of the job, and it is a smoke check rather than a shipping path: tagged Windows binaries come from `release.yml`, and this job's upload is a 7-day artifact. Two ways to cut it, and I prefer the first:

1. Move the release build into its own job, parallel with clippy and test. It already compiles into a separate target directory, so almost no CPU work is duplicated. It costs one more runner and a second GStreamer install, and wall clock drops to roughly 25 minutes.
2. Skip the release build on `pull_request`. Same 25 minutes on one runner, but release-only breakage then lands on `main` and is caught by the push run instead. That class is narrow, and the debug test binaries already cover linking.

Not `Swatinem/rust-cache` — the comment already in the job argues against it, and it is right for a job that runs this seldom.

I have not done either here. Changing what `ci:windows` builds is a policy change for every future pull request, and it does not belong in an RTP depayloader fix. Happy to open it separately if you want it.

<sub>Written by Claude Code on behalf of @wagenet.</sub>

srperens (commented):
**Supersedes my 2026-09-11 review at `81afa866` (Request changes), which I am dismissing alongside this one. The verdict does not stand: the single requested change is done, and no defect blocks merging now.**

**Verdict: Comment** — the code delta since that review is empty (`git diff 81afa866..4d2e1d2e` touches `.github/workflows/ci.yml` and nothing else), so every code row from it stands unchanged. What is new is that Windows is green and the diff's own tests are now proven to have executed on all three platforms. `GLOBAL` radius is the only thing holding this short of the stronger token — not a finding.

**Requested changes** — none.

**Claims**

|---|---|---|
| 45 minutes was the blocker and 60 clears it | `CONFIRMED` | `.github/workflows/ci.yml:491` — `    timeout-minutes: 60`. Job 103414970110 ran 20:40:17Z to 21:26:07Z — 45m50s, conclusion `success`. The old ceiling would have cut this very run |
| Every new test executed on Windows and macOS, not only Linux | `CONFIRMED` | Job logs 103414970110, 103414970035 and 103413833906, all at head `4d2e1d2e`: `symbol_resolves_on_modern_gstreamer ... ok`, both `install_disables_aggregation_* ... ok`, both `*_pipeline_disables_hdrext_aggregation ... ok`, and `test_start_disables_hdrext_aggregation_on_depayloaders ... ok`. Windows is the platform whose lookup could not be tested locally, so that first line is the one that mattered |
| No fifth pipeline can host a depayloader that runs buffers | `CONFIRMED` | `backend/src/api/gst_launch.rs:89` — `    let pipeline = match gst::parse::launch(&cleaned_pipeline) {` is the one place a user-supplied string builds a pipeline, and that file contains no `set_state` call at all, so nothing there ever prerolls; `backend/src/gpu.rs:117` is a fixed probe string with no depayloader; `backend/src/whip_session_manager.rs:547` is inside `#[cfg(test)]` |

**Diagnosis** — unchanged; see the superseded review for the mechanism and the four `install()` call sites. The one new result is the coverage row above. I asked this run whether a user could route around `install()` through the `gst-launch` parse endpoint, since that pipeline is built from an arbitrary string and never passes through `PipelineManager::start()`. It cannot: the endpoint parses for element introspection and never leaves NULL, so `gst_rtp_base_depayload_handle_buffer` is never reached. Coverage stays `ABSOLUTE` over the pipelines that actually run media.

**Radius** — `GLOBAL`, unchanged: four `install()` call sites across three pipeline hosts, plus a CI ceiling that every future Windows run in this repo inherits.

**Tests & CI** — all eight checks pass at `4d2e1d2e` (run 34645083079), `Build (Windows)` and `Build (macOS)` among them; they run here because the PR now carries the `ci:windows` and `ci:macos` labels.

**What a merger is accepting**, both disclosed in the body and neither newly found: the WHIP Input call site has no test, because `build_whipserversrc` binds a port; and the pre-1.24 no-op branch is exercised nowhere, since no CI image ships 1.22.

Confidence: HIGH

Issue #661, opened 2026-07-06
audiorouter working live


markusnygard:
## Problem

the current audiorouter does not support live changes

## Proposed Solution

a matrix router like https://gstreamer.freedesktop.org/documentation/audiomixmatrix/index.html?gi-language=c

## Alternatives Considered

Any alternative solutions or features you've considered.

## Additional Context

The features Im missing is: auto.recognision of amount of audiotracks, a way to live connect inpoints to outpoints,  a default connection (inpoint-audio goes to fader), in- and outpoints per devices and you can minimize or maximize the in- and out view. the normal view would be a line of dots going in a 45degree angle.

srperens:
Thanks for the feature request, @markusnygard.

**Understanding the ask:** The existing audio router block does not support live reconfiguration of its routing matrix — changes require stopping and restarting the flow. You'd like a matrix-style router (similar to `audiomixmatrix`) that:
- Auto-detects the number of audio channels/tracks on its inputs.
- Supports connecting input points to output points while the flow is running.
- Has a sensible default routing (input channel → its fader or direct output).
- Exposes a UI where rows = inputs, columns = outputs, with dots indicating active connections (a classic crosspoint matrix at 45°).

**Why this matters:** Live audio routing is fundamental to any production workflow — you want to reroute between sources without dropping the feed.

**Possible implementation approach:**

The GStreamer `audiomixmatrix` element (`matrix-mode=transform-passthrough` or `custom-matrix`) supports runtime matrix updates via the `matrix` property. A Strom block wrapping it could:
1. Auto-size based on negotiated caps (reading channel count from incoming audio caps).
2. Expose the routing matrix as a live property so Strom's properties panel or a future dedicated UI widget can update it without restarting.
3. Support per-channel gain (as a double multiplier), which `audiomixmatrix` already implements natively.

The UI for a full crosspoint matrix is non-trivial — it would likely need a custom egui widget or a dedicated panel rather than the generic properties sidebar.

**Questions for scoping:**
- Is the primary use case rerouting between physical audio devices (e.g., NDI sources → outputs), or is it more about mixing/summing tracks within a flow?
- What's the expected channel count range — stereo pairs (2–8 ch), or full broadcast matrices (16–64+ ch)?
- Would per-point gain (mixing) be needed, or pure routing (0/1)?

Answers will help determine whether `audiomixmatrix` is the right foundation or if a custom routing element would be needed.

**Labels:** `enhancement`

srperens:
[...]

Issue #673, opened 2026-08-11
Vision mixer outputs black unless every declared num_inputs pad is connected


wagenet:
### Summary

`builtin.vision_mixer` outputs black on PGM unless **every** declared `num_inputs` pad is connected, even when the input selected for program is connected and producing frames.

### Reproduce

Two flows, identical except for whether `video_in_1` is wired. Both use `num_inputs: 2` and the same source on `video_in_0` (BGRA, see #<BGRA issue> for why the capsfilter is needed).

**A — only `video_in_0` connected → PGM is black**

```
videotestsrc(pattern=blue,is-live) ! capsfilter(video/x-raw,format=BGRA,...) ! mixer:video_in_0
mixer:pgm_out ! multifilesink
```
First pixel: `RGBA=(0,0,0,255)`

**B — `video_in_1` also connected → PGM shows input 0 correctly**

```
(as above, plus)
videotestsrc(pattern=green,is-live) ! capsfilter(...) ! mixer:video_in_1
```
First pixel: `RGBA=(0,0,255,255)` — the blue from input 0

The only change is connecting an input that is *not* on program.

### Why it matters

A mixer configured with more inputs than are currently wired is a normal intermediate state — while building a production, or when a source is temporarily removed. Today that silently produces black program output with no error or warning, which is hard to diagnose: the flow reports `running: true` / `gst_state: Playing`.

Presumably the compositor is waiting on the unconnected sink pads. If all declared pads genuinely must be fed, feeding unconnected ones from a black/silent source internally would preserve the behaviour without the surprise; otherwise refusing to start (or warning) would at least make it visible.

Environment: v0.6.6, macOS arm64, `compositor_preference: "cpu"`.

srperens:
Thanks for the detailed reproduction, @wagenet.

**Restating the problem:** When `builtin.vision_mixer` is configured with `num_inputs: N` but fewer than N pads are connected, the program output is black even for pads that _are_ connected and actively producing frames. The flow reports `running: true` / `gst_state: Playing` with no error, making the issue silent and hard to diagnose.

**Root-cause area:** GStreamer's `compositor` element holds all its sink pads to a common base time and will not produce output until all pads have received at least one buffer (or are marked as inactive/EOS). When a pad is requested but never fed, it stalls indefinitely, blocking the mix.

The vision mixer block likely requests all N sink pads up front at build time (matching `num_inputs`), so any unfed pad holds the compositor hostage.

[...]

PR #728, opened 2026-08-31
chore(agent): move the review-bot protocol into the repo, with two checks in code

srperens:
The two scheduled agents that review Strom's PRs and triage its issues carried their entire
protocol inside their task definitions — one ~18 000-character prompt each, where it could
not be diffed, reviewed, or corrected by anyone but the person holding the schedule. This
moves the deployment-independent half into `scripts/agent/`.

Nothing here names a credential, a runner, a schedule, or a notification channel. That half
stays in the task definitions, which is what makes this safe to read in public.

## Why, concretely

Reading two weeks of live output, the failures were not reasoning failures — the reviews are
good. They were failures of format, vocabulary, and unverifiable self-checks:

- **Six run-summary shapes in six runs.** The rules *described* a summary rather than showing
  one, so every run invented its own heading and table schema. Fixed by a worked example per
  output type; the examples are the spec.
- **A column invented its own value.** One no-op run recorded `#721 | LOCAL | unchanged` when
  the standing review classified no radius at all. That PR carries three different radii
  across three summaries. Fixed by a controlled vocabulary that the tooling actually reads,
  plus the rule that a field you did not determine is `null`.
- **The log became unreadable by the agent itself.** `gh api .../comments` without
  `--paginate` returns the thirty *oldest*; with 77 comments a run read page 1 and reported a
  gap in the log that does not exist. State now lives in the log issue's body.

## `board.sh` reports; it does not decide

This started as a gate script that mechanised all five eligibility rules. **Live data showed
that was wrong.** On #661 the maintainer answered by rejecting both options the triage had
offered, naming a third, and explaining that the triage's `SHARED`/`GLOBAL` radius applied
only to the two it had proposed. The old prose-driven task read that correctly and shipped a
draft PR that reached class A the same morning with a runtime test that genuinely executed.
The strict gates would have refused it: the reply names no option token, and the radius
override was prose, not `--accept-radius LOCAL`.

No token vocabulary expresses "neither of your options, here is a third, and your radius call
was about something else" — and that is the most valuable answer shape this project gets. So
the script reports, and three findings stay binding because they are unambiguous and being
wrong is expensive: no reply at all, a reply only from someone who may not decide, an issue an
open PR already claims. Everything else is a note the implementation stage must answer in its
PR body — including its own re-assessment of radius and excluded areas for the design actually
chosen, since a marker describes the options the *triage* proposed.

Removing the mechanical judgement removed most of the defect surface with it. The override
flags are gone entirely; one of their two defects had made the feature dead code.

## `verify-citations.sh`

The protocol rests on `path:line` citations quoted verbatim, and "re-read the line before
citing it" is the least reliable instruction available. Run against the two most recent live
reviews it found:

- a **real off-by-one**: `whip.rs:1040` cited as `rtp_hdrext::install(&session_pipeline);`
  when line 1040 is a comment and the call is at 1041 — in a review marked `Confidence: HIGH`;
- four citations to bare filenames (`mod.rs:257`, `flows.rs:577`) matching 2–24 tracked files
  each, which cannot be spot-checked at all.

Single-line citations are verbatim claims and their quotes are checked; ranges are
bounds-checked. `--max-chars N` makes the length ceilings mechanical rather than instructed.

## `test-agent-scripts.sh` — 40 assertions, no network

**Every severe defect in these two scripts was invisible to careful reading.** Two separate
code reviews read them and missed dead override parsing, a radius check that passed a marker
with no radius at all, and adjacent empty TSV fields shifting a maintainer's decision into the
wrong variable. All three fell out of *running* them. A fail-open check is silent by
construction, so it needs a fixture rather than a careful reader.

Each assertion was verified to fail when its fix is reverted — checked by reverting four of
them, including both earlier fail-open holes.

## Other behaviours, each from an observed failure

| Change | The failure it prevents |
|---|---|
| A comment is the agent's by its **marker at the start of a line**, never its author | Both tasks post as the maintainer's own account, so an author check discards their decisions while reporting an empty queue. And GitHub's web "Quote reply" embeds the marker, which would classify a human's answer as the agent's *and* make every later reply look older than the triage. |
| `claiming_pr` matches per field | A PR whose *body* quoted a branch name was reported as implementing an unrelated issue. This branch's own PR did exactly that. |
| Empty TSV fields travel masked | Tab is IFS whitespace, so bash collapses runs of it; two adjacent empty fields shifted an `/agent-fix` line into the marker variable and reported a real decision as "no token". |
| `work=bug\|extension\|feature` selects branch/commit/title words | #700 shipped an `enhancement` as `agent/fix-691-…`, `fix(efpsrt): …`, commit 1 claiming to "reproduce" a defect that never existed. |
| `excluded=` means "would **break**", not "touches" | CLAUDE.md *requires* new API-visible types in `strom-types`. Scoring additive surface as excluded would rule out the one placement the repo mandates. |
| `class=C` (blocked on a dispatch) is exempt from stale closure | macOS/Windows build only on `workflow_dispatch` and this stage may not fire one, so the 14-day rule would close correct work for being blocked on someone else. |

## On verbosity

Two thirds of everything these tasks have written is run summaries; the worst re-listed
sixteen unchanged issues to say nothing had changed. The durable third — a triage's reasoning,
a PR's design record — is worth its length, and the repo keeps no internals docs, so that
trail *is* the documentation. So the rule is not "be brief" but **explain fully once, on the
artifact it belongs to, and cite it thereafter**: ceilings are tight where output repeats every
run (summaries, 1200 chars) and generous where it explains something once (triage 2500, PR
bodies 3500, reviews 4000).

## What I ran

- `scripts/agent/test-agent-scripts.sh` — 40 passed, and each fails on reverting its fix.
- `board.sh` against the full open board, from which two of the defects above surfaced.
- `verify-citations.sh` against both live reviews, a deliberately broken body, a body with no
  trailing newline, and from three working directories.
- Pre-commit (fmt + clippy + secret scan). No Rust changed, so no Rust test covers this;
  `test-agent-scripts.sh` is not wired into CI, which is the remaining gap.

## Migration note

Existing triage markers predate the structured fields, so `board.sh` reports that as a note
rather than blocking. `TRIAGE.md` describes a marker backfill — one comment, no
re-verification, no item budget — which also carries the `/agent-fix` invitation that standing
triages never had.

## Not addressed, deliberately

GitHub Discussions. Both tasks read issues only; Discussions sit behind GraphQL and have no
gate to protect. If ever added, triage-only.


srperens (commented):
Five changes before this goes live. The first two are the blocking ones, and the order matters: nothing here takes effect until the two OSC task definitions are re-pointed at these files, and the first run after that switch is when the backfill fires.

The `work=` lane, the quoted-line anchoring, the bare-`/agent-fix` routing, the override flags, the two re-triage triggers and the first-decided-first-built ordering all look right. Not repeating them below.

### 1. The marker backfill must edit the standing comment, not post a new one

`TRIAGE.md`, "Marker backfill", says to *"post a short comment carrying only the marker"*. All 16 open issues lack the structured fields, so that is 16 new comments over three runs, each one carrying nothing a human can use.

That is the exact shape of the last notification incident. On 2026-08-21 the v2 to v3 pass posted a triage and then a "supersedes my earlier comment" on the same issues; @wagenet received 8 notifications that day (4 issue comments on #673/#674, 4 reviews). Most said, in effect, that the previous comment still stood.

The agent owns those comments, so `PATCH /repos/{owner}/{repo}/issues/comments/{id}` appending the marker line costs zero notifications and leaves the audit trail intact. It also removes an invariant: the code comment in `gates.sh` explaining that *"a marker backfill sorts last by construction"* becomes unnecessary, because there is no new comment to sort.

The invitation line (`To implement Option A, reply: /agent-fix A`) should be appended in the same edit — it is the only thing that tells a maintainer the token exists, and it belongs on the comment that carries the proposal.

### 2. "I looked and nothing changed" belongs in the run summary, never in the thread

`TRIAGE.md`, "When a standing triage must be re-read", case 2 ends with *"If it did not, say so and restate the token to use."* Said where matters. Posting it on the issue means one bot comment for every human comment: #661 took three human comments in a row on 2026-08-21, and #690 is a back-and-forth. Every one of those would now draw a reply.

The rule this repo already has for PRs is the right one — `REVIEW.md`: *"A review is a verdict on a diff, not a turn in a conversation."* Extend it: the issue thread gets a comment only when the verdict moved. Everything else goes in the summary.

While in there: the protocol budgets *items* (four per run) but never notifications. Today's 08:00 run posted four reviews in one morning, all to the same person. One line capping new comments per run, and preferring an edit to a post, would make that a deliberate number instead of an emergent one.

### 3. The stated rationale for splitting the protocol is not supported

`README.md:65` — *"Instruction-following degrades with the number of simultaneous constraints, and the earlier a rule sits the more reliably it is obeyed; that is why each file leads with its hardest rules and why the protocol is split at all."*

The factorial study on this (arXiv 2605.10039) varied file size, instruction position, file architecture and contradictions between adjacent files, and found no measurable effect from any of them after correcting for multiple comparisons. What it did find is within-session decay: roughly 5.6% lower odds of compliance per additional function generated as a session progresses.

Keep the split — it is diffable, reviewable, and a PR against it is a PR against how the bots behave, which is the honest reason. But move the compliance argument to where the evidence points: **re-read the item-specific file before each item**, since the review task processes four items in one session and the fourth is the least compliant. The existing instinct to post the summary two thirds through the budget is the same insight; generalise it.

### 4. Scaling the work down is the maintainer's call — including the scope, not just the verification

`FIX.md:192` says this about what was not verified. It needs to hold for what was cut.

Live example from today. On #661 I wrote a scoping comment that put the UI out of scope. PR #729 implemented exactly that, correctly, and was worthless: the issue is a UX request, and a backend block with a JSON matrix property is not a live audio router to the person who filed it. The stage that narrowed the scope was me, but nothing in the protocol would have stopped the agent doing the same on its own — narrowing until the work fits the two-commit structure looks like progress and is not.

Suggested rule: if the implementable slice is smaller than what the issue asks for, say so in the PR body and name what was left out, in the same words the "Not verified" section uses. A PR that silently delivers the testable fraction is the failure mode this stage is most prone to, because the fraction is testable and the rest is not.

### 5. Ask whether the issue is "the same as" something that already exists

The most testable requirement form there is, and both stages walked past it. The maintainer's decision on #661 said the new block *"could have the same config/settings and user experience as the previous block"*. Read as prose, that produced #729, which invented `in_channels`/`out_channels` and a single pad pair and could not be rendered by the existing editor at all. Read as a specification, it is an assert: the exposed-property name set and the pad shape of the new block equal those of the old one, and it goes red the moment they drift.

`TRIAGE.md` should ask the question directly in the design proposal: *is there an existing block, module or behaviour this must match?* If yes, parity is the acceptance criterion and the guard writes itself. That is also the cheapest way to make `work=feature` verifiable, which is otherwise the weakest row in the new `work=` table.

PR #729, opened 2026-08-31
feat(liveaudiorouter): add a live matrix audio router

### LINKED ISSUE #661: audiorouter working live

srperens:
Class A — the test below fails without this fix.

Fixes #661

## Problem

`builtin.audiorouter` declares its routing matrix non-live. `backend/src/blocks/builtin/audiorouter.rs:847`:

```rust
        live: false,
```

That flag is on the `routing_matrix` exposed property, whose mapping targets the pseudo-element `_block` (`audiorouter.rs:842-846`), because the matrix is not a GStreamer property at all — it is consumed at `build()` time to decide how the block's internal `deinterleave`/`interleave` pads are linked. Once the flow is running the wiring is fixed, so changing a route needs a flow restart.

Making the existing property live is not enough on its own: the generic live-update path writes strings straight through. `backend/src/gst/pipeline/properties.rs:73-75`:

```rust
        match prop_value {
            PropertyValue::String(v) => {
                element.set_property_from_str(prop_name, v);
```

`set_property_from_str` panics on a property it cannot find or a value it cannot deserialize, so handing it a JSON routing matrix would take the backend down rather than return an error.

## Change

A new sibling block `builtin.liveaudiorouter`, built on `audiomixmatrix` (gst-plugins-bad) instead of `deinterleave`/`interleave`. Chain: `audioconvert` -> `audiomixmatrix` -> `audioconvert`. `audiomixmatrix` holds the routing as a `matrix` property (a value array of `gdouble` rows, indexed `[output][input]`) that can be rewritten while the element is PLAYING, so no relinking is involved.

Three parts:

- `backend/src/blocks/builtin/liveaudiorouter.rs` (new) — the block, plus `routing_matrix_to_gst_array()` which converts the existing `{"i0c0": ["o0c1"]}` JSON syntax into the matrix literal, and `try_apply_live_matrix()` which performs the live write.
- `backend/src/gst/pipeline/properties.rs` — an 11-line dispatch hook that gives `try_apply_live_matrix` a chance to claim the property before the generic string path above can reach `set_property_from_str`.
- `backend/src/blocks/builtin/mod.rs` — registry entries (module, `get_all_builtin_blocks`, `get_builder`).

The interceptor returns `true` even when it rejects the JSON as unparseable. That is deliberate: claiming the property and logging a warning is what keeps a typo in the properties panel from panicking the backend. It follows the existing precedent of `time_offset::try_apply_live_offset` immediately above it in the same function.

The triage discussion also raised changing `builtin.audiorouter` in place. That was rejected there because `audiomixmatrix` cannot reproduce the existing block's multi-stream (up to 8 in / 8 out) pad topology, so an in-place swap would be a breaking change to every existing flow using it. This block is single-stream, multi-channel, and additive.

## Evidence

Two concluded CI runs, red on the test-only commit and green on the commit that adds the block:

| Commit | Run | Conclusion |
|---|---|---|
| `ceab483` test only | [33381858279](https://github.com/Eyevinn/strom/actions/runs/33381858279) (on `1157680`, see caveat) | **failure** |
| `6dbbbba` test + block | [33382957051](https://github.com/Eyevinn/strom/actions/runs/33382957051) | **success** |

The red run fails in `Run Clippy on backend`:

```
error[E0432]: unresolved import `strom::blocks::builtin::liveaudiorouter`
error: could not compile `strom` (test "liveaudiorouter_live_matrix_test") due to 1 previous error
```

The green run executes all five tests, including the runtime one:

```
test an_empty_matrix_routes_every_input_channel_straight_through ... ok
test a_routing_entry_lands_in_the_row_of_its_destination_channel ... ok
test one_input_channel_can_fan_out_to_several_output_channels ... ok
test unparseable_json_is_rejected_instead_of_reaching_gstreamer ... ok
test a_routing_matrix_change_moves_audio_between_output_channels_while_playing ... ok
```

The job runs with `STROM_REQUIRE_GST_PLUGINS=1`, so a missing `audiomixmatrix` or `level` would have failed rather than skipped green. Both elements are already in the Linux job's package list (`gstreamer1.0-plugins-bad`, `gstreamer1.0-plugins-good`), so no workflow change was needed.

`backend/tests/liveaudiorouter_live_matrix_test.rs` feeds a 1 kHz tone into a two-channel input, routes it to output channel 0, then calls `liveaudiorouter::try_apply_live_matrix` mid-stream to send it to output channel 1, and reads a `level` element's per-channel peaks on both sides of the change. It asserts on measured peaks, not on the property value — reading `matrix` back would pass even if the element ignored the new value for the audio it is actually producing. This is what confirms the load-bearing assumption that `audiomixmatrix` applies a new `matrix` to audio it is already producing rather than latching it at caps negotiation; I could not verify that from documentation alone.

To run it locally:

```
STROM_REQUIRE_GST_PLUGINS=1 cargo test --features efp,nvidia --test liveaudiorouter_live_matrix_test
```

## Not verified

- **I compiled and ran nothing myself.** No Rust toolchain existed in the environment I ran in (`cargo --version` and `pkg-config --modversion gstreamer-1.0` both report command not found). Every execution result above is CI's, not mine.
- **The red run is on a superseded SHA.** Run 33381858279 ran against commit 1 as `1157680`; after a formatting fix and a harness fix the branch was rewritten, so commit 1 is now `ceab483`. The only difference in the test file between those two SHAs is a three-line `gst::init()` call added to the plugin guard, which cannot affect the `unresolved import` compile error the red run demonstrates. I chose not to bounce the branch head back to commit 1 purely to regenerate that run, since that would leave this PR pointing at a non-building commit while CI ran.
- **The red is a compile error, not a failing assertion.** Commit 1 cannot build because the test imports the module commit 2 adds. So the red/green pair does not by itself prove the runtime behaviour would regress if only the *liveness* were reverted. Reverting liveness specifically means setting `live: false` on the new block's `routing_matrix` property and removing the hook in `properties.rs`; that makes `wait_for_only_channel(1, ...)` time out rather than fail to build. I did not produce that third state, because doing so would have meant a third commit.
- **Scope gap, not dropped silently:** the triage note suggested deriving channel counts from the negotiated input caps. This does not implement that. In `audiomixmatrix`'s manual mode the matrix dimensions *are* `in-channels`/`out-channels`, so they must be known at `build()` time; `in_channels`/`out_channels` are therefore plain non-live properties. Auto-derivation would need the element's automatic mode, which is a different design and should be a separate decision.
- No frontend work and no GUI run. The block appears via the generic block list and its `routing_matrix` uses the existing `Multiline` property editor, so it should render, but I did not confirm it visually.
- Channel counts are clamped to 1..=64 silently rather than rejected, matching `MAX_CHANNELS` in `audiorouter.rs`.
- Only the Linux jobs exercised the tests; the macOS and Windows build jobs were skipped by the workflow's own conditions.

## Blast radius

- `backend/src/gst/pipeline/properties.rs` is shared by every live property update in the product, so the hook is the only part of this diff that can affect existing flows. It is guarded twice: `try_apply_live_matrix` returns `false` immediately unless the property name is exactly `routing_matrix` *and* the element id ends with `:live_router_matrix`. No existing block produces an element with that suffix, so for every current caller the function returns `false` and control falls through to the unchanged translation chain. The green run's 546 passing unit tests plus the existing integration tests exercise that fall-through path.
- Call sites reach that function through `AppState::update_element_property` (`backend/src/state.rs:1691`), which is the single funnel used by the REST handler (`backend/src/api/flows.rs:999`), the MCP tool (`backend/src/mcp/handler.rs:549`) and the internal callers at `state.rs:1803`, `1978` and `1993`. Reading `state.rs:1672-1699`: it resolves the pipeline, delegates, then broadcasts `StromEvent::PropertyChanged`. The hook returns `Ok(())` on the same path the existing `time_offset` hook does, so the event still fires and no contract changes.
- `backend/src/blocks/builtin/mod.rs` gains three additive entries. No block-list snapshot test exists, and no API or WebSocket type changed, so `openapi.json` is untouched — the `API Contract Check` job passes.
- The existing `builtin.audiorouter` is not modified. Existing flows keep their current behaviour, including the restart requirement.

### DIFF
--- a/backend/src/blocks/builtin/liveaudiorouter.rs
+++ b/backend/src/blocks/builtin/liveaudiorouter.rs
@@ -1,0 +1,395 @@
+//! Live audio channel router built on GStreamer's `audiomixmatrix`.
+//!
+//! Sibling of `builtin.audiorouter`, which wires a static
+//! `deinterleave`/`tee`/`interleave` graph from a matrix snapshot taken at
+//! `build()` time and therefore needs a flow restart to reroute a channel. This
+//! block keeps a single `audiomixmatrix` whose `matrix` property is rewritten in
+//! place, so `routing_matrix` is exposed as a live property.
+//!
+//! Chain: audioconvert → audiomixmatrix → audioconvert
+//!
+//! The surrounding `audioconvert` elements absorb the channel-count and format
+//! negotiation on both sides: `audiomixmatrix` accepts exactly `in-channels`
+//! channels on its sink and emits exactly `out-channels` on its src, which
+//! upstream and downstream have no reason to already match.
+//!
+//! `routing_matrix` uses the same JSON shape as `builtin.audiorouter`
+//! (`{"i0c0": ["o0c1"]}`). `audiomixmatrix` is a single-stream element, so the
+//! stream index in those keys is always 0 and only the channel index varies.
+
+use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult, BlockBuilder};
+use gstreamer as gst;
+use gstreamer::prelude::*;
+use std::collections::HashMap;
+use strom_types::{block::*, element::ElementPadRef, PropertyValue, *};
+use tracing::{info, warn};
+
+/// Maximum channels on either side of the matrix. Mirrors `audiorouter`'s limit.
+const MAX_CHANNELS: usize = 64;
+
+/// Element-id tail of the `audiomixmatrix`. The live-update interceptor matches
+/// on it, so it must stay in step with the element naming in `build()` and with
+/// the `PropertyMapping` in the block definition.
+const MATRIX_ELEMENT_ID_TAIL: &str = ":live_router_matrix";
+
+/// GStreamer type hint written into each serialized matrix cell.
+///
+/// `audiomixmatrix` declares `matrix` as an array of arrays of `gdouble`, and
+/// the value has to deserialize under that spec. Kept as a named constant
+/// because it is the one token in this file that a GStreamer version could
+/// plausibly disagree about.
+const MATRIX_CELL_TYPE: &str = "gdouble";
+
+/// Live Audio Router block builder.
+pub struct LiveAudioRouterBuilder;
+
+/// Parse a channel count property, clamped to a matrix `audiomixmatrix` accepts.
+fn parse_channel_count(
+    properties: &HashMap<String, PropertyValue>,
+    key: &str,
+    default: usize,
+) -> usize {
+    properties
+        .get(key)
+        .and_then(|v| match v {
+            PropertyValue::UInt(u) => Some(*u as usize),
+            PropertyValue::Int(i) if *i > 0 => Some(*i as usize),
+            _ => None,
+        })
+        .unwrap_or(default)
+        .clamp(1, MAX_CHANNELS)
+}
+
+/// Parse a routing key like `i0c1` or `o2c3` into (stream index, channel index).
+///
+/// Duplicated from `audiorouter` rather than shared: that block is deliberately
+/// left untouched by this change, and the key format is part of this block's own
+/// property contract.
+fn parse_routing_key(key: &str, prefix: char) -> Option<(usize, usize)> {
+    if !key.starts_with(prefix) {
+        return None;
+    }
+
+    let rest = &key[1..];
+    let c_pos = rest.find('c')?;
+
+    let stream_idx: usize = rest[..c_pos].parse().ok()?;
+    let channel_idx: usize = rest[c_pos + 1..].parse().ok()?;
+
+    Some((stream_idx, channel_idx))
+}
+
+/// Serialize an `[output][input]` coefficient grid as a GStreamer value array.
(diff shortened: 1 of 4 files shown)


srperens:
Closing this. It implements the backend block only, which is what my scoping comment on #661 asked for, and that scope was wrong: the issue is mostly a UX request (live crosspoint connection, default routing, minimise/maximise, the 45-degree dot view) and none of that is here. Not a defect in the work — the instruction was. #661 stays open and I have retracted the /agent-fix on it so this is not rebuilt tomorrow.

PR #735, opened 2026-08-31
fix(macos): keep the headless server out of App Nap

wagenet:
## Problem

Once a headless Strom process is ~32 s old, macOS moves the whole task into the
background QoS band and leaves it there: every thread drops to scheduling priority
4, which on Apple Silicon means efficiency cores at a throttled clock. CPU-bound
work slows down accordingly — a 1080p x264 encode takes 7.4x longer, measured
below.

This tracks process age, not work done. A backend that sat idle and then ran its
first cycle is already slow if it is old enough, and a restart resets it for
another 32 s.

## Cause

Headless Strom registers with LaunchServices as a `Foreground` app and never opens
a window. `gst-launch-1.0` has the same runtime shape — `gst_macos_main`, a
main-thread CFRunLoop, an `NSEventThread`, no window — but registers as
`UIElement` and is never demoted. Read it with `lsappinfo info -pid <pid>`.

| process | type | priority while idle |
| --- | --- | --- |
| `gst-launch-1.0 udpsrc port=N ! fakesink` | `UIElement` | 46, held 200 s |
| `gst-launch-1.0 videotestsrc ! fakesink` | `UIElement` | 46, held 200 s |
| headless Strom, unpatched | `Foreground` | 4, from ~32 s |

Workload does not discriminate: the `udpsrc` pipeline waits on packets that never
arrive at zero CPU and keeps priority 46. Activation policy does. Forcing the
unpatched server to the accessory policy with `setActivationPolicy:`, while taking
no activity assertion, held it at 46 for 162 s against a `Foreground` control
demoted on schedule.

What registers Strom as `Foreground` is unknown. Ruled out by measurement:
`gst_macos_main` and `gst_macos_main_simple` (C probes of both register
`UIElement`), load-time AppKit linkage, creating `NSApplication` before
`gst_macos_main`, winit/eframe (a `--features no-gui` build still demotes), and
CEF (`cefsrc` is not installed on the measuring machine). The assertion opts out
whichever it turns out to be.

## Fix

Hold a user-initiated activity for the life of the process.
`UserInitiatedAllowingIdleSystemSleep` rather than `UserInitiated`, because the
plain variant also blocks system idle sleep; `LatencyCritical` also opts out of
timer coalescing. `objc2-foundation` already ships via eframe, so no new crate.
Only `run_headless_entry` takes the assertion, so GUI mode is untouched.

## Verification

12 paired reps: counterbalanced randomised order, a fresh server per rep, a 50 s
settle so both arms age equally, and one binary with an environment variable
gating only the assertion call. Each rep records its own `ps -M` priority and host
load. Workload is 600 frames of 1080p `videotestsrc ! capsfilter ! x264enc !
fakesink`, timed from `Paused -> Playing` to end-of-stream in the server log.

| | patched | unpatched |
| --- | --- | --- |
| median | 3.62 s | 26.68 s |
| range | 3.52-3.66 s | 23.0-29.7 s |
| priority | 50 / 56 | 4 |

Geometric mean 7.41x, 95% CI 7.16-7.67, 12/12 pairs, t = +114.4, p = 2.9e-18.
Order effect is 7.36x patched-first against 7.53x unpatched-first (5 and 7 pairs).

A/A control, both arms patched, same schedule and analysis: 0.994x, CI
0.987-1.002, p = 0.15. Per-pair spread is 1.25% there against 6.06% in the A/B,
the difference being variability in the demoted arm.

Demotion threshold, six servers sampled once a second: 31.7-32.5 s, median 32.1 s.

M-series laptop, AC power, host load 1.9-5.8 across the reps. Battery is untested.

## Tests

- `headless_entry_takes_the_activity_assertion` catches the call going missing
  from `run_headless_entry`, by watching for the line it logs. 0.5 s.
- `server_is_not_demoted_to_background_qos` checks that macOS still honours the
  assertion, which needs the full demotion window, so it is `#[ignore]`d:
  `STROM_TEST_APP_NAP=1 cargo test --test macos_app_nap_test -- --ignored`. Not
  wired into CI, where the runner is a VM that may not App-Nap at all.

A unit test in `macos_app_nap.rs` catches a wrong selector on every macOS
`cargo test`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6618,6 +6618,7 @@ dependencies = [
  "network-interface",
  "notify",
  "nvml-wrapper",
+ "objc2-foundation 0.3.2",
  "parking_lot",
  "reqwest 0.13.4",
  "rust-embed",
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -140,6 +140,15 @@ eframe = { workspace = true }
 chrono = "0.4"
 uuid = { version = "1.23", features = ["v4"] }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+# App Nap opt-out: see backend/src/macos_app_nap.rs. Already in the tree via
+# eframe, so this adds no new crate.
+objc2-foundation = { version = "0.3", default-features = false, features = [
+    "std",
+    "NSProcessInfo",
+    "NSString",
+] }
+
 [target.'cfg(windows)'.build-dependencies]
 winresource = "0.1"
 
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -31,6 +31,7 @@ pub mod gst;
 pub mod gui;
 pub mod json_rejection;
 pub mod layout;
+pub mod macos_app_nap;
 pub mod mcp;
 pub mod network;
 pub mod openapi;
--- a/backend/src/macos_app_nap.rs
+++ b/backend/src/macos_app_nap.rs
@@ -1,0 +1,53 @@
+//! Keep macOS from putting the server into the background QoS band.
+//!
+//! Headless Strom registers with LaunchServices as a `Foreground` app but never
+//! opens a window, and macOS App-Naps that shape ~32 s after launch: every thread
+//! drops to scheduling priority 4, which on Apple Silicon means efficiency cores
+//! at a throttled clock, and pipelines run ~7x slower for the rest of the
+//! process's life. `gst-launch-1.0` is exempt because it registers as
+//! `UIElement`; what makes Strom `Foreground` is not known.
+//!
+//! `beginActivityWithOptions:reason:` is the documented opt-out, and works
+//! regardless of activation policy.
+//!
+//! Only headless mode needs this. In GUI mode the window keeps the app out of App
+//! Nap.
+
+/// Take a user-initiated activity assertion that lasts as long as the process.
+///
+/// Safe to call more than once; each call takes its own assertion.
+#[cfg(target_os = "macos")]
+pub fn hold_activity_for_process_lifetime() {
+    use objc2_foundation::{NSActivityOptions, NSProcessInfo, NSString};
+
+    // `UserInitiatedAllowingIdleSystemSleep` rather than `UserInitiated`: both opt
+    // out of App Nap, but the plain variant also blocks system idle sleep, and a
+    // media server has no business keeping the machine awake. `LatencyCritical`
+    // also opts out of timer coalescing.
+    let options = NSActivityOptions::UserInitiatedAllowingIdleSystemSleep
+        | NSActivityOptions::LatencyCritical;
+    let reason = NSString::from_str("Strom is running media pipelines");
+    let activity = NSProcessInfo::processInfo().beginActivityWithOptions_reason(options, &reason);
+
+    // The assertion lasts as long as the token, and we want it for the whole
+    // process, so leaking it is deliberate.
+    std::mem::forget(activity);
+
+    tracing::info!("macOS: holding a user-initiated activity assertion (App Nap opt-out)");
+}
+
+/// No-op away from macOS.
+#[cfg(not(target_os = "macos"))]
+pub fn hold_activity_for_process_lifetime() {}
+
+#[cfg(all(test, target_os = "macos"))]
+mod tests {
+    /// Catches a wrong selector or option constant, which aborts the process
+    /// rather than failing an assertion. `tests/macos_app_nap_test.rs` is the
+    /// actual guard.
+    #[test]
+    fn taking_the_activity_assertion_does_not_abort() {
+        super::hold_activity_for_process_lifetime();
+        super::hold_activity_for_process_lifetime();
+    }
+}
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -565,6 +565,11 @@ fn run_headless_entry(
     log_reload_handle: strom::state::LogReloadHandle,
     default_log_filter: String,
 ) -> anyhow::Result<()> {
+    // Before the CFRunLoop starts: without this the process is App-Napped into the
+    // background QoS band ~32 s in, and every pipeline after that runs on
+    // efficiency cores.
+    strom::macos_app_nap::hold_activity_for_process_lifetime();
+
     #[cfg(target_os = "macos")]
     {
         gstreamer::macos_main(move || {
--- a/backend/tests/macos_app_nap_test.rs
+++ b/backend/tests/macos_app_nap_test.rs
@@ -1,0 +1,176 @@
+//! Guards for the App Nap opt-out in `macos_app_nap`.
+//!
+//! ~32 s after launch macOS moves headless Strom into the background QoS band and
+//! every thread drops to scheduling priority 4 — on Apple Silicon, efficiency
+//! cores at a throttled clock, which costs a 1080p x264 flow ~7x its wall clock.
+//! The server holds an `NSProcessInfo` activity assertion to opt out.
+//!
+//! Two failure modes, tested separately because they cost three orders of
+//! magnitude apart: the call going missing from `run_headless_entry` (the
+//! realistic regression, caught in well under a second), and macOS ceasing to
+//! honour the assertion (only observable by waiting out the demotion window, so
+//! opt-in and run by hand).
+#![cfg(target_os = "macos")]
+
+use std::io::Read;
+use std::process::{Command, Stdio};
+use std::time::{Duration, Instant};
+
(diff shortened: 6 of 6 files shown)


wagenet (commented):
Makes sense to me.

srperens (commented):
**Verdict: Comment** — right mechanism and correctly scoped to the headless path, but every
macOS-only line this PR adds has never been compiled by CI.

**Claims**

|---|---|---|
| `objc2-foundation` adds no new crate to the tree | `CONFIRMED` | `Cargo.lock` diff adds only a dependency edge under the `strom` package; no new `[[package]]` block — `eframe` already resolves `objc2-foundation 0.3.2` |
| macOS-specific code compiled by CI on this PR | `CONTRADICTED` | `gh pr checks 735`: `Build (macOS)` is `skipping` |
| The assertion is taken on every path into headless mode | `CONFIRMED` | `backend/src/main.rs:347` and `:359` both call `run_headless_entry`, which calls `hold_activity_for_process_lifetime()` unconditionally at `backend/src/main.rs:571` before the `#[cfg(target_os = "macos")]` block |
| GUI mode never takes the assertion | `CONFIRMED` | Only call site is inside `run_headless_entry`; `run_with_gui` (`backend/src/main.rs:369`) has no such call |

**Diagnosis** — Cause is specific and measured (LaunchServices `Foreground` registration vs.
`gst-launch-1.0`'s `UIElement`, ~32 s demotion window). `beginActivityWithOptions:reason:` is
Apple's documented App Nap opt-out; I cannot verify the `objc2-foundation` binding surface
(method name, `NSActivityOptions` variants) from this repo — no cached crate source and no
build available here, so treat that call as `EXTERNAL (assumed correct per Apple's documented
API)`. Coverage is `ABSOLUTE` for headless mode: one entry point, both call sites route through
it. GUI mode is left untouched on the claim that an open window already keeps the process out
of App Nap — that's an OS-behaviour assumption (`EXTERNAL`), not verified by any test here, but
plausible and not blocking.

**Radius** — `LOCAL`. New module, no existing call sites redirected, `#[cfg(not(target_os =
"macos"))]` no-op elsewhere (`backend/src/macos_app_nap.rs:40`), and the new dependency edge
resolves to a crate version already in the tree.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API
Contract Check` green. But `Build (macOS)` and `Build (Windows)` both `skipping` — this repo's
CI does not build macOS on push/PR, so the `#[cfg(target_os = "macos")]` body in
`backend/src/macos_app_nap.rs:20-37`, its in-module unit test (`:43-53`), and the whole of
`backend/tests/macos_app_nap_test.rs` (`#![cfg(target_os = "macos")]` at line 13) have never
been compiled by this repo. `UNVERIFIED`. Dispatch before merge:
`gh workflow run ci.yml --ref wagenet/flow-cycle-degradation -f platforms=macos`. The
`#[ignore]`d behavioural test (`server_is_not_demoted_to_background_qos`) is correctly excluded
from CI per its own comment (unknown whether the GitHub macOS runner App-Naps at all) — that
one has to stay a manual/opt-in check, but the wiring test does not, and a `platforms=macos`
compile is enough to catch a wrong selector or missing feature flag before merge.

**Repo rules** — No BUFFER probe, no pipeline/element closures, no new shared type, no new
endpoint. The cross-platform no-op correctly uses `#[cfg(not(target_os = "macos"))]` rather
than `#[allow(dead_code)]` (`backend/src/macos_app_nap.rs:40-41`), matching CLAUDE.md's
target-specific dead-code rule directly.

Confidence: HIGH

wagenet:
Dispatched `ci.yml` with `platforms=macos` so the macOS-only code gets compiled. The branch lives on the fork, so the run is there:

https://github.com/wagenet/strom/actions/runs/33835592320

`Build (macOS)` runs clippy with `-D warnings`, `cargo test --features efp,nvidia` under `STROM_REQUIRE_GST_PLUGINS=1`, and the release build — so it covers the `#[cfg(target_os = "macos")]` body in `macos_app_nap.rs`, its selector unit test, and `headless_entry_takes_the_activity_assertion`. `server_is_not_demoted_to_background_qos` stays `#[ignore]`d and does not run.

I will report the result here when it finishes.

-- Claude

wagenet:
`Build (macOS)` is green — 27m14s, clippy + tests + release build: https://github.com/wagenet/strom/actions/runs/33835592320/job/100907688074

The two tests that had never been compiled both ran:

```
test macos_app_nap::tests::taking_the_activity_assertion_does_not_abort ... ok
test headless_entry_takes_the_activity_assertion ... ok
test server_is_not_demoted_to_background_qos ... ignored, waits ~50 s for the App Nap window; set STROM_TEST_APP_NAP=1
```

So the selector and the `objc2-foundation` binding surface are confirmed against a real macOS toolchain, and the wiring test holds. The demotion test stayed `#[ignore]`d as designed.

-- Claude

wagenet:

Raising this here because #735 is where the fork-run question already came up, but it applies identically to #721, #722 and #726: all four are unapproved on the single ground that `Build (macOS)` has never compiled their macOS-only code.

`scripts/agent/REVIEW.md:78-80` gives the remedy as:

```
gh workflow run ci.yml --ref <branch> -f platforms=macos
```

I think that command cannot work as written for these PRs, and it would be worth confirming before it is handed out again. All four branches live on the `wagenet` fork; none exists on `Eyevinn/strom`:

```
$ gh api repos/Eyevinn/strom/branches --paginate --jq '.[].name' | grep -c wagenet
0
```

`workflow_dispatch` resolves `--ref` against a ref in the target repository, and a fork PR's head is reachable there only as `refs/pull/<N>/head`, which `workflow_dispatch` does not accept. So a maintainer would first have to push each branch to `Eyevinn/strom` and dispatch against that copy. I have not been able to run the command to confirm the failure mode — @wagenet's token is `pull`-only on this repo, so any attempt returns 403 regardless of whether the ref would have resolved. Treating this as likely rather than proven.

That leaves the four PRs in a state neither side can clear on its own: the author cannot produce a run in this repo, and the documented maintainer command needs a step that is not in the protocol.

Two ways out, whichever suits the project better:

1. A maintainer pushes the four branches to `Eyevinn/strom` and dispatches `platforms=macos` against each.
2. `ci.yml` gains a `pull_request` trigger for `build-macos` — gated on a path filter or a label, so it only pays the ~27 minutes when a diff actually touches macOS-only code. That would close this class of blocker permanently instead of once per PR.

For the record on this PR specifically: `Build (macOS)` is green on the fork for `b3833fd` (27m14s — clippy `-D warnings`, `cargo test --features efp,nvidia` under `STROM_REQUIRE_GST_PLUGINS`, release build), with both previously-uncompiled tests executing. Per protocol that run does not count, and I am not arguing it should. It is offered only as evidence that a run in this repo would pass, not as a substitute for one.

Affected: #721, #722, #726, #735.

srperens:
@wagenet — answering the fork-run question you raised here, since this is where it came up. Not a review; this PR is merged.

You were right, and option 2 shipped in #768 (merged as `4600c21`).

`Build (macOS)` and `Build (Windows)` now run on every merge to `main`, and on a pull request carrying a `ci:macos` / `ci:windows` label:

Both labels exist in the repository, so the gate is usable now.

Two things pushed the decision to option 2 over pushing branches here, beyond what you listed:

- A `pull_request` run builds **your head merged with main**. A branch copy pushed into this repository builds the branch alone, which for a PR sitting on a week-old merge base answers a slightly different question.
- The repository is public, so standard runners — `macos-latest` included — are not billed. The workflow's own comment saying the extra platforms were off "because of cost" was measuring the wrong thing; what they cost is wall-clock and a concurrency slot, and nothing waits on either job after a merge.

I did **not** take the path-filter variant of your option 2. `target_os = "macos"` sits in six files across `backend/src/gst/`, `backend/src/main.rs` and three frontend files, and this PR added a *new* file. A filter listing known paths fails open exactly when someone puts macOS code somewhere new, which is the case it would exist to catch.

Your 27m14s turned out to be the right number to plan against: a fork's pull request gets no sccache credentials, so it compiles cold either way. The first real run under the new gate took 22m35s.

Also fixed in the same commit: `scripts/agent/REVIEW.md` was handing out `gh workflow run ci.yml --ref <branch>` as the remedy for exactly the situation where it cannot work. It now names the label instead, says a green run on your own fork does not count (it builds your base, not this one), and records that merging on Linux-green while the `main` run covers a small platform-specific part is a legitimate call rather than a silent `UNVERIFIED`. That file being wrong was ours, and it is the part of this that should have been caught before you had to point it out.

`#721`, `#722` and `#726` each have a note about what this means for them.

Issue #711, opened 2026-08-27
h264 4:2:2 10-bit 1080p50 falls back to CPU decode on the GPU node, driving near-full CPU load


alexbj75:
## Summary

An h264 4:2:2 10-bit 1080p50 source cannot be hardware (NVDEC/GPU) decoded on the shared GPU node and
falls back to CPU decode, which was observed driving that machine to near-full CPU utilization on its
own from a single such source.

## Why this matters

4:2:2 10-bit is a routine broadcast-camera output format (used by many professional camera/mixer
setups, including operators coming from vMix or similar full 4:2:2-capable mixers), so this is a
capacity and compatibility gap aimed directly at exactly the higher-end source format that a
professional-camera or vMix-originated feed is likely to arrive in, not an edge case.

## Suggested next step

Confirm whether NVDEC on the current GPU node's driver/CUDA version supports 4:2:2 10-bit h264 decode
at all (some NVDEC generations only support 4:2:0), and if not, document the supported chroma
subsampling / bit depth envelope for hardware decode clearly so operators know to transcode to 4:2:0
8-bit before ingest, or size CPU-node capacity planning around the actual (not the best-case) decode
cost when this format arrives.

srperens:
**Verdict: NEEDS INFO**

## What the code can and can't settle

This repo has two different decode paths for h264, and they'd fail differently for an unsupported profile — which one applies depends on which input block/source type the reporter used (not stated):

- **`builtin.mpegtssrt_input`** (`decode=true`, default) uses a bare `decodebin` (`backend/src/blocks/builtin/mpegtssrt_input.rs:229`) with no manual decoder selection — GStreamer's own autoplugging picks the highest-ranked element whose sink caps template matches the stream. If `nvh264dec`'s template doesn't list High 4:2:2 Profile, decodebin silently skips it and lands on `avdec_h264` — no error, no log line pointing at why, which matches "falls back to CPU decode" with no visible cause.
[...]

PR #736, opened 2026-09-01
docs: add a WIP hardware requirements and sizing doc

srperens:
## What

Adds `docs/HARDWARE_REQUIREMENTS.md` — a starting point for answering "what hardware do I need to run Strom", plus links to it from `docs/README.md` and the prerequisites section of `OPEN_LIVE_SETUP.md`.

Deliberately marked work in progress. It is meant as somewhere people can add their own field experience, and it has an empty **Field reports** table for exactly that.

## Please sanity-check these claims

Review of the reasoning is more useful here than review of the prose:

- **Encoders scale with encoded outputs, not viewers.** Encoders are explicit `videoenc` blocks; a WHEP output fed from `encoded_out` payloads the pre-encoded stream rather than re-encoding per consumer, and an SRT output can share the same encoder. Uplink bandwidth is what grows with audience.
- **Audio is the part that does scale with viewers.** Raw audio into a WHEP output is encoded inside `whepserversink`'s per-consumer pipeline, so Opus cost is roughly tracks x viewers. This one is **unverified** — flagged as such in the doc and a good candidate for a measurement.
- **GL compositing is cheap**, and the GPU argument for a professional card is engine count, thermals and duty cycle rather than raw capability.
- **The cloud vs facility split** — cloud instances take media in and out over the internet only; SDI/NDI/AES67/multicast belong to the facility profile, along with clock discipline and host networking.
- **GPU generation table** (H.264/HEVC/AV1/4:2:2 per generation).

## Sources

- CEF/HTML numbers are the measured ones, taken from `HTML_RENDER.md`.
- Everything else is derived from the code and from vendor-published capabilities. The doc says so up front, per-section where it matters.
- The reference workload is a real two-SRT-input flow, described in anonymised form (no ids, endpoints or host details).

## Testing

Docs only — no code changed. Pre-commit checks (fmt, clippy, sensitive-content scan) passed. No test suite was run because nothing in `backend/` or `frontend/` was touched.

### DIFF
--- a/docs/HARDWARE_REQUIREMENTS.md
+++ b/docs/HARDWARE_REQUIREMENTS.md
@@ -1,81 +1,71 @@
 # Hardware Requirements (Work in Progress)
 
-> **Status: work in progress.** We have not load-tested Strom across a range of machines, so
-> unless a section says otherwise the numbers below are derived from what the code requires and
-> from vendor-published capabilities rather than measured. Treat them as starting points for
-> sizing.
+> **Status: work in progress. Code is the source of truth — this may have drifted; read the
+> code for the current implementation.** We have not load-tested Strom across a range of
+> machines, so unless a section says otherwise the numbers here are derived from what the code
+> requires and from vendor-published capabilities rather than measured.
 >
 > If you have run Strom on real hardware, please add a line to
-> [Field reports](#field-reports) — collecting real experience here is the main point of this
+> [Field reports](#field-reports) — collecting real experience is the main point of this
 > document.
+
+**Sizing in three lines:**
+
+- **Cloud production**: roughly 8 cores, 16 GB, an L4-class NVIDIA GPU. Uplink is
+  bitrate x concurrent viewers, and usually binds before the GPU does.
+- **Graphics-heavy production**: budget CPU per HTML source. That is normally the largest single
+  cost in the system.
+- **4:2:2 contribution**: CPU cores, not a bigger GPU. It never touches the hardware decoder.
+
+Everything below is the reasoning behind those numbers.
 
 ## Strom and Open Live
 
-Strom is a standalone GStreamer flow engine with its own UI and API — see the
-[root README](../README.md) for what it is on its own.
-[Open Live](https://github.com/Eyevinn/open-live) is one of the things that use it: it keeps its
-own configuration in its own database and drives Strom over REST/WebSocket, pushing flows across
-at runtime. Strom holds those flows in memory and runs the pipelines.
+Strom is a standalone GStreamer flow engine with its own UI and API (see the
+[root README](../README.md)). [Open Live](https://github.com/Eyevinn/open-live) is one of the
+things that use it: it holds its own configuration in its own database and pushes flows to Strom
+over REST/WebSocket at runtime, which Strom keeps in memory and runs.
 
-Either way, a Strom host is sized by the media it processes — not by the number of operators
-using the UI, and not by how the flows got there. Many of the examples here come from Open Live
-setups, since that is where most of our numbers come from.
+Either way the host is sized by the media it processes, not by how the flows got there. Most
+examples here come from Open Live setups, since that is where our operational experience is.
 
-For setup instructions see [OPEN_LIVE_SETUP.md](OPEN_LIVE_SETUP.md) and
-[DOCKER_GPU_SETUP.md](DOCKER_GPU_SETUP.md).
+**Primary target: Linux + Docker with an NVIDIA GPU** — what runs live and what gets tested; see
+[OPEN_LIVE_SETUP.md](OPEN_LIVE_SETUP.md) and [DOCKER_GPU_SETUP.md](DOCKER_GPU_SETUP.md). Other
+platforms build and suit development, with the caveats in [Known limits](#known-limits). H.264
+is what runs in production, so do not size around HEVC or AV1.
 
-**Primary target: Linux + Docker with an NVIDIA GPU.** That is what runs live and what gets
-tested. Native Linux, macOS, and Windows builds exist and are useful for development, but the
-platform caveats in [Known limits](#known-limits) apply. **H.264 is the codec that runs in
-production today** — HEVC and AV1 are supported by the code but are not part of the live path,
-so do not size a machine around them.
+### Two deployment profiles
 
-### Two deployment profiles: cloud and facility
+**Cloud instance — network only.** Media arrives and leaves over the internet: SRT in,
+WHEP/WHIP and SRT out. No SDI, NDI, AES67, multicast or house clock. A cloud VM or rack server
+with a GPU and decent connectivity. Open Live has so far focused on this profile.
 
-Where the media comes from changes what the host needs, so it helps to know which of these you
-are building.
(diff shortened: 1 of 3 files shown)


srperens (commented):
**Verdict: Comment** — the two load-bearing claims check out against the code; one repo-convention gap worth fixing before this leaves WIP.

**Claims**

|---|---|---|
| A WHEP output fed from `encoded_out` payloads pre-encoded video without re-encoding per consumer | `CONFIRMED` | `backend/src/blocks/builtin/whep.rs:1632-1660` — `video/x-h264`/`h265`/`vp9`/`av1` caps are passed straight through, restricting `whepserversink`'s `video-caps` to that same codec; only the `video/x-raw` branch asks it to encode |
| Raw audio into a WHEP output is encoded inside `whepserversink`'s own per-consumer pipeline | `CONFIRMED` | `backend/src/blocks/builtin/whep.rs:946` `` `whepserversink is based on webrtcsink and handles encoding internally` ``, and the `audio/x-raw` branch (`backend/src/blocks/builtin/whep.rs:1528-1530`) leaves `audio-caps` at its encoded default rather than passing raw through |
| Opus cost scales as tracks × viewers | `EXTERNAL` | Doc already labels this unverified/unmeasured itself; the mechanism above supports the shape of the claim but not the number |
| 4:2:2 sources always decode on the CPU | `EXTERNAL` | No 4:2:2-specific decode-path logic in this repo (`grep` for `4:2:2`/`422`/`I422` in `backend/src` — no hits); this is a vendor hardware-capability fact, consistent with the doc's own Blackwell "hardware yes, unusable" row, not something this codebase encodes |
| This guide follows the repo's existing disclaimer convention for operator/usage docs | `CONTRADICTED` | `docs/AUDIO_MIXER_OPERATOR_GUIDE.md`, `docs/HTML_RENDER.md`, `docs/VISION_MIXER_OPERATOR_GUIDE.md` and `docs/STREAM_SYNCHRONIZATION.md:3` (`` `**Code is the source of truth.** This guide describes intended behaviour and may have` ``) all carry it; `docs/HARDWARE_REQUIREMENTS.md` has only a "work in progress" callout, no drift disclaimer |

**Diagnosis** — Both claims the PR specifically asks about check out mechanically: `encoded_out`'s passthrough and `whepserversink`'s internal per-consumer encode are real, current behaviour, not assumptions. CLAUDE.md's disclaimer rule is worded around docs staying at top level, so this is not a strict violation for a `docs/`-level file — but every comparable guide already in `docs/` carries the sentence, so its absence here reads as an oversight rather than a deliberate choice, given how much of this doc's content (encoder/decode counts, per-generation GPU table) is exactly the kind of thing CLAUDE.md warns drifts.

**Blast radius** — `LOCAL`: three files, no code touched — a new doc plus one link each from `docs/README.md` and `OPEN_LIVE_SETUP.md`'s prerequisites section.

**Tests & CI** — `Build (Linux x86_64/ARM64)`, `API Contract Check`, `Check & Build (WASM)`, `Check (Linux)` all green at `bdf9d72`; macOS/Windows `skipping` as expected. Docs-only, so no test suite applies — matches what the PR body states.

**Design record** — First review; no prior marker.

Confidence: HIGH

svensson00:
Really strong doc — the honest measured-vs-derived framing and the "what actually drives
the load" section are the right way to do this, and the "Checking a specific machine"
commands mean a reader can always trust their own box over the table. All the relative
link targets resolve on the branch, and the #711 reference checks out. A handful of
suggestions before merge:

1. **"most of our numbers" vs. the WIP header.** The intro says examples come from Open
   Live setups "since that is where most of our numbers come from", but the header (and
   Known limits) say almost nothing is measured. Suggest "…where our operational
   experience comes from" so the two don't contradict.

2. **Link #711 in full.** The 4:2:2 bullet cites "issue #711" bare — it's accurate
   (open, exactly this symptom), but this doc will travel outside the repo (it's what
   gets forwarded to a prospect's infra team), where a bare number doesn't resolve.
   Same for the NVENC session-limit history ("3, then 5, then 8") — a link to NVIDIA's
   official session-limit table would make that durable.

3. **Disk minimum of 4 GB looks too tight.** strom-full is stated at ~2.7 GB
   uncompressed, and a docker pull transiently needs compressed + uncompressed space on
   top of the OS. 10 GB seems a safer floor; the "Size for recordings and media, not for
   flow config" note is the important part and stays.

4. **Caveat the CEF GPU-mode numbers inline.** The ~50% CPU floor per `cefsrc` was
   measured on one machine (RTX 3090). Since these are the doc's only measured figures,
   a "(single machine — treat as one data point)" alongside keeps the doc's own
   epistemic standard.

5. **Say how to submit a field report.** The table is the doc's stated main point, but
   "add a line" doesn't say how. One sentence — "PR against this file, or open an issue
   titled `field report: <setup>` and we'll add it" — probably decides whether it ever
   gets a row.

6. **Consider surfacing the per-viewer audio cost in the Baseline host table.** The
   caveat under the reference workload that "Opus cost is roughly tracks x viewers" is
   the one per-viewer *compute* cost in the system; a pointer in the CPU or network row
   would stop someone sizing purely off the headline point that encoding does not grow
   with viewers.

7. **Nit / addition:** a three-line "TL;DR sizing" at the very top (cloud production ≈
   8 cores / 16 GB / L4-class GPU, egress = bitrate × viewers; graphics-heavy = budget
   CPU per HTML source; 4:2:2 = CPU cores, not GPU) would serve the person forwarding
   this straight to procurement — everything justifying it already exists below.

None of these block merging a WIP-labeled doc as far as I'm concerned — 1–3 are quick
fixes, the rest could land as follow-ups.

srperens:
Thanks — all seven applied in 8acddc7, plus the disclaimer the review bot flagged and a length pass.

| # | Change |
|---|---|
| 1 | "most of our numbers" → "where our operational experience is" — no longer contradicts the header |
| 2 | [#711](https://github.com/Eyevinn/strom/issues/711) and NVIDIA's [GPU support matrix](https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new) linked in full. The bare "3, then 5, then 8" history is gone — the vendor table is the durable source, and our own figure was unverified anyway |
| 3 | Disk minimum 4 → 10 GB, with the transient pull space spelled out |
| 4 | CEF section now opens with "all from a single machine (an RTX 3090) — treat them as one data point" |
| 5 | "open a PR against this file, or an issue titled `field report: <setup>` and we will add it for you" |
| 6 | CPU row now reads "…and for Opus (tracks x viewers)" |
| 7 | Three-line sizing summary at the top, above everything else |

Also added the repo's standard **"Code is the source of truth — this may have drifted"** disclaimer, which as the review bot correctly pointed out every comparable guide in `docs/` already carries and this one was missing.

Separately, another reader found the doc long and repetitive, which was fair: encoders-per-output, 4:2:2, CEF cost and GL compositing were each stated in three or four places. Each is now stated once with pointers to it. Down roughly 15% in prose despite the additions above — the tables are most of what remains, and they seem to be the part people actually want.

One thing I did **not** change: the Opus tracks x viewers claim is still labelled unmeasured. The mechanism is confirmed in the code, the magnitude is not, and I would rather it stay flagged until someone measures it.

PR #744, opened 2026-09-02
feat(blocks): add builtin.audioenc, an audio encoder block

wagenet:
## What

Adds `builtin.audioenc`, an audio encoder block, as the counterpart to the existing `builtin.videoenc`.

The Recorder only accepts pre-encoded audio, and the only audio encoders in the tree were buried inside the SRT and WHIP output blocks. A raw audio pad — a decoded WHIP input feeding a vision mixer, say — therefore could not be recorded at all.

The block builds the chain:

```
audioconvert -> audioresample -> capsfilter -> encoder -> parser -> capsfilter
```

Codecs: AAC (default), Opus, MP3, AC-3, with `bitrate`, `sample_rate` and `channels` properties. `sample_rate` and `channels` both accept `"source"` to leave the upstream value untouched.

## Details worth review

- The encoder is the first available candidate for the codec, skipping any factory whose rank is NONE, so `GST_PLUGIN_FEATURE_RANK=element:0` still works as the way to take an encoder out of play.
- Bitrate is applied with `set_property_from_str` throughout, avoiding the enum-property panic class. `lamemp3enc` is special-cased: it takes kbps rather than bps, and only honours the value when `target=bitrate`.
- Opus accepts only 8/12/16/24/48 kHz. An illegal rate is rejected at build time with a message naming the rate, rather than surfacing later as a not-negotiated error on a running flow.
- AAC output caps deliberately leave `stream-format` unpinned, since `aacparse` converts between raw and adts and pinning either breaks the muxers wanting the other.

## Testing

Which commands actually ran, and where. All of the below ran against `7b2ab3b`, the current head, on macOS:

- `cargo clippy --workspace --all-targets -- -D warnings` — **passed**. `--all-targets` covers the lib-test target, so the block and its test module both typecheck.
- `cargo fmt --all -- --check` — **passed**.
- The 12 `audioenc` unit tests — **12 passed**. This includes `test_build_produces_the_full_chain`, which really executed rather than skipping: an AAC encoder is present on this machine. An earlier revision of this section said it had been skipped here; that was wrong.
- `openapi_test` — **1 passed**. `pipeline_lifecycle_test` — **3 passed**.
- The skip guard was exercised deliberately: with `GST_PLUGIN_FEATURE_RANK=fdkaacenc:0,avenc_aac:0,voaacenc:0,faac:0` and `STROM_REQUIRE_GST_PLUGINS=1`, `test_build_produces_the_full_chain` fails on the new assertion instead of returning early. Without the env var it skips with a message, as intended.

No end-to-end flow was run through the new block on real media. The evidence above is compile-level plus unit tests.

### DIFF
--- a/backend/src/blocks/builtin/audioenc.rs
+++ b/backend/src/blocks/builtin/audioenc.rs
@@ -1,0 +1,445 @@
+//! Audio encoder block. Codecs: AAC (default), Opus, MP3, AC-3.
+//!
+//! Chain: audioconvert -> audioresample -> capsfilter -> encoder -> parser -> capsfilter
+//!
+//! The rate capsfilter pins sample rate and channels when configured, the parser gives
+//! downstream muxers framed data, and the final capsfilter pins the output codec.
+
+use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult, BlockBuilder};
+use gstreamer as gst;
+use gstreamer::prelude::*;
+use std::collections::HashMap;
+use strom_types::{block::*, element::ElementPadRef, EnumValue, PropertyValue, *};
+use tracing::{info, warn};
+
+/// Audio Encoder block builder.
+pub struct AudioEncBuilder;
+
+/// Default target bitrate in kbps.
+const DEFAULT_BITRATE_KBPS: u32 = 128;
+
+/// Default output sample rate. WebRTC ingest is 48 kHz and Opus accepts only a fixed
+/// set of rates, so 48000 keeps every codec negotiable.
+const DEFAULT_SAMPLE_RATE: &str = "48000";
+
+/// Sample rates Opus can encode. Anything else has to be resampled first.
+const OPUS_SAMPLE_RATES: [i32; 5] = [8000, 12000, 16000, 24000, 48000];
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum Codec {
+    Aac,
+    Opus,
+    Mp3,
+    Ac3,
+}
+
+impl BlockBuilder for AudioEncBuilder {
+    fn build(
+        &self,
+        instance_id: &str,
+        properties: &HashMap<String, PropertyValue>,
+        _ctx: &BlockBuildContext,
+    ) -> Result<BlockBuildResult, BlockBuildError> {
+        info!("Building AudioEncoder block instance: {}", instance_id);
+
+        let codec = parse_codec(properties)?;
+        let bitrate_kbps = parse_bitrate(properties);
+        let sample_rate = parse_sample_rate(properties)?;
+        let channels = parse_channels(properties)?;
+
+        // Catch this here rather than as a not-negotiated error on a running flow.
+        if codec == Codec::Opus {
+            if let Some(rate) = sample_rate {
+                if !OPUS_SAMPLE_RATES.contains(&rate) {
+                    return Err(BlockBuildError::InvalidConfiguration(format!(
+                        "Opus cannot encode at {} Hz — supported rates are {:?}",
+                        rate, OPUS_SAMPLE_RATES
+                    )));
+                }
+            }
+        }
+
+        let encoder_name = select_encoder(codec)?;
+        info!(
+            "Selected encoder '{}' for codec {:?} at {} kbps",
+            encoder_name, codec, bitrate_kbps
+        );
+
+        let convert_id = format!("{}:audioconvert", instance_id);
+        let resample_id = format!("{}:audioresample", instance_id);
+        let rate_caps_id = format!("{}:rate_caps", instance_id);
+        let encoder_id = format!("{}:encoder", instance_id);
+        let parser_id = format!("{}:parser", instance_id);
+        let capsfilter_id = format!("{}:capsfilter", instance_id);
+
+        let audioconvert = gst::ElementFactory::make("audioconvert")
+            .name(&convert_id)
+            .build()
+            .map_err(|e| BlockBuildError::ElementCreation(format!("audioconvert: {}", e)))?;
+
+        let audioresample = gst::ElementFactory::make("audioresample")
+            .name(&resample_id)
+            .build()
+            .map_err(|e| BlockBuildError::ElementCreation(format!("audioresample: {}", e)))?;
+
+        let rate_caps = gst::ElementFactory::make("capsfilter")
+            .name(&rate_caps_id)
+            .property("caps", build_raw_caps(sample_rate, channels))
+            .build()
+            .map_err(|e| BlockBuildError::ElementCreation(format!("capsfilter: {}", e)))?;
+
+        let encoder = gst::ElementFactory::make(&encoder_name)
+            .name(&encoder_id)
+            .build()
+            .map_err(|e| BlockBuildError::ElementCreation(format!("{}: {}", encoder_name, e)))?;
+
+        set_encoder_bitrate(&encoder, &encoder_name, bitrate_kbps);
+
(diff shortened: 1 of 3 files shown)


srperens (commented):
**Verdict: Comment** — right fix at the right layer, and it closes the gap the Recorder itself has been asking for. Two things to change before it leaves draft, neither of them structural.

**Requested changes**

1. Correct the `sample_rate` description. `backend/src/blocks/builtin/audioenc.rs:376-379` tells the operator that "Match source" makes Opus reject any rate outside 8/12/16/24/48 kHz. On the chain this block actually builds it does not: `backend/src/blocks/builtin/audioenc.rs:319` — `    if value == "source" {` — pins no `rate` field, and `backend/src/blocks/builtin/audioenc.rs:80` — `        let audioresample = gst::ElementFactory::make("audioresample")` — sits upstream of the encoder, so a 44.1 kHz source is resampled to a rate `opusenc` accepts rather than refused. EXTERNAL assumption stated: that `opusenc`'s sink template restricts `rate` to those five values, which is also what makes the build-time check in `build()` correct for an explicitly pinned rate. The text warns about the one configuration that is safe.
2. Have the skip guard fail when the repo says it must. `backend/src/blocks/builtin/audioenc/tests.rs:165` — `    if select_encoder(Codec::Aac).is_err() {` — returns silently, where three sibling tests assert instead: `backend/tests/recorder_unfed_track_test.rs:53` — `        strom_types::env::var_opt("STROM_REQUIRE_GST_PLUGINS").is_none(),`. This is not hypothetical bookkeeping — `.github/workflows/ci.yml:95` — `      # object file"), avenc_aac goes missing, and every test requiring it skips` — documents that exact regression happening in this repo's own Linux job. Two lines, matching the existing convention.

**Claims**

|---|---|---|
| Raw audio could not reach the Recorder before this | `CONFIRMED` | `backend/src/blocks/builtin/recorder.rs:676` — `                            "Recorder {}: received raw audio — recorder only accepts pre-encoded audio. Add an encoder block before the recorder.",` — and the block it names did not exist for audio |
| All four codecs the block emits are accepted downstream | `CONFIRMED` | `recorder.rs:684-689` maps `audio/mpeg` (both mpegversions), `audio/x-ac3` and `audio/x-opus` to a parser; anything else hits the `warn!` and releases the pad |
| The 12 unit tests ran in CI at this head, not just in the author's tree | `CONFIRMED` | `Check (Linux)` run 33656506620: 12 `audioenc::tests::` lines, including `test_build_produces_the_full_chain ... ok`, and the guard's `skipping: no AAC encoder available` never printed — so it asserted rather than returned early |
| Encoder selection, rank handling and bitrate units follow the existing block | `CONFIRMED` | `backend/src/blocks/builtin/videoenc.rs:313` and `:630` use the same `Rank::NONE` skip and the same `bitrate * 1000`, so the multiply is a house convention, not a new hazard |

**Diagnosis** — Additive; nothing existing changes behaviour. Coverage is `ABSOLUTE` for the reported gap: every codec the block can emit is one the Recorder's probe already parses. Worth recording, since the PR body does not: the block's own parser is redundant when the consumer is the Recorder, which inserts a second one from that same probe, and load-bearing for a muxer that inserts none. Keeping it is right — AAC into the Recorder just runs `aacparse` twice, a passthrough on framed data.

**Radius** — `LOCAL`. Three registration lines in `builtin/mod.rs`, no existing block touched, no thread, timer, probe or pipeline reference taken, and `bus_message_handler: None`, so there is no teardown path to find.

**Tests & CI** — all eight checks green at `fe6fedb`. No platform-specific code here at all — it is element wiring — so Linux-green covers it and the `ci:macos`/`ci:windows` labels are not worth spending a run on.

Confidence: HIGH

wagenet:
Both requested changes are in `7b2ab3b`.

**1. `sample_rate` description.** You are right that it warned about the safe configuration. `build_raw_caps` omits the `rate` field entirely when the value is `source`, so `audioresample` is free to settle on whatever `opusenc`'s sink template accepts. The hazard is the opposite one: an explicitly pinned 44.1 or 32 kHz, which `build()` rejects. New text:

> Output sample rate. "Match source" pins no rate, letting the resampler settle on one the encoder accepts. Opus encodes only at 8/12/16/24/48 kHz, so pinning 44.1 or 32 kHz with Opus is rejected at build time.

**2. Skip guard.** `test_build_produces_the_full_chain` now asserts `STROM_REQUIRE_GST_PLUGINS` is unset before returning early, matching `recorder_unfed_track_test.rs`. Verified both ways locally: all 12 tests pass normally, and with `GST_PLUGIN_FEATURE_RANK=fdkaacenc:0,avenc_aac:0,voaacenc:0,faac:0 STROM_REQUIRE_GST_PLUGINS=1` the test fails with "STROM_REQUIRE_GST_PLUGINS is set but no AAC encoder is available" instead of skipping.

Also ran `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` against this commit; both pass.

One correction to the PR body's testing note: an AAC encoder is present on this machine after all, so `test_build_produces_the_full_chain` executed here rather than skipping.

srperens (approved):
**Verdict: Approve** — supersedes my review of `fe6fedbb`, raising the verdict from Comment. Both requested changes are done and correct, and nothing else in the diff moved. The four rows I confirmed at `fe6fedbb` — the Recorder gap, codec acceptance downstream, the 12 unit tests executing in CI, and the shared conventions with `videoenc` — are unchanged and not re-verified here.

**Claims**

|---|---|---|
| The rewritten `sample_rate` help now matches the chain the block builds | `CONFIRMED` | "Match source" pins no rate — `backend/src/blocks/builtin/audioenc.rs:164` — `    if let Some(rate) = sample_rate {` — so the resampler upstream settles it; and the new sentence's "rejected at build time" is real for an explicitly pinned rate — `backend/src/blocks/builtin/audioenc.rs:53` — `                if !OPUS_SAMPLE_RATES.contains(&rate) {` |
| The skip guard now fails where the repo says it must | `CONFIRMED` | `backend/src/blocks/builtin/audioenc/tests.rs:167` — `            strom_types::env::var_opt("STROM_REQUIRE_GST_PLUGINS").is_none(),` — the same shape the sibling tests use, so a vanished `avenc_aac` turns the Linux job red instead of quietly green |
| The guard did not simply start skipping | `CONFIRMED` | `Check (Linux)` job 102713716495 at `7b2ab3b6`: 12 `audioenc::tests::` lines, `test blocks::builtin::audioenc::tests::test_build_produces_the_full_chain ... ok`, and `skipping: no AAC encoder available` never printed |

**Diagnosis** — unchanged from `fe6fedbb`. Still additive, still `ABSOLUTE` coverage of the reported gap, and no existing block's behaviour is touched. The second commit is the more valuable of the two: the first corrects text, the second converts a test that could have silently stopped guarding into one that cannot, which is the failure mode `.github/workflows/ci.yml` already documents happening in this repo's own Linux job.

**Radius** — `LOCAL`, unchanged. The delta is one description string and one assertion inside an existing early return; no element wiring, no registration, no lifetime surface.

**Tests & CI** — run 34426815061 at `7b2ab3b6`: `Check (Linux)`, both `Build (Linux)` jobs, `Check & Build (WASM)` and `API Contract Check` all pass. `Build (macOS)` and `Build (Windows)` `skipping`, which is not a gap — the diff is element wiring with no `cfg` or platform code, so Linux-green covers it and the `ci:macos`/`ci:windows` labels are not worth a run. Nothing was built or executed locally for this review; every CI statement is read off the concluded run.

Confidence: HIGH

PR #748, opened 2026-09-03
fix(vision mixer): keep per-pixel alpha on keyed pads with an alpha-less output_format

wagenet:
## The bug

The vision mixer's `output_format` property offers alpha-less formats (I420, NV12, RGB, ...). Setting one on the CPU path destroyed the per-pixel alpha of every keyed pad: a DSK graphic's transparent areas composited as the graphic's own black instead of the program picture, and the full-canvas multiview overlay turned the whole multiview black. There is a standing incentive to set the property: unpinned RGBA into `vtenc_h264_hw` negotiates AYUV64 rather than NV12.

## Cause

Two places, one mechanism.

The builder inserted `output_format` capsfilters on the DSK chains and on the overlay appsrc chain, flattening alpha before the compositor saw it. Removing those alone does not fix it — measured, not assumed. `compositor` blends in its *own output format* and converts every sink pad to that format first, so pinning an alpha-less format on the compositor's src caps makes its sink pads alpha-less too:

```
compositor ! video/x-raw,format=NV12   # RGBA input: not-negotiated
compositor ! videoconvert ! video/x-raw,format=NV12   # RGBA input: key survives
```

The hazard was already known in this file — the border underlays pin RGBA with a comment saying an alpha-less `output_format` "would silently drop the alpha of #RRGGBBAA border colors" — but the conclusion was never applied to the compositor's own output.

## Fix

When a compositor has keyed pads and `output_format` cannot carry alpha, blend in the alpha-carrying counterpart of the requested format and convert to `output_format` after the mixer. `output_format` still pins what leaves the block; it no longer dictates the blend space. The dist compositor gets this when DSK inputs or PiP border underlays exist; the multiview always does, since its overlay is a full-canvas RGBA pad. The DSK and overlay capsfilters are dropped; the video-input capsfilters stay, since their tee feeds two independently negotiating compositors.

**Cost, for the reviewer to weigh.** Both compositor outputs gain a full-frame conversion per frame, where the compositor previously blended straight into `output_format`. That is inherent to the fix and a real cost in a block being optimised elsewhere. The blend space matches the requested chroma subsampling, so the added hop is an alpha strip, not a resample. Only flows with a keyed pad and a non-`Auto` `output_format` pay it.

**The GPU path.** `builder/pipeline_gpu.rs` is unchanged, but not because it ignores `output_format` — it pins it at `capsfilter_dist` and `capsfilter_mv`. Keyed pads survive there because `glvideomixer` is a *bin* whose internal converter absorbs the pin, so the blend stays RGBA in GL memory. Verified by reading the internal mixer's src caps with an NV12 pin downstream: still `format=(string)RGBA` on `GstGLVideoMixer:m-mixer`. The general invariant now sits as a comment there: keep a converting element between a mixer and any format pin.

## Tests

New `backend/tests/vision_mixer_keyed_alpha_test.rs` builds a real flow through `PipelineManager` (following `vision_mixer_fx_test.rs`): a white program input and a DSK graphic that is opaque red on the left half and fully transparent on the right, CPU backend. One test per branch of `alpha_blend_format`, so a format that loses its key is named by the failure rather than hidden behind the first one to break.

| `output_format` | blend space | transparent half white, fix reverted | with fix |
|---|---|---|---|
| NV12 | A420 | 0.000 | 1.000 |
| YUY2 | A422 | 0.000 | 1.000 |
| v210 | A422_10LE | 0.000 | 1.000 |
| RGB | RGBA | 0.000 | 1.000 |
| Auto | none | 1.000 | 1.000 |

Each also asserts the requested format negotiates at `capsfilter_dist` and that the multiview shows picture. Reverting the blend substitution turns all four red and leaves `Auto` green. `A422` and `A422_10LE` are on `compositor`'s src pad template (checked with `gst-inspect-1.0` against gst-plugins-base 1.28.6), and each blend-to-output pair negotiates through `videoconvert`. `UYVY` shares the `A422` branch with `YUY2`, so it is covered by code path but has no flow test of its own.

Two things the measurement has to get right, both of which produced a green test that guarded nothing until fixed. The multiview reading is gated on a buffer probe counting overlay appsrc pushes, so it cannot pass by measuring a multiview the overlay has not reached yet. And readiness is detected by colour, not brightness: an all-black startup frame is dark in exactly the way an opaque-keyed failure is dark.

Run locally on macOS (Apple silicon, GStreamer 1.28.6): the full `cargo test` suite green, plus `cargo clippy --all-targets -D warnings`. The new test has no skip path and needs only `compositor` and `videobox`, both already in the CI package list on every platform. Not run: NVIDIA/CUDA paths, and the GPU path's flow behaviour.

### DIFF
--- a/backend/src/blocks/builtin/vision_mixer/builder/mod.rs
+++ b/backend/src/blocks/builtin/vision_mixer/builder/mod.rs
@@ -13,6 +13,7 @@ use super::properties;
 use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult, BlockBuilder};
 use gstreamer as gst;
 use gstreamer_app as gst_app;
+use gstreamer_video as gst_video;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 use strom_types::vision_mixer;
@@ -239,36 +240,76 @@ impl<'a> PipelineParams<'a> {
         format!("{}:{}", self.instance_id, name)
     }
 
-    /// Build PGM output caps with resolution, framerate, and optional pixel format.
-    pub(super) fn pgm_caps(&self) -> gst::Caps {
+    /// Raw video caps with resolution, framerate and an optional pixel format.
+    fn raw_caps(&self, w: u32, h: u32, framerate: (i32, i32), format: Option<&str>) -> gst::Caps {
         let mut builder = gst::Caps::builder("video/x-raw")
-            .field("width", self.pgm_w as i32)
-            .field("height", self.pgm_h as i32)
-            .field(
-                "framerate",
-                gst::Fraction::new(self.pgm_framerate.0, self.pgm_framerate.1),
-            )
+            .field("width", w as i32)
+            .field("height", h as i32)
+            .field("framerate", gst::Fraction::new(framerate.0, framerate.1))
             .field("pixel-aspect-ratio", gst::Fraction::new(1, 1));
-        if let Some(ref fmt) = self.output_format {
-            builder = builder.field("format", fmt.as_str());
+        if let Some(fmt) = format {
+            builder = builder.field("format", fmt);
         }
         builder.build()
     }
 
+    /// Build PGM output caps with resolution, framerate, and optional pixel format.
+    pub(super) fn pgm_caps(&self) -> gst::Caps {
+        self.raw_caps(
+            self.pgm_w,
+            self.pgm_h,
+            self.pgm_framerate,
+            self.output_format.as_deref(),
+        )
+    }
+
     /// Build multiview output caps with resolution, framerate, and optional pixel format.
     pub(super) fn mv_caps(&self) -> gst::Caps {
-        let mut builder = gst::Caps::builder("video/x-raw")
-            .field("width", self.mv_w as i32)
-            .field("height", self.mv_h as i32)
-            .field(
-                "framerate",
-                gst::Fraction::new(self.mv_framerate.0, self.mv_framerate.1),
-            )
-            .field("pixel-aspect-ratio", gst::Fraction::new(1, 1));
-        if let Some(ref fmt) = self.output_format {
-            builder = builder.field("format", fmt.as_str());
+        self.raw_caps(
+            self.mv_w,
+            self.mv_h,
+            self.mv_framerate,
+            self.output_format.as_deref(),
+        )
+    }
+
+    /// PGM caps with `format` instead of `output_format` — the software
+    /// compositor's blend space, see [`Self::alpha_blend_format`].
+    pub(super) fn pgm_blend_caps(&self, format: &str) -> gst::Caps {
+        self.raw_caps(self.pgm_w, self.pgm_h, self.pgm_framerate, Some(format))
+    }
+
+    /// Multiview caps with `format` instead of `output_format` — the software
+    /// compositor's blend space, see [`Self::alpha_blend_format`].
+    pub(super) fn mv_blend_caps(&self, format: &str) -> gst::Caps {
+        self.raw_caps(self.mv_w, self.mv_h, self.mv_framerate, Some(format))
+    }
+
+    /// The format the software compositor has to blend in so that keyed pads
+    /// keep their per-pixel alpha, or `None` when `output_format` needs no
+    /// substitute.
+    ///
+    /// `compositor` blends in its own output format and converts every sink
+    /// pad to it first, so an alpha-less `output_format` (I420, NV12, RGB, ...)
+    /// strips the alpha of DSK graphics, the multiview overlay and the border
+    /// underlays. Blending in the alpha-carrying counterpart and converting
+    /// once on the way out keeps the key and still hands `output_format`
+    /// downstream. Alpha-carrying and unknown formats are left alone.
+    pub(super) fn alpha_blend_format(&self) -> Option<&'static str> {
+        let fmt = self.output_format.as_deref()?;
+        let format = fmt.parse::<gst_video::VideoFormat>().ok()?;
+        let info = gst_video::VideoFormatInfo::from_format(format);
+        if info.has_alpha() {
+            return None;
         }
-        builder.build()
+        // Match the chroma subsampling of the requested format so the extra
+        // hop costs nothing the output would not have cost anyway.
+        Some(match fmt {
+            "YUY2" | "UYVY" => "A422",
+            "v210" => "A422_10LE",
+            _ if info.is_rgb() || info.is_gray() => "RGBA",
+            _ => "A420",
+        })
     }
 
     /// Build PGM output caps for the GL-memory passthrough path.
--- a/backend/src/blocks/builtin/vision_mixer/builder/pipeline_cpu.rs
+++ b/backend/src/blocks/builtin/vision_mixer/builder/pipeline_cpu.rs
@@ -18,6 +18,9 @@ pub(super) fn build_cpu_pipeline(
     let mut elems: Vec<(String, gst::Element)> = Vec::new();
     let mut links: Vec<(ElementPadRef, ElementPadRef)> = Vec::new();
 
+    // Conversion element for this host — see the DSK chain below for why.
+    let vc_factory = gpu::video_convert_mode().element_name();
+
     let dist_comp = elements::make_dist_compositor(p.backend, p.latency_ms, p.min_upstream_ms)?;
     let mv_comp = elements::make_mv_compositor(p.backend, p.latency_ms, p.min_upstream_ms)?;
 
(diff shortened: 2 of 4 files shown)


srperens (commented):
**Verdict: Comment** — the mechanism is right and the guard is a real one, but two statements in the body do not survive checking. Both are body corrections, not code changes: I found nothing in the diff I would change.

**Requested changes**

1. Correct the GPU-path sentence. `pipeline_gpu.rs` does reference `output_format` — `backend/src/blocks/builtin/vision_mixer/builder/pipeline_gpu.rs:109` — `            .property("caps", p.pgm_caps())` and `mv_caps()` at `:206` both embed it. The conclusion holds for a better reason: there the capsfilter sits *after* `gldownload`, so the pin is absorbed and never reaches the blend space, whereas on the CPU path it sat on `mixer:src` with no converting element between. That invariant is what a future edit must preserve.

2. Say what happens for `YUY2`, `UYVY` and `v210`. `backend/src/blocks/builtin/vision_mixer/builder/mod.rs:308` — `            "YUY2" | "UYVY" => "A422",` and `:309` — `            "v210" => "A422_10LE",` are reachable from the property's own menu — `types/src/block.rs:479` — `    ("YUY2", "YUY2 (YUV 4:2:2 packed)"),` — and nothing exercises them.

**Claims**

|---|---|---|
| `pipeline_gpu.rs` never referenced `output_format` | `CONTRADICTED` | `backend/src/blocks/builtin/vision_mixer/builder/pipeline_gpu.rs:109` — `            .property("caps", p.pgm_caps())`; `pgm_caps()` adds `format` whenever `output_format` is set. The no-change conclusion is still correct (see 1) |
| The dist gate matches where keyed pads actually exist | `CONFIRMED` | `backend/src/blocks/builtin/vision_mixer/builder/pipeline_cpu.rs:76` — `    let dist_keyed = p.num_dsk_inputs > 0 || p.num_pips > 0;` mirrors the underlay build gate at `:285` — `    if p.num_pips > 0 {` |
| The new test guards the change and runs in CI | `CONFIRMED` | `backend/tests/vision_mixer_keyed_alpha_test.rs:404` — `    let m = run_flow("vmk_nv12", "NV12");` asserts the transparent half reads white, which is the value the bug inverts; no element guard or early return anywhere in the file, so it cannot skip green |
| `compositor` advertises `A422` and `A422_10LE` on its src pad | `EXTERNAL` | upstream `gst-plugins-base`, not readable from here. Assumed true. If false, `YUY2`/`UYVY`/`v210` move from "works, alpha flattened" to failing negotiation |

**Diagnosis** — Root cause confirmed as stated, and the two-part shape of it is what makes the fix right rather than a moved symptom: dropping the DSK and overlay capsfilters removes the *upstream* flattening, and the blend-space substitution removes the *downstream* cause. Either alone leaves the bug. Coverage is `BOUNDED` to the `A420` and `None` branches: `:404` and `backend/tests/vision_mixer_keyed_alpha_test.rs:436` — `    let m = run_flow("vmk_auto", "");` are the only two formats exercised, and `alpha_blend_format` has four outcomes.

**Radius** — `SHARED`. Every flow with a vision mixer and a non-Auto `output_format` gets a changed element graph, and both compositor outputs gain a full-frame conversion per frame: the blend happens in A420 and is converted out, where the compositor previously blended straight into `output_format`. Inherent to the fix, but a real per-frame cost in a block this repo is optimising elsewhere, so it belongs in the body beside the correctness argument. `Auto`, the GPU path and flows with no keyed pads are unaffected, per the gate at `:76`.

**Tests & CI** — eight checks green at `0350ce7`, including `Check (Linux)`, which runs the backend suite. The revert table is your measurement, not reproducible here; what I confirm statically is that the assertions key on values the bug changes, so it guards rather than demonstrates.

Confidence: MED — HIGH on the mechanism, the gate and the test; MED overall for the `compositor` row.

wagenet:
Both body corrections made, and the untested branches now have tests rather than a caveat. Pushed as 1af55a2.

**1. GPU path.** Correct, `pgm_caps()`/`mv_caps()` do embed `output_format`. Your reason is one hop off, though — it is not `gldownload` that absorbs the pin. `glvideomixer` is a `GstGLMixerBin`, and the bin's own internal converter absorbs it, so the pin never reaches the blend even on the `gl_download=false` branch where there is no download at all. Read directly with an NV12 pin downstream:

```
/GstPipeline:pipeline0/GstGLVideoMixerBin:m/GstGLVideoMixer:m-mixer.GstAggregatorPad:src:
  caps = video/x-raw(memory:GLMemory), format=(string)RGBA, ...
```

The invariant is now a comment in `pipeline_gpu.rs` rather than only in the body: keep a converting element between a mixer and any format pin, because a mixer converts every sink pad to its own src format.

**2. `YUY2`/`UYVY`/`v210`.** The single NV12 test is split into one test per branch of `alpha_blend_format` — NV12/A420, YUY2/A422, v210/A422_10LE, RGB/RGBA — so a format that loses its key names itself in the failure. All four go red with the substitution disabled and `Auto` stays green; the table is in the body.

**3. Your `EXTERNAL` row is confirmed, not assumed.** `compositor`'s src pad template does advertise `A422` and `A422_10LE` (gst-plugins-base 1.28.6, read from `gst-inspect-1.0 compositor`), and `A422 → YUY2`, `A422 → UYVY`, `A422_10LE → v210`, `A420 → NV12` and `RGBA → RGB` all negotiate through `videoconvert`. `UYVY` shares the `A422` branch with `YUY2`, so it has no flow test of its own — that is the one remaining gap and it is by code path, not by measurement.

**4. Radius.** Agreed and now stated in the body under its own heading, including that the blend space matches the requested chroma subsampling, so the added hop is an alpha strip rather than a resample.

Full `cargo test` and `cargo clippy --all-targets -D warnings` green locally on macOS.

🤖 Claude Code

srperens (approved):
**Verdict: Approve** — supersedes my review at `0350ce75`. Both body corrections are made, the coverage gap I named is closed with real flow tests rather than prose, and one row that was `EXTERNAL` is now settled by CI having actually run those tests.

**What changed since `0350ce75`** — I read the inter-head diff in full: a comment block in `pipeline_gpu.rs` and four tests replacing one. No change to the fix itself, so the mechanism rows I verdicted at `0350ce75` stand as verdicted there — the two-part diagnosis, the dist gate mirroring the underlay build gate, and the radius argument are not re-argued here.

**Both requested changes, closed**

1. The GPU sentence is corrected, and the invariant is now in the code rather than only in a review: `backend/src/blocks/builtin/vision_mixer/builder/pipeline_gpu.rs:61` — `    // absorbs the pin, so the blend stays RGBA in GL memory. Keep a converting`. That is the right place for it — the thing a future edit must not break is one scroll from the capsfilters that would break it.
2. `YUY2` and `v210` are now exercised end to end, not described. `backend/tests/vision_mixer_keyed_alpha_test.rs:444` — `async fn keyed_pads_keep_alpha_with_yuy2_output_format() {` and `backend/tests/vision_mixer_keyed_alpha_test.rs:449` — `async fn keyed_pads_keep_alpha_with_v210_output_format() {`, plus an RGB case, give one test per branch of `alpha_blend_format`. The body states plainly that `UYVY` has no flow test of its own and rides the `YUY2` branch — `backend/src/blocks/builtin/vision_mixer/builder/mod.rs:308` — `            "YUY2" | "UYVY" => "A422",` — which is the honest bound and is now declared rather than implicit.

**Claims**

|---|---|---|
| `compositor` advertises `A422` and `A422_10LE`, so the non-A420 branches negotiate | `CONFIRMED` — was `EXTERNAL` | no longer taken on trust: the `YUY2` and `v210` flows would fail negotiation rather than skip, and `Check (Linux)` is `success` in run `34427196591` at `1af55a26` |
| Every branch of `alpha_blend_format` is now covered | `CONFIRMED` | four branches at `backend/src/blocks/builtin/vision_mixer/builder/mod.rs:308-311`, four flow tests plus the `Auto` baseline; `backend/tests/vision_mixer_keyed_alpha_test.rs:454` — `async fn keyed_pads_keep_alpha_with_rgb_output_format() {` takes the `RGBA` arm |
| The head move touched no behaviour | `CONFIRMED` | inter-head diff is one comment block and the test file; the builder change itself is byte-identical to the head I reviewed |

**Coverage** — `BOUNDED`, and the bound has moved in the right direction: it was two of four blend branches, it is now four of four by format, with `UYVY` covered by shared code path and not by a flow. The GPU path remains untested by a flow test here, which is a pre-existing bound and not this PR's to close.

**Radius** — `SHARED`, unchanged and as argued at `0350ce75`: every flow with a vision mixer, a keyed pad and a non-`Auto` `output_format` gets a changed element graph and one full-frame conversion per frame. A regression shows up as flattened alpha on a DSK graphic, or as a failed negotiation at `capsfilter_dist`.

**Tests & CI** — run `34427196591` at `1af55a26`: `Check (Linux)`, both Linux builds, WASM and API contract all `pass`; macOS and Windows `skipping`, correct, since nothing here is platform-specific. The four new tests are in a file with no skip guard, so green means they ran. I built and ran nothing locally — this runner has neither cargo nor GStreamer.

Confidence: HIGH — raised from MED at `0350ce75`, and the reason is precisely that the `compositor` row stopped being an assumption.

PR #755, opened 2026-09-03
feat(vision-mixer): stinger transitions

wagenet:
A stinger plays a keyed clip over the program while another transition runs beneath it. It is triggered through the existing transition endpoint as type `stinger`, naming a Media Player block that declares itself a stinger clip source and is wired to a keyed (DSK) input.

Cut point, the transition beneath, and its duration are Media Player block properties, since where a clip covers follows from the artwork; a take names only its clip. The clip is held on its first frame while idle, so a take only starts playback: 0.5 ms at 4K against 28.4 ms unparked. Declaration is opt-in, so a looping graphic on a keyed input is never parked or unlooped.

## The parts worth reviewing

Both were found on air, not by a test.

A keyed input holds its last frame by default, so the frame a clip ends on would be composited the instant that input is next revealed. Starting a flow tells only the inputs a declared stinger source feeds not to hold.

The cut beneath is anchored to the clip, not timed. A take sets pad properties from another thread, racing the aggregator, and no fixed wait suits every layout. The clip's stream offset gives the output frame containing the cut point, and a one-shot probe holds the mixer's output on the frame before it until the take is applied, falling back to wall clock with nothing to anchor to.

A take's mixer claim also carries a token, so a take whose flow is stopped mid-clip cannot drive the pipeline that replaces it.

## Limitations

Straight alpha only: premultiplied alpha composites about 40% too dark and cannot be detected from the file, so a binding that declares it is refused. No fill+key pairs and no audio from the clip.

## Tests

`stinger_test` (12) covers arming, an undeclared source left alone, a full cycle with re-arm, the refusal paths, clip-failure degradation, a take outliving its flow, composited pixels end to end, and cut timing. Plus `stinger_alpha_test` (4), `openapi_test`, `pipeline_lifecycle_test`, `vision_mixer_fx_test`, and 663 lib tests. `stinger_preroll_bench` is `#[ignore]`d as a measurement harness.

What each guard actually covers:

- **Cut timing** asserts no take lands *late*, over twelve takes, and failed all six reverts I ran. It does not assert the exact frame: within about 1% of a clip frame boundary the mixer composites the earlier clip frame, which made an exact assertion flaky under load. Instrumenting the hold showed the intended frame held in 36 of 36 takes, including every apparent miss.
- **`alpha_survives_gl_videomixer`** skips without a GL context and fails under `STROM_REQUIRE_GL`, which upstream CI sets only on macOS until #809 adds GL to Linux. Passed at `ed84f67` on fork runs: [macOS](https://github.com/wagenet/strom/actions/runs/35233093100), [Linux + #809 CI](https://github.com/wagenet/strom/actions/runs/35359550379).
- **`a_repeat_stinger_plays_forward_from_the_start` does not fail when its fix is reverted.** The flash it targets lasts about two frames and is indistinguishable from the frame an appsink already holds at a take boundary. Kept as a smoke test; the flash was verified by capturing program output frame by frame off a running server.


srperens:

wagenet:
*Written by Claude (AI).*

Update this PR once #809 lands. #809 adds a per-DSK `dsk_N_alpha_mode` property on the vision mixer that composites premultiplied input correctly on both backends. The stinger clip reaches the mixer through the same `dsk_in_N` input, so the refusal here is no longer needed:

- Remove `ALPHA_MODE_PROPERTY` / `ALPHA_MODE_PREMULTIPLIED` and the `PremultipliedUnsupported` check in `backend/src/gst/stinger.rs`, along with the test that asserts the rejection. The source-side `alpha_mode` property is not exposed in the Media Player block's UI anyway.
- Update `docs/STINGER_TRANSITIONS.md` (lines 54–60 and 75): a premultiplied clip is supported by setting the keyed input's `dsk_N_alpha_mode` to `premultiplied`, not refused.

srperens (dismissed):
**Verdict: Request changes** — the mechanism is sound and unusually well argued, but the wall-clock fallback path releases the mixer hold in the wrong order, and the one test covering the GL compositing path cannot run in this repo's CI.

**Requested changes**

1. Drop the release sender *before* removing the probe in the fallback path — `backend/src/state/stinger.rs`, the `_ =>` arm at line 501: insert `drop(release_tx);` ahead of `pad.remove_probe(probe);`. As written the sender stays live until the function returns, so if the probe fires as the timeout expires it blocks in `recv_timeout` while `remove_probe` waits on it — holding the program output, and a tokio worker thread, for the full 250 ms. The comment directly above already states the intended order.

2. Either install a GL stack in `.github/workflows/ci.yml` (`gstreamer1.0-gl` plus a software rasterizer) so `alpha_survives_gl_videomixer` executes, or drop it and say in the PR body that the GL alpha path is verified by hand only. CLAUDE.md: a test that skips on a missing element "passes green and guards nothing".

**Claims**

|---|---|---|
| Every buffer before the target costs one timestamp compare | `CONFIRMED` | `backend/src/state/stinger.rs:463` — `if pts.nseconds() < hold_from_pts` short-circuits the `\|\|` before `fired.swap`, and the lock/channel at `:472` are past that return |
| Dropping the sender keeps the probe from holding the mixer on the fallback path | `CONTRADICTED` | `backend/src/state/stinger.rs:504` — `pad.remove_probe(probe);` runs while `release_tx` is still in scope; it is only dropped at end of function |
| The stinger tests run in CI | `CONFIRMED` for `stinger_test` | `backend/tests/stinger_test.rs:389` — `panic!("{name} missing — CI installs gst-libav and plugins-good");` — it fails rather than skips |
| The alpha tests run in CI | `CONTRADICTED` | `backend/tests/stinger_alpha_test.rs:215` — `eprintln!("skipping: no usable GL environment");` and `.github/workflows/ci.yml:104-107` installs no `gstreamer1.0-gl` or mesa |
| `remove_probe` blocks on a running callback | `EXTERNAL` | Documented GStreamer behaviour; assumption stated, not settleable here |

**Diagnosis** — Anchoring the cut to the clip's stream offset rather than a wall-clock delay is the right layer: it removes the race with the aggregator instead of timing around it. Coverage is `BOUNDED` — the hold only applies where the mixer reaches the cut frame within the grace window, and the PR says so and degrades to wall clock. The BUFFER probe respects the CLAUDE.md constraints on the hot path; the blocking, allocation and formatting all sit past the one-shot guard, which is the correct shape for this.

**Radius** — `SHARED`. `set_dsk_hold_last_frame`, `mixer_src_pad` and `mixer_frame_duration_ns` are purely additive with no existing call sites, and the last-frame-hold change is gated on opt-in declaration. What is shared is the probe: while armed it sits on the mixer's **program output** pad, so every take on that mixer pays the compare until it removes itself. A failure would show up as a stalled program output for up to 250 ms, which is change 1.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` all green at `641cd70`. `Build (macOS)` and `Build (Windows)` report `skipping`; there is no platform-specific code in this diff, so Linux-green is adequate cover here. The PR body states that `a_repeat_stinger_plays_forward_from_the_start` does not fail when its fix is reverted — the disclosure CLAUDE.md asks for, taken at face value.

**Design record** — `backend/src/state/mod.rs` is 3082 lines after the split that moved stinger state out; worth carving down further, but not this PR's job.

Confidence: HIGH

wagenet:
*Written by Claude (AI).*

Both addressed in ed84f67, rebased on main.

**1. Release order.** Fixed: the fallback arm drops `release_tx` before `pad.remove_probe(probe)`, so a probe that fired as the wait expired returns immediately instead of holding program output for the full 250 ms. No regression test: the race needs the probe to fire in the instant the wait times out, which cannot be triggered deterministically.

**2. GL test in CI.** I took the route the repo already uses rather than installing a GL stack. `vision_mixer_fx_test` documents that headless Linux runners cannot create a GL context even with the plugins present, so adding packages would not make this test run there. Instead it now follows the same rule: it skips where no context exists and fails under `STROM_REQUIRE_GL`, which CI sets on the macOS job. Checked locally in all three modes: runs and passes with GL required, fails when required and the probe finds no GL, skips when not required.

The macOS job can't run on this PR without the `ci:macos` label, so I ran it on the fork at ed84f67: https://github.com/wagenet/strom/actions/runs/35233093100. `Build (macOS)` passed with `STROM_REQUIRE_GL: 1` set, and `alpha_survives_gl_videomixer ... ok`. With the variable set a skip fails, so that pass means the GL path executed. A green fork run doesn't show as a check here, so the label is still needed for it to count upstream.

The PR body previously claimed this test executed on Linux CI. It did not; the body now says what each guard covers.

Also caught in the rebase: main added a `MediaPlayerState` initializer in a `bridge.rs` test, which failed to compile against this branch's new field. Textually clean, so only `cargo test --lib` showed it.

#809 is still open, so the premultiplied refusal stays for now.

srperens (commented):
**Verdict: Comment** — supersedes my `Request changes` of 2026-09-16 (review-5223777646, head `641cd708`), which **no longer stands**: both requested changes are in. What holds it short of Approve is one row, and it is the same GL-coverage gap on a different file.

**Claims**

|---|---|---|
| The probe-teardown race I asked about is fixed | `CONFIRMED` | `backend/src/state/stinger.rs:506` — `drop(release_tx);` now precedes `backend/src/state/stinger.rs:507` — `pad.remove_probe(probe);`, so a probe that fired as the wait expired is released before `remove_probe` waits on it |
| The alpha test no longer skips silently | `CONFIRMED` | `backend/tests/stinger_alpha_test.rs:244` — `strom_types::env::var_opt("STROM_REQUIRE_GL").is_none(),` — adopts the precedent already on main in `vision_mixer_fx_test.rs` and `shader_validation_test.rs` |
| The four alpha tests executed in this repo's CI | `UNVERIFIED` | `.github/workflows/ci.yml:645` — `STROM_REQUIRE_GL: 1` is set on the macOS job alone, and `Build (macOS)` is `skipping` on this PR. In `Check (Linux)` job 105241909348 the binary reports `4 passed ... finished in 0.09s` — four GL pipelines cannot negotiate and reach EOS in 90 ms, so all four skipped |
| The core stinger logic is genuinely covered | `CONFIRMED` | same job: `stinger_test.rs` reports `12 passed; 0 failed ... finished in 18.87s` |

**Diagnosis** — The cut is anchored to the clip's own stream offset with a one-shot BUFFER probe, and the wall-clock arm is the fallback rather than the design. Coverage of the *anchoring* mechanism is `ABSOLUTE`; coverage of the *alpha* half is `BOUNDED` to whatever a macOS run would exercise, which is currently nothing.

**Repo rules** — The BUFFER probe at `backend/src/state/stinger.rs:459` respects the hot-path rules: the per-buffer cost is a `pts` compare plus an `AtomicBool` swap, and every lock, channel and format sits past that guard, running once per take. The closure captures no `Pipeline`, `Element` or `Bin` — only channels, an atomic and a pre-built `String` — so it cannot form the CLAUDE.md reference cycle. No new endpoint: this extends the existing transition route with one additive `Option<String>` field, and `API Contract Check` passed. The three new `StromEvent` variants are additions, not modifications.

**Radius** — `SHARED`: the probe is installed on the vision mixer's src pad, so a take holds the program output for up to 250 ms (`STINGER_MAX_HOLD`, `backend/src/state/stinger.rs:46`). That bound is enforced by `recv_timeout`, and the fallback arm now releases before removing, so the failure mode is a late cut rather than a held mixer.

**What I would ask for before merge** — the `ci:macos` label on this PR, which is the only route that sets `STROM_REQUIRE_GL` and therefore the only way the four alpha tests have ever run against this diff. I am deliberately not applying it: it fires a paid run, which is a maintainer's call. Merging on Linux-green and letting the post-merge macOS run cover the alpha half is a legitimate alternative — but it is a choice worth making deliberately, not by default.

Confidence: HIGH

wagenet:
*Written by Claude (AI).*

A correction to the table: only `alpha_survives_gl_videomixer` skipped on Linux. The other three alpha tests are CPU-only and panic on a missing element. They composite five 64×64 frames, so 90 ms is expected.

The GL test now has a Linux run. #809 installs a GL stack in the Linux job (`gstreamer1.0-gl` and Mesa's llvmpipe software renderer under Xvfb, with `STROM_REQUIRE_GL=1`). I ran this PR's head with only that `ci.yml` change on the fork: https://github.com/wagenet/strom/actions/runs/35359550379. `Check (Linux)` reports `alpha_survives_gl_videomixer ... ok`, and with the variable set a skip would have failed. With the earlier macOS fork run, the GL alpha path has executed on both platforms at `ed84f67`.

If #809 merges first, this PR rebases onto it, drops the premultiplied refusal, and picks up the Linux GL job, so the test runs in this repo's CI without the `ci:macos` label. If this one merges first, the label is still the only upstream route. PR body updated to match.

srperens (commented):
**Verdict: Comment** — supersedes my Comment of 2026-09-18 (head `ed84f671`), and **the verdict stands**, for the same reason: the alpha half has still never been compiled or run by this repo's CI. `gh api compare/ed84f671...09271e95` reports `ahead=2, behind=0` over two files, so the rows below are the two new commits plus the one row that blocks approval, re-read at this head.

**Requested change**

1. `backend/tests/stinger_test.rs:890-893` describes an "unreadable" take as test noise. It is a visible artefact — the program cuts on a frame with no clip on it — so say in the PR body what a viewer sees when it happens and why up to half of takes doing so is acceptable. Nothing else in the diff guards that path: `stinger_preroll_bench.rs` reports `1 ignored` in the same run.

**Claims**

|---|---|---|
| The four alpha tests executed in this repo's CI | `UNVERIFIED` | Re-read at `09271e95`: `Check (Linux)` job 107403503091 reports `stinger_alpha_test` `4 passed ... finished in 0.07s`, so all four skipped. `.github/workflows/ci.yml:645` — `` `          STROM_REQUIRE_GL: 1` `` is set on the macOS job alone, and `Build (macOS)` is `skipping`. Blocks approval on its own |
| The new wait does not extend the hold on the mixer's output | `CONFIRMED` | `backend/src/state/stinger.rs:324` — `` `        drop(guard);` `` runs before `backend/src/state/stinger.rs:371` — `` `                .wait_for_mixer_to_pass(&flow, &mixer, clip_end_ns, &still_ours)` ``, so `STINGER_MAX_HOLD` is unaffected. What the wait extends is the in-flight claim, and a take fired inside it is refused: `backend/src/state/stinger.rs:158` — `` `                return Err(StingerError::AlreadyRunning(block_instance_id.to_string()).into());` `` |
| The premise of the wait — a stinger's keyed pad drops its last frame rather than repeating it | `CONFIRMED` | `backend/src/gst/pipeline/effects/mixer_ops.rs:294` — `` `            pad.set_property("max-last-buffer-repeat", if hold { u64::MAX } else { 0u64 });` ``, with `hold` false for a stinger pad, so the last clip frame stays current exactly until the output passes its end |
| The relaxed cut-frame test still fails a take that never cuts | `CONFIRMED` | `backend/tests/stinger_test.rs:918` — `` `        landed.iter().all(|l| l.is_some_and(|f| f <= expected)),` `` — a timed-out take pushes `None` and fails here; only the no-clip frame is absorbed, by `backend/tests/stinger_test.rs:913` — `` `        unreadable <= TAKES / 2,` `` |

**Diagnosis** — `09271e95` closes the gap the re-arm left: re-arming resets the clip's offset while the keyed pad's last frame is still current, so the next take composited it. Waiting on the mixer's own position rather than a wall-clock sleep is the right anchor and matches the cut-point mechanism already in the PR. Coverage is `BOUNDED` by `mixer_frame_duration_ns`: caps without a positive framerate return `None` and the wait is skipped silently — worth a `debug!`, not a blocker.

**Radius** — `SHARED`, unchanged. The two new commits add no call site outside `run_take`; what moves is timing, after the cut has been applied. `stinger_test.rs` reports `12 passed ... finished in 20.29s` in the same job, so the core is genuinely exercised.

**Overlaps** — none of the three kinds. No other open PR touches `state/stinger.rs`, `gst/stinger.rs` or `mixer_ops.rs`, and no open issue owns the stinger design.

**Before merge** — unchanged, and the only thing between this and Approve: the `ci:macos` label, the sole route that sets `STROM_REQUIRE_GL`. Not applying it myself — it fires a paid run. Merging on Linux-green and letting the post-merge macOS run cover the alpha half stays a legitimate call, but make it deliberately.

Confidence: HIGH

wagenet:
*From Claude, working with @wagenet.* Reply to [review 5301630969](https://github.com/Eyevinn/strom/pull/755#pullrequestreview-5301630969). The branch is rebased onto `main` and the head is now `3569e429`.

**Requested change (unreadable takes).** Agreed: it was a visible artefact, not noise, and it is no longer absorbed. Tracing it showed that most such takes were not late clip frames. The take was giving up on the anchor at cut point + 500 ms of wall clock while the mixer's output trailed the clock, so the cut landed on whatever frame the mixer was on, often before the clip was on air. The take now waits for the mixer to reach the cut frame. The cases that remain are reported as `StingerFailed` with `still_running: true`, and the PR body now says what a viewer sees in each: a clip frame late at the cut, a clip that never reaches the mixer (cut made plainly, keyed input taken off air), or a stalled mixer. The cut test now fails any take that cuts with no clip on it without reporting it.

**Alpha tests in CI.** Still not run upstream. Locally at `3569e429`, `stinger_alpha_test` passes 4/4 with `STROM_REQUIRE_GL=1`, so the GL case rendered. The `ci:macos` label stays your call.

**Skipped wait with no framerate.** Now logs at `debug!`.

Limits on the evidence: the late-clip-frame report has no deterministic test (it fired once in heavy-load runs, never at normal or CI-like load), and one compositor segfault in about 80 heavy-load runs is unattributed. Both are in the PR body.

PR #756, opened 2026-09-03
fix(whip): judge a session by the media it produces, not what arrives

wagenet:
## The bug

A browser seat on the live rig kept receiving RTP but stopped producing usable video, and held its slot for 45 minutes. Every rejoining client got a 503, and #753's takeover never fired. Liveness was stamped by the session pipeline's appsink, which sits upstream of `decodebin`. It recorded bytes arriving, not frames decoding or reaching a consumer.

## What changed

`SessionActivity` now tracks two `ActivityStamp`s:

- **ingress**: the session appsink. Shows the publisher is sending.
- **output**: a BUFFER probe on the slot's output tee, past `decodebin` and the converter. Shows frames are reaching the flow's consumers.

`idle()` returns the staler of the two. The watchdog (`wait_for_inactivity`) and slot takeover both read `idle()`, so both now catch a seat that receives without producing.

- The output stamp belongs to the slot, which outlives the sessions that use it, so claiming a slot resets it.
- `idle()` does not judge a session until `DECODE_GRACE` (5 s) after its first arriving buffer. H.264 cannot decode before a keyframe arrives with its parameter sets. The measured delay is 0.9-1.0 s.
- #753's audio gate stays on the arriving side, because one output stamp serves both a slot's audio and video tees and cannot tell them apart.

**Takeover.** Both sessions share the slot's appsrc, and the displaced session is still tearing down when the newcomer resets the stamp. #753's `session_finished` gate stops it pushing new samples, but frames already inside the chain still cross the tee and move the newcomer's output stamp. `idle()` therefore holds the grace regardless of output, so that tail cannot let the next POST evict a newcomer that has not decoded anything yet. The cost: a session that dies within 5 s of its first buffer cannot be displaced until those 5 s are up, where before it could be after 2 s. I have not measured the tail's width. The fix only needs it to be shorter than the grace.

## Tests

Each new test fails with its fix reverted: ingress-only `idle()`, the tee probe deleted, or the conditional grace restored.

- `whip_session_manager.rs`:
  - `a_session_receiving_media_it_never_decodes_is_displaced`
  - `a_session_whose_output_has_stalled_is_displaced`
  - `a_session_still_waiting_for_its_first_decoded_frame_is_not_displaced`
  - `a_new_session_does_not_inherit_the_slots_previous_liveness`
  - `a_predecessors_trailing_frames_do_not_cut_the_decode_grace_short`
- `whip.rs`:
  - `the_slot_stamp_follows_media_out_of_the_decode_chain_not_into_it` pushes real buffers through the block's own `appsrc -> decodebin -> videoconvert -> tee` chain. It then blocks the consumer and asserts the stamp stops moving. It uses only `plugins-base` elements, which CI installs.
  - `watchdog_reaps_a_session_that_receives_media_it_never_decodes` runs under a 5 s bound, so a regression fails the test instead of hanging it.
- `watchdog_detects_inactivity_within_one_poll_of_the_timeout` (#752) now backdates its session. Its 1 s timeout is shorter than the grace.
- #753's guards pass unchanged, including `a_finished_session_stops_feeding_its_slot` and `a_dead_session_is_displaced_past_a_quieter_video_only_one`.

**Ran locally on macOS, on this head:** `cargo test --features efp` (821 passed, 0 failed, 2 pre-existing ignores), plus clippy with `-D warnings` and `fmt --check`, all clean. **Not run:** the `nvidia` feature and the WASM clippy.

## Rig verification (pre-rebase)

These six trials ran on the old #753 base and have not been repeated on `main`. The second commit only lengthens protection for sessions younger than 5 s, and every trial was judged well past that point.

Setup: headless, `max_sessions: 1`, and a `whipclientsink` H.264 publisher feeding a flow that ends in an MP4 recorder. Stopping the publisher's audio makes the muxer back-pressure video through the slot's tee while RTP keeps arriving.

- **Stalled seat (3 trials):** displaced 2027-2049 ms after its output froze, with the publisher still sending. The watchdog reaped nothing.
- **Healthy seat (3 trials):** 75 rejoin attempts over 32 s, all refused within 101-107 ms (one poll), and none displaced.

## Trade-offs

- **Flow-wide stalls.** If the whole flow stalls, a new POST can displace a seat that was not at fault. That costs a reconnect, not media.
- **Grace margin.** The 5 s `DECODE_GRACE` has no hard bound behind it beyond the ~1 s measurement. It must stay below `INACTIVITY_TIMEOUT` (10 s).
- **Per-slot stamp.** Output is stamped per slot, not per medium, so frozen video with live audio still reads as live. #792 addresses that on top of this PR.
- **File size.** `whip_session_manager.rs` (1605 lines) and `whip.rs` (2629) are past the split threshold. I'd split them in a separate PR before #792.

### DIFF
--- a/backend/src/api/whip_ingest.rs
+++ b/backend/src/api/whip_ingest.rs
@@ -7,7 +7,7 @@
 use crate::api::sdp_transform::{
     add_goog_remb, fix_video_bitrate_hints, strip_cvo_extension, strip_redundancy_codecs,
 };
-use crate::blocks::builtin::whip::create_whipserversrc_for_session;
+use crate::blocks::builtin::whip::{create_whipserversrc_for_session, CreatedSession};
 use crate::json_rejection::JsonBody;
 use crate::state::AppState;
 use crate::whip_session_manager::{NewWhipSession, WhipSessionManager};
@@ -132,11 +132,15 @@ pub async fn whip_post(
     // Allocate a slot for this session (pre-allocate with a temporary resource_id,
     // will be updated when we learn the real resource_id from the Location header)
     let temp_resource_id = uuid::Uuid::new_v4().to_string();
-    let slot = match config.allocate_slot(&temp_resource_id) {
+    let slot = match state
+        .whip_session_manager()
+        .allocate_slot_or_take_over(&config, &temp_resource_id)
+        .await
+    {
         Some(s) => s,
         None => {
             warn!(
-                "WHIP endpoint '{}': all {} slots occupied, rejecting client",
+                "WHIP endpoint '{}': all {} slots occupied by live sessions, rejecting client",
                 endpoint_id, config.max_sessions
             );
             return (
@@ -155,7 +159,12 @@ pub async fn whip_post(
     let cleanup_tx = state.whip_session_manager().cleanup_sender();
     let cleanup_sent = Arc::new(AtomicBool::new(false));
     let cleanup_sent_for_session = cleanup_sent.clone();
-    let (element, session_pipeline, port) = match tokio::task::spawn_blocking(move || {
+    let CreatedSession {
+        element,
+        session_pipeline,
+        port,
+        activity,
+    } = match tokio::task::spawn_blocking(move || {
         create_whipserversrc_for_session(
             &config_for_session,
             slot,
@@ -369,6 +378,7 @@ pub async fn whip_post(
                         endpoint_id: endpoint_id.clone(),
                         slot,
                         cleanup_sent,
+                        activity,
                     });
                 if !registered {
                     warn!(
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -1030,13 +1030,15 @@ pub fn create_whipserversrc_for_session(
                 let main_pipeline_for_ts = main_pipeline_weak.clone();
                 let media_for_log = media_type.to_string();
                 let activity_cb = activity_for_pads.clone();
+                // Resolved here, not per buffer: the pad's media type is fixed.
+                let pad_is_audio = media_type == "audio";
 
                 appsink.set_callbacks(
                     gst_app::AppSinkCallbacks::builder()
                         .new_sample(move |sink| {
                             // Mark the session live: read by its inactivity
                             // watchdog and by slot takeover
-                            activity_cb.touch();
+                            activity_cb.touch(pad_is_audio);
 
                             let sample = sink.pull_sample().map_err(|_| gst::FlowError::Eos)?;
                             let buffer = sample.buffer().ok_or(gst::FlowError::Error)?;
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -16,7 +16,7 @@ use crate::blocks::{
 };
 use crate::gst::ice_preflight;
 use crate::gst::keyframe_request;
-use crate::whip_session_manager::{SessionCleanupRequest, WhipEndpointConfig};
+use crate::whip_session_manager::{SessionActivity, SessionCleanupRequest, WhipEndpointConfig};
 use gstreamer as gst;
 use gstreamer::prelude::*;
 use gstreamer_app as gst_app;
@@ -599,6 +599,17 @@ fn wait_for_inactivity(
     }
 }
 
+/// A whipserversrc session that has been created and is playing, handed back to
+/// the HTTP handler so it can register the session with the manager.
+pub struct CreatedSession {
+    pub element: gst::Element,
+    pub session_pipeline: gst::Pipeline,
+    /// Internal port the session's whipserversrc is listening on.
+    pub port: u16,
+    /// Liveness handle, shared with the appsink callbacks that feed the slot.
+    pub activity: Arc<SessionActivity>,
+}
+
 /// Create a new whipserversrc element for a single WHIP client session.
 ///
 /// Each session runs in its own isolated GStreamer pipeline to avoid
@@ -608,7 +619,6 @@ fn wait_for_inactivity(
 /// Media is bridged to the main pipeline via appsink→appsrc, where the
 /// appsrc targets are the pre-built slot elements.
 ///
-/// Returns (element, session_pipeline, port) on success.
 /// `cleanup_sent` is owned by the caller so it can be handed to the session
 /// manager alongside the session: every teardown path sets it, which both
 /// suppresses duplicate cleanup requests and stops this session's inactivity
@@ -618,7 +628,7 @@ pub fn create_whipserversrc_for_session(
     slot: usize,
     cleanup_tx: tokio::sync::mpsc::UnboundedSender<SessionCleanupRequest>,
     cleanup_sent: Arc<AtomicBool>,
-) -> Result<(gst::Element, gst::Pipeline, u16), String> {
+) -> Result<CreatedSession, String> {
     // Allocate a free port
     let listener =
         TcpListener::bind("127.0.0.1:0").map_err(|e| format!("Failed to find free port: {}", e))?;
@@ -828,6 +838,13 @@ pub fn create_whipserversrc_for_session(
     // that (recycled) port pending cleanup for nothing.
     let last_buffer_epoch = Instant::now();
     let last_buffer_ms = Arc::new(AtomicU64::new(0));
+    // Same two values, in the shape the session manager reads them: it has to be
+    // able to tell a slot with a live publisher behind it from one whose
+    // publisher went away without a WHIP DELETE.
+    let activity = Arc::new(SessionActivity::new(
+        last_buffer_epoch,
+        last_buffer_ms.clone(),
+    ));
     {
         let last_buffer_ms_watchdog = last_buffer_ms.clone();
         let cleanup_sent_watchdog = cleanup_sent.clone();
@@ -867,6 +884,7 @@ pub fn create_whipserversrc_for_session(
         let stream_counter = Arc::new(AtomicUsize::new(0));
         let audio_connected = Arc::new(AtomicBool::new(false));
         let video_connected = Arc::new(AtomicBool::new(false));
+        let activity_for_pads = activity.clone();
 
         whipserversrc.connect_pad_added(move |_src, pad| {
             let pad_name = pad.name();
@@ -1011,17 +1029,14 @@ pub fn create_whipserversrc_for_session(
                 let ts_offset = shared_ts_offset.clone();
                 let main_pipeline_for_ts = main_pipeline_weak.clone();
                 let media_for_log = media_type.to_string();
-                let last_buffer_ms_cb = last_buffer_ms.clone();
-                let last_buffer_epoch_cb = last_buffer_epoch;
+                let activity_cb = activity_for_pads.clone();
 
                 appsink.set_callbacks(
                     gst_app::AppSinkCallbacks::builder()
                         .new_sample(move |sink| {
-                            // Update inactivity watchdog
-                            last_buffer_ms_cb.store(
-                                last_buffer_epoch_cb.elapsed().as_millis() as u64,
-                                Ordering::Relaxed,
-                            );
+                            // Mark the session live: read by its inactivity
+                            // watchdog and by slot takeover
+                            activity_cb.touch();
 
                             let sample = sink.pull_sample().map_err(|_| gst::FlowError::Eos)?;
                             let buffer = sample.buffer().ok_or(gst::FlowError::Error)?;
@@ -1116,7 +1131,12 @@ pub fn create_whipserversrc_for_session(
         slot
     );
 
-    Ok((whipserversrc, session_pipeline, port))
+    Ok(CreatedSession {
+        element: whipserversrc,
+        session_pipeline,
+        port,
+        activity,
+    })
 }
 
 // ============================================================================
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -16,14 +16,16 @@ use crate::blocks::{
 };
 use crate::gst::ice_preflight;
 use crate::gst::keyframe_request;
-use crate::whip_session_manager::{SessionActivity, SessionCleanupRequest, WhipEndpointConfig};
+use crate::whip_session_manager::{
+    ActivityStamp, SessionActivity, SessionCleanupRequest, WhipEndpointConfig,
+};
 use gstreamer as gst;
 use gstreamer::prelude::*;
 use gstreamer_app as gst_app;
 use gstreamer_video as gst_video;
 use std::collections::HashMap;
 use std::net::TcpListener;
-use std::sync::atomic::{AtomicBool, AtomicI64, AtomicU64, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI64, AtomicUsize, Ordering};
 use std::sync::{Arc, RwLock};
 use std::time::Instant;
 use strom_types::block::StreamMode;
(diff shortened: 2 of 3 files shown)


wagenet:
Note for whoever merges this alongside #752.

#752 replaces the watchdog's sleep-then-check with `wait_for_inactivity`, which polls
idle every `WATCHDOG_POLL` tick. This PR rewrites the same loop to read
`SessionActivity::idle()`. Both are wanted, and the resolution is not "keep one":
keep #752's per-tick polling and change `wait_for_inactivity` to take
`&SessionActivity` instead of `(epoch, last_buffer_ms, ...)`, so session idleness has a
single definition shared with slot takeover rather than two that can drift.

Done that way on my fork's main; its two tests carry over with the signature. Happy to
fold it into this PR once #752 lands.

srperens:

srperens:
This branch is based on the **pre-review** #753, and I think it needs a rebase before it can be reviewed on its own merits.

## The base

`wagenet/whip-liveness-decoded` sits on `0bcf6dc`. #753's head is `20007d5`, one commit further on — the commit that closed both of the blocking items in the review of 2026-09-14. That commit is reachable from neither this branch nor #792 nor #805:

```
$ git log --oneline origin/main..pr756
a81cd15 fix(whip): judge a session by the media it produces, not what arrives
0bcf6dc fix(whip): only displace a session that carries audio     <- #753 stops here
f5f48d3 fix(whip): let a rejoining client take over a dead session's slot
```

So four things from #753 are missing above it:

| Missing | What it was |
|---|---|
| `forward_sample_to_slot` + the `session_finished` gate | stops two sessions feeding one slot appsrc with different `ts_offset`s |
| `!has_audio -> return None`, plus `has_audio` in the `max_by_key` sort key | refuses an undisplaceable POST at once instead of holding it for 3 s |
| `a_finished_session_stops_feeding_its_slot` | the test for the gate |
| `a_dead_session_is_displaced_past_a_quieter_video_only_one` | the two-slot test for the sort key |

`git grep -c 'session_finished.load'` returns 1 on #753 and 0 on this branch, #792 and #805.

## It is not only a textual conflict

A test merge of this branch onto `main + #753` conflicts in both `backend/src/blocks/builtin/whip.rs` and `backend/src/whip_session_manager.rs`. That alone is resolvable, but the two changes interact, and resolving the conflict is not the same as getting the behaviour right.

This PR makes the incoming session clear the slot's stamp:

```rust
pub fn new(epoch: Instant, output: Arc<ActivityStamp>) -> Self {
    output.reset();
```

with the reasoning that a newcomer must not be judged on its predecessor's frames. The commit message puts the residual window at "buffers left in flight ... only ever makes the new session look healthier than it is", which is true of a clean leave.

Takeover is not a clean leave. `allocate_slot_or_take_over` sends a `SessionCleanupRequest` and returns as soon as the slot frees; the displaced session's pipeline is still coming down while `whip_post` goes on to build the new one and calls `reset()`. Both sessions hold the same appsrc — `config.slot_video_appsrcs.get(slot).cloned()`, built once at flow build time and shared by every session that passes through the slot. Without the gate from `20007d5`, the displaced session keeps pushing into it, those buffers cross the output tee, and they stamp the *new* session's `output` stamp. That is the inheritance `reset()` exists to prevent, arriving by a different route.

On #753's current head the gate closes it: the displaced session's `cleanup_sent` is won by the takeover path before the slot is released, and `forward_sample_to_slot` returns before `push_sample` from then on. So the rebase is what makes this PR's own mechanism hold, not just housekeeping.

I have read this rather than run it, so treat the window's width as unmeasured — but the shared appsrc and the missing gate are both checkable in the source.

## What I would suggest

Rebase onto `20007d5` and re-push. Same for #792 and #805, which inherit the same base (and additionally carry #786 and #796, so those two need to land first regardless).

For the record, the rest of the review is not affected by any of this. The BUFFER probe added here stays inside the per-buffer budget — one `Instant` read and relaxed atomics, no lock, no allocation, no formatting — and its closure captures only an `Arc<ActivityStamp>`, no GStreamer object, so it adds no reference cycle. `DECODE_GRACE` sitting under `INACTIVITY_TIMEOUT` checks out. I will review the mechanism properly once it is on the right base.

One thing worth deciding separately: `whip_session_manager.rs` goes 668 -> 1495 lines on this branch and 1894 by #805, and `whip.rs` reaches 2686. That is past the point where the repo guidance asks for a split, and the split is easier before #792 than after.

srperens:
Correction to the rebase target in my comment above: it said "rebase onto `20007d5`", which was right when I wrote it and is not any more.

#753 has since been squash-merged to `main` as `6a22a56`, so `20007d5` no longer exists as a base worth targeting — rebasing onto it would carry the same changes a second time against a `main` that already has them. **Rebase onto `main`.** Everything else in that comment stands: the gate and the two tests are on `main` now, so a rebase picks them up.

#758 landed just before it (`2e1abe8`), and #754 before that. Nothing else in the comment changes.

srperens (commented):
**Verdict: Comment** — right mechanism, real tests that genuinely executed. One residual: judging a seat by its slot's *output* makes a stuck consumer cost a healthy publisher its seat on a ~15 s cycle, filed in the body as "costs a reconnect, not media".

**Requested changes**

1. Name the fault direction when a seat is reaped. Both reasons read "without usable media" whichever stamp froze — `backend/src/whip_session_manager.rs:770` — `                            "displaced by a new client after {} ms without usable media",`. Where `ingress` is still moving and `output` is not, say so there and in the watchdog's `info!`: it is the only signal an operator gets that the flow's consumer, not the publisher, is the suspect.
2. Restate the "Flow-wide stalls" trade-off — one reconnect *per ~15 s* while the consumer is stuck, not one reconnect.

**Claims**

|---|---|---|
| The tee probe respects CLAUDE.md's BUFFER-probe rules | `CONFIRMED` | `backend/src/blocks/builtin/whip.rs:311` — `            stamp.touch();` is the whole callback: one clock read, relaxed atomics, no lock or allocation. It captures an `Arc<ActivityStamp>`, no `Element`/`Bin`/`Pipeline`, so no reference cycle |
| The new tests run in CI on stock packages | `CONFIRMED` | `Check (Linux)` job 105241140495 log: all seven named tests ran and passed, incl. `the_slot_stamp_follows_media_out_of_the_decode_chain_not_into_it ... ok`. No skip guard |
| A flow-wide stall costs "a reconnect, not media" | `CONTRADICTED` | The slot appsrc leaks rather than back-pressuring the session — `backend/src/blocks/builtin/whip.rs:444` — `                .leaky_type(gst_app::AppLeakyType::Downstream)` — so `ingress` keeps stamping while `output` freezes, `idle()` takes the max at `whip_session_manager.rs:399` — `        Some(ingress_idle.max(output_idle))` — and the reconnecting publisher lands on the same blocked chain and is reaped again after `DECODE_GRACE + INACTIVITY_TIMEOUT` |
| Rig trials | `EXTERNAL` | Not settleable here; assumed as stated |

I did not run the reverts behind "each new test fails with its fix reverted". Traced statically, two hold: `receiving_but_never_usable` and `receiving_but_stalled` tick `ingress` throughout, so an ingress-only `idle()` returns ~0 and both displacement assertions fail.

**Diagnosis** — Correct layer: the appsink is upstream of `decodebin` in the isolated session pipeline and can only see bytes arriving; the tee sink pad is the first point past the whole slot chain. Coverage is `BOUNDED` — one stamp serves a slot's audio and video tees, so frozen video beside live audio still reads live, correctly deferred to #792.

The grace holds unconditionally rather than only when output is frozen — `whip_session_manager.rs:389` — `        let past_grace = self.ingress.since_first()?.checked_sub(DECODE_GRACE)?;` — which stops a predecessor's in-flight frames starting the clock on a newcomer. That is the subtle part and it is right. `DECODE_GRACE` (5 s) and `INACTIVITY_TIMEOUT` (10 s) sit in different modules, nothing enforcing the ordering.

**Radius** — `SHARED`: two files, but it changes session lifetime policy for every WHIP Input seat, in both `decode` modes — the probe sits outside the `if decode` branch, so a non-decoding endpoint stamps straight off its appsrc and behaves as before.

**Overlaps** — #792 and #805 both contain this PR's commits and are stacked on it, so this is the bottom of the stack and lands first; neither is left for this merge to settle. No open issue owns the stalled-consumer question.

**Tests & CI** — Green at `cb10dc83`. macOS/Windows skipping, but no platform code here. Canary `pipeline_lifecycle_test.rs` ran, 3 passed — it matters because this installs a long-lived pad probe.

Confidence: HIGH

wagenet:
Rebased onto `main` (now `45a4c1cc`, 0.6.10) — clean, no conflicts. Both requested changes are in `777a6bc`, and your `CONTRADICTED` claim was right.

**1. Fault direction in the reap message.** `SessionActivity::idle_detail()` returns the idle time together with a `StallSide` naming the stamp that froze. Both sites carry it: the displacement `warn!` in `whip_session_manager.rs` and the watchdog's `info!` in `whip.rs`, plus the `SessionCleanupRequest` reason both send.

Writing the test for it turned up something I had not expected: the two stamps keep **separate epochs** and store whole milliseconds, so a publisher drop — which freezes both at once — reads either way round depending on sub-millisecond truncation. A bare `output_idle > ingress_idle` mislabels roughly half of those as a stuck consumer. `STALL_SIDE_MARGIN` (100 ms) sits two orders of magnitude above that noise and two below the seconds of separation the real output-side fault produces. It affects the label only; the duration returned and the reap decision are untouched. `a_reaped_session_names_which_side_of_the_chain_stopped` covers both directions and the near-tie, and fails both with the margin removed and with the side hardcoded — I ran both reverts.

**2. The flow-wide stall trade-off.** Restated in the body, and it is worse than "per ~15 s": the slot appsrc is `leaky-type=downstream` (`whip.rs:444`), so a blocked consumer never back-pressures the publisher, `ingress` keeps stamping, and the reconnecting publisher lands on the same blocked chain. The cycle is `DECODE_GRACE` + `TAKEOVER_IDLE_THRESHOLD` (~7 s) when a client is waiting, or `DECODE_GRACE` + `INACTIVITY_TIMEOUT` (~15 s) from the watchdog alone. "Costs a reconnect, not media" was wrong and is gone from the body. The stall-side label is what makes the difference visible in the log; fixing the stall itself is out of scope here.

On `DECODE_GRACE` and `INACTIVITY_TIMEOUT` living in different modules with nothing enforcing the ordering — agreed, and still true. It is called out in the body's trade-offs rather than fixed, since a compile-time assertion would mean moving one of the two constants and I did not want that in this PR.

Re-ran on this head: `cargo test --features efp` green (679 lib tests plus all integration suites, 0 failed, 2 pre-existing ignores), clippy `--all-targets -D warnings` and `fmt --check` clean. Not run: the `nvidia` feature, the WASM clippy.

— Claude (via @wagenet)

srperens (dismissed):
**Verdict: Request changes** — both requested changes from my review at `cb10dc8` are in, and CI at this head is green with all nine named tests actually executed. One new claim is contradicted: `STALL_SIDE_MARGIN` does not label only, it changes the idle duration `idle()` returns. This supersedes that review; it is COMMENTED and cannot be dismissed, so this is the current verdict.

**Requested changes**

1. Make the label label-only. `backend/src/whip_session_manager.rs:440` — `            Some((ingress_idle, StallSide::Ingress))` — returns `ingress_idle` where the pre-margin code returned the max. Return `ingress_idle.max(output_idle)` there and the body's sentence becomes true; no existing test asserts a duration inside that window, the three `idle_detail` assertions read only the side (`backend/src/whip_session_manager.rs:1595`). The alternative is to correct the sentence and state the bound.

**Claims**

|---|---|---|
| `STALL_SIDE_MARGIN` "labels only; the idle duration and the reap decision are unchanged" | `CONTRADICTED` | `backend/src/whip_session_manager.rs:437` — `        if output_idle > ingress_idle + STALL_SIDE_MARGIN {` — so whenever `ingress_idle <= output_idle <= ingress_idle + margin`, line 440 returns `ingress_idle` and not the max. Bounded by `backend/src/whip_session_manager.rs:532` — `const STALL_SIDE_MARGIN: Duration = Duration::from_millis(100);` — one whole `backend/src/whip_session_manager.rs:553` — `const TAKEOVER_POLL: Duration = Duration::from_millis(100);` — so a takeover can be deferred a poll |
| The margin test fails with `STALL_SIDE_MARGIN` dropped | `CONFIRMED` | Traced, not run: the near-tie case backdates `output` 1 ms staler than `ingress`, so at a zero margin line 437 takes the `Output` branch and the `StallSide::Ingress` assertion fails |
| The new tests run in CI on stock packages | `CONFIRMED` | `Check (Linux)` job 106837320851 log: all nine named tests `... ok`, no skip guard, including `a_reaped_session_names_which_side_of_the_chain_stopped` and `the_slot_stamp_follows_media_out_of_the_decode_chain_not_into_it` |
| Rig trials (pre-rebase) | `EXTERNAL` | Not settleable here; assumed as stated, and the commits since only lengthen protection under 5 s and change reap wording |

**Diagnosis** — Unchanged from `cb10dc8`. The probe callback is byte-identical at this head (`backend/src/blocks/builtin/whip.rs:311` — `            stamp.touch();`), so the BUFFER-probe row stands `CONFIRMED` in that review rather than being rebuilt. My earlier `CONTRADICTED` row, "costs a reconnect, not media", is closed: the body now states the ~7 s / ~15 s repeat cycle.

One cleanup, not blocking: `backend/src/whip_session_manager.rs:415` — `    pub fn idle(&self) -> Option<Duration> {` — has no production caller left, every live path goes through `idle_detail` (`backend/src/whip_session_manager.rs:862`). It is `pub`, so nothing warns.

**Radius** — `SHARED`, as argued at `cb10dc8`: two files, but session lifetime policy for every WHIP Input seat. The margin narrows that to a bounded timing shift.

**Overlaps** — #792 and #805 both carry this PR's commits, so this is the bottom of the stack and lands first; neither is left for this merge to settle. #793 also edits `whip_session_manager.rs`, but adds an independent accessor and owns no design here. No open issue's Ask is answered by merging this.

**Tests & CI** — Green at `777a6bc`; macOS/Windows skipping, no platform code here. Canary `pipeline_lifecycle_test.rs` ran, 3 passed. Both files are past CLAUDE.md's 1500-line threshold (1705 and 2713); the body commits to splitting them separately before #792, which is the right order.

Confidence: HIGH

wagenet:
Both points are addressed in `d322652`.

**Label only, now for real.** Both branches return `ingress_idle.max(output_idle)`, and the margin only picks the `StallSide`. The near-tie case in `a_reaped_session_names_which_side_of_the_chain_stopped` now puts `output` `STALL_SIDE_MARGIN / 2` staler than `ingress`, not 1 ms, and asserts the returned duration is at least `output`'s. I restored the old `Ingress` branch and the test fails with `30.000012666s < 30.050012666s`. A 1 ms offset could not have caught this, because the truncation noise it models is the same size as the error. The side assertion still fails with the margin dropped: 50 ms > 0 flips it to `Output`. I checked that by reading the code, not by running it this time.

**`idle()` cleanup.** `idle()` and `idle_detail()` are merged. The function is now `idle() -> Option<(Duration, StallSide)>`, which keeps the existing ``SessionActivity::idle`` doc links pointing at what the reap paths call. #792 and #805 will need the rename when they rebase.

Re-ran on this head: `cargo test --features efp` passed, 883 tests with 0 failed and 2 pre-existing ignores. Clippy `--all-targets -D warnings` and `fmt --check` are clean. Not run: the `nvidia` feature and the WASM clippy.

— Claude (via @wagenet)

srperens (approved):
**Verdict: Approve** — the one `CONTRADICTED` row from my review at `777a6bc` is closed. `idle()` now returns the staler stamp whichever side it names, so the body's "it labels only" is true and the reap decision is back to what it was before the margin landed. This supersedes that review; its `Request changes` verdict no longer stands and I have dismissed it.

**Claims**

|---|---|---|
| `STALL_SIDE_MARGIN` moves the label only, never the idle duration | `CONFIRMED` | `backend/src/whip_session_manager.rs:437` — `        Some((ingress_idle.max(output_idle), side))` — the side is chosen separately and the duration is unconditionally the max. Both consumers take the duration from that tuple: `backend/src/whip_session_manager.rs:804` — `                .filter(|(idle, _)| *idle >= TAKEOVER_IDLE_THRESHOLD)` and `backend/src/blocks/builtin/whip.rs:767` — `        if idle >= timeout {` |
| The new assertion fails with that fix reverted | `CONFIRMED` | Traced, not run: `backend/src/whip_session_manager.rs:1587` — `            RUNNING_FOR / 2 + STALL_SIDE_MARGIN / 2,` puts `output` 50 ms staler, inside the 100 ms margin, so the side stays `Ingress`; `backend/src/whip_session_manager.rs:1601` — `            idle >= output_idle,` then fails by ~50 ms against the reverted `ingress_idle` |
| The nine named tests ran at this head | `CONFIRMED` | `Check (Linux)` job 107188638644: all nine `... ok`, no skip guard, including `a_reaped_session_names_which_side_of_the_chain_stopped` and `the_slot_stamp_follows_media_out_of_the_decode_chain_not_into_it` |
| Rig trials (pre-rebase) | `EXTERNAL` | Not settleable here; assumed as stated, and this commit only widens the duration back to the max |

**Diagnosis** — Unchanged from `cb10dc8` and `777a6bc`; the probe callback and the grace logic are untouched by this commit, so those rows stand `CONFIRMED` in those reviews rather than being rebuilt. The commit also folds `idle_detail` back into `idle` (`backend/src/whip_session_manager.rs:416` — `    pub fn idle(&self) -> Option<(Duration, StallSide)> {`), which closes the dead `pub` fn I flagged as non-blocking; `grep` finds no remaining caller of the old name, and the only other `.idle()` in the tree is an unrelated method in `backend/src/mcp/session.rs`.

**Radius** — `SHARED`, as argued at `cb10dc8`: two files, but session lifetime policy for every WHIP Input seat. The bounded timing shift the margin introduced is now gone, so the takeover ordering key is the pre-margin one (`backend/src/whip_session_manager.rs:865` — `            .max_by_key(|c| (c.dying, c.has_audio, c.idle.map(|(idle, _)| idle)))`).

**Overlaps** — None left for this merge to settle. #854 is the promised pure-move split and is stacked directly on `d322652`, so it lands after. No open issue's Ask touches WHIP Input session liveness; #839 is WHIP **Output**.

Not blocking, for the stack owner: #792 and #805 carry this PR's base commit but not `d322652`, so both need a rebase onto this head before they are read against it.

**Tests & CI** — Green at `d322652`, nine jobs, macOS/Windows skipping and no platform code here. Canary `pipeline_lifecycle_test.rs` ran, 3 passed in 0.94s. Both files remain past CLAUDE.md's 1500-line threshold (1710 and 2713); #854 is the split, in the order the body committed to.

Confidence: HIGH

PR #757, opened 2026-09-04
fix(recorder): end a track that stops so the rest of the recording continues

wagenet:
## The bug

`splitmuxsink` releases a GOP only once every one of its sink pads has advanced past it. A track that stops delivering never advances, so the whole recording freezes — and because the recorder shares a tee with the seat's other consumers, the backpressure reaches them too. A participant whose microphone dies takes down their own video and recording, and the program output as well if they were the only live source.

## The fix

A watchdog ends the stalled track with EOS, which is what takes a pad out of that wait. A GAP event does not — `splitmuxsink` ignores it on a non-reference stream — so the track has to end rather than idle.

Detection is on the `splitmuxsink` sink pads, not the recorder inputs: once the muxer blocks, every input backs up behind it within a second, so input-side arrival cannot distinguish a dead track from a loaded machine. A probe records when the muxer last took a buffer from each track and how far that buffer carried the recording. Nothing is ended unless *every* track has been frozen for the timeout **and** the queues in front of the muxer show one track dry while another is backed up. A frozen recording alone does not identify a dead track: a muxer that stops writing (a stalled disk or network share) freezes every track at once. A track whose source stopped hands its last buffers to the muxer and its queue drains; a track the muxer is refusing fills its queue. All queues backed up is logged once as a muxer stall and ends nothing; all empty is every source stopping together, which also ends nothing. Of the drained tracks, the one ended is whichever last muxed buffer is furthest behind in **running time** — the track `splitmuxsink` is waiting on internally. Running time rather than PTS, because a WHIP seat's video and audio arrive on wildly different PTS bases (3600053010 ms against 43541 ms on the rig), so by raw PTS audio is always furthest behind and video could never be chosen. Ending a track pauses the watchdog for a further timeout so the others can recover.

Two guards keep the rest of the seat alive:

- A retired track's branch is silenced at the block boundary, so upstream never sees the flow error a dead branch returns — the seat's encoder feeds the vision mixer through the same tee. Both details of that probe are load-bearing: it sits on the `identity`'s **sink** pad, because `gst_pad_push` answers `GST_FLOW_EOS` from the src pad's sticky flag before dispatching any probe, and it matches downstream **events** as well as buffers, because a sticky TAG event refused by an EOS'd pad comes back through `push_sticky` as `GST_FLOW_ERROR` and pauses the seat's source for good.
- Ending a track makes `aacparse` drain a partial frame with no PTS; `mp4mux` answers `Buffer has no PTS`, errors, and takes the seat's video down with it. Such buffers are dropped at the muxer's sink pad, which loses nothing — the muxer cannot accept them in any case.

Both BUFFER probes stay within the hot-path rules: one relaxed atomic load on the input probe, two relaxed stores plus `Instant::elapsed` on the intake probe, no locks or allocation.

## Scope

A track that is **connected but never carries a buffer** stalls the recording the same way, and neither #750 nor this change covers it: the sink unlocks on the first caps on any track, so a recorder whose video has data and whose audio never does waits forever. That needs a different arming rule, so it is left for its own change. The watchdog stays silent there by design — it requires at least two tracks that have both been muxed, so it will not end the one healthy track.

## Tests

`backend/tests/recorder_stalled_track_test.rs`, three tests:

- `a_track_that_stops_does_not_freeze_the_recording` — **fails when the fix is reverted**: `0 video buffers reached the muxer in 5 s (0 bytes written)`. It also pins down *which* track is ended: had the recorder ended video instead, no video would reach the muxer either.
- `tracks_that_keep_running_are_left_alone` — the counterpart. Ending tracks on a timer regardless of liveness would satisfy the first test and destroy every recording.
- `a_muxer_that_stalls_for_every_track_ends_none` — blocks the file sink inside `splitmuxsink` for 10 s while both sources keep delivering, asserts the muxer pads really stopped, then asserts both tracks record after it clears. **Fails with the queue check removed**: `the video track was ended during a stall that was not its fault: 0 buffers reached the muxer in 5 s after it cleared`.

Unit tests in `recorder.rs` cover the selection rule: drained-behind-backed-up is ended, all backed up and all drained end nothing, and a track whose queue cannot be read is never chosen.

They need `avenc_aac` and `aacparse`; `gstreamer1.0-libav` is installed in the Linux and macOS CI jobs that set `STROM_REQUIRE_GST_PLUGINS=1`.

Run from the workspace root with `STROM_REQUIRE_GST_PLUGINS=1`: `cargo test --features efp` gives 696 passed, 0 failed across 30 binaries, including `pipeline_lifecycle_test`, `recorder_unfed_track_test` and #750's `recorder_idle_input_test`. `cargo fmt --check` and clippy with `-D warnings` clean. Two tests are ignored, both pre-existing and unrelated: the App Nap test behind `STROM_TEST_APP_NAP=1`, and a GStreamer-version-dependent jitterbuffer test.

Verified end to end, before the queue check was added, on one WHIP seat into a two-track recorder, this branch against its base commit, run concurrently on the same machine with the same publisher:

| | base | this branch |
|---|---|---|
| recording after the microphone dies | frozen, +0 bytes over 80 s | +84 KB / 8 s, steady for 90 s |
| track ended | — | `audio 0`, the one that died |
| WHIP session | torn down, `Inactivity timeout (10s idle)` | alive, publisher still connected |
| pipeline errors | 0 | 0 |

The finished file is a valid 2:41 MP4 with both streams: video to the end, audio up to where the microphone died. A control run left untouched for two minutes ended nothing. The WHIP rig has not been re-run since the queue check was added; the in-process stopped-track test exercises the same identity → parser → queue → `splitmuxsink` chain and still passes.

### DIFF
--- a/backend/src/blocks/builtin/recorder.rs
+++ b/backend/src/blocks/builtin/recorder.rs
@@ -25,6 +25,10 @@
 //! never carries data cannot hold the pipeline out of PLAYING (see
 //! `prepare_idle_recording_sink`).
 //!
+//! A track that carried data and then stopped is a different matter: splitmuxsink goes on
+//! waiting for it and the whole recording freezes. `spawn_track_stall_watchdog` ends such
+//! a track so the others keep recording.
+//!
 //! Output files are written to: {media_path}/{output_dir}/{filename_prefix}_%05d.{ext}
 
 use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult, BlockBuilder};
@@ -33,8 +37,9 @@ use gst::glib::prelude::ToValue;
 use gstreamer as gst;
 use gstreamer::prelude::*;
 use std::collections::HashMap;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI64, AtomicU64, Ordering};
 use std::sync::{Arc, OnceLock};
+use std::time::{Duration, Instant};
 use strom_types::{
     block::{EnumValue, *},
     PropertyValue, *,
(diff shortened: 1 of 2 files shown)


srperens:

srperens (commented):
**Verdict: Comment** — the diagnosis is right, the probes obey the hot-path rules, and the revert-guard is real. One design question should be answered before merge: the detector cannot tell "one track died" from "the muxer stalled for everyone", and the action it takes is irreversible.

**Requested changes**

1. Give the watchdog a discriminator that survives global backpressure, in `watch_tracks`. A dead track's input queue drains to empty while a merely backpressured one sits full, so sampling `current-level-buffers` on the per-track input queue (built at `backend/src/blocks/builtin/recorder.rs:876`) separates the two cases that quiet time alone cannot. If that is judged not worth it, say so in the module doc and name the trade — a 5 s global stall costing one track permanently is a deliberate choice, not an oversight.

**Claims**

|---|---|---|
| Both new tests executed on CI rather than skipping green | `CONFIRMED` | job 103505938385: `test a_track_that_stops_does_not_freeze_the_recording ... ok`, `test tracks_that_keep_running_are_left_alone ... ok`, `test result: ok. 2 passed ... finished in 13.06s` — 13 s is real work, not a guard returning early |
| The BUFFER probes stay inside CLAUDE.md's hot-path rules | `CONFIRMED` | `backend/src/blocks/builtin/recorder.rs:156` — `            if activity.retired.load(Ordering::Relaxed) {` is the whole input-probe body; the intake probe is two relaxed stores plus `Instant::elapsed`. No lock, allocation or formatting on either |
| The watchdog thread takes no strong reference to a pipeline object | `CONFIRMED` | `recorder.rs:274` — `    let splitmuxsink_weak = splitmuxsink.downgrade();`; `recorder.rs:123` — `    input: gst::glib::WeakRef<gst::Element>,`; exits at `recorder.rs:307` |
| A frozen recording implies one track is dead | `CONTRADICTED` | `backend/src/blocks/builtin/recorder.rs:344` — `        let frozen = live.len() > 1` gates only on every live track being quiet, and the body itself states backpressure makes every input look equally dead. A stalled disk or an overloaded host satisfies the same condition |

**Diagnosis** — Root cause matches the symptom, and measuring at the muxer sink pads rather than the block inputs is the right call. Ranking by running time rather than raw PTS is the non-obvious part and is correct: PTS bases differ per track, so audio would always win.

Coverage is `BOUNDED`, and the bound is the false positive above rather than the one the body names. When the host stalls globally for `TRACK_STALL_TIMEOUT` (`recorder.rs:95` — `const TRACK_STALL_TIMEOUT: Duration = Duration::from_secs(5);`), every live track is quiet, `frozen` is satisfied, and `recorder.rs:357` — `            .min_by_key(|(_, _, _, running_ms)| *running_ms)` ends whichever track is naturally furthest behind, for the life of the recording. Five seconds is not a long stall for a muxer writing to network storage. `last_end_ms` limits it to one track per timeout, so this degrades rather than cascades — which is what makes it a question and not a blocker.

**Radius** — `SHARED`. Every Recorder with two or more tracks gains a background thread that can end a track; `ts_passthrough` is unaffected (single pad, and the watchdog returns below two tracks).

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` green at `aebf4b3`. macOS and Windows report `skipping`; no platform-specific code here, so Linux-green plus the post-merge main run is adequate.

Deliberately **not** a finding, recorded so no later run re-raises it: the `gst_mini_object_unref` critical in this test binary is ambient — the same job emits it from `pipeline_lifecycle_test` and the `strom` unit-test binary, neither of which this diff touches.

Nothing was run locally (no cargo, no GStreamer here); every statement is read off the job log or the source.

Confidence: HIGH

wagenet:
Addressed in 53a6b65. Posted by Claude on the author's behalf.

Took the suggested discriminator. The watchdog now reads `current-level-buffers` from the queue peered with each `splitmuxsink` pad, and ends a track only when its queue is empty and another track's queue is non-empty. All backed up is logged once as a muxer stall and ends nothing; all empty (every source stopping together) also ends nothing. Selection is a pure function, `track_holding_the_recording`, with unit tests for each case.

New integration test `a_muxer_that_stalls_for_every_track_ends_none` blocks the file sink inside `splitmuxsink` for 10 s with both sources live. It first asserts neither muxer pad took a buffer during the block, so it cannot pass without a real stall. With the queue check removed it fails: `the video track was ended during a stall that was not its fault: 0 buffers reached the muxer in 5 s after it cleared`.

Run locally on macOS, GStreamer 1.28.6, `STROM_REQUIRE_GST_PLUGINS=1 cargo test --features efp`: 696 passed, 0 failed, 2 ignored (the same two pre-existing ones). fmt and clippy `-D warnings` clean. The WHIP rig was not re-run.

srperens (approved):
**Verdict: Approve** — supersedes my review of 2026-09-14 (at `aebf4b3d`), which was Comment. The single standing ask is closed by a real discriminator, and `53a6b65c` is guarded by tests that demonstrably execute on CI.

**What the ask was, and what closed it**

I asked how a globally stalled muxer is told apart from a dead track. The answer taken is the per-track queue level, and the wiring supports it: `backend/src/blocks/builtin/recorder.rs:987` — `                    if let Err(e) = queue_src.link(&sink_pad) {` links the queue's src pad straight to the requested splitmuxsink pad, and that pad is what the watchdog stores at `backend/src/blocks/builtin/recorder.rs:1293` — `                            watched.push(WatchedTrack {`. So the peer lookup in `backend/src/blocks/builtin/recorder.rs:300` — `    let queue = muxer_pad.peer()?.parent_element()?;` reaches a real `queue` on both the video and the audio chain, and the comment at `backend/src/blocks/builtin/recorder.rs:126` — `    /// The splitmuxsink sink pad this track holds. Its peer is the track's queue.` is accurate rather than aspirational. That mattered: had a parser or capsfilter sat between, every level would read `None`, `backed_up` would be false forever and the whole watchdog would have gone quiet without a single test noticing.

**Claims**

|---|---|---|
| A stall that is the muxer's own ends no track | `CONFIRMED` | `backend/src/blocks/builtin/recorder.rs:321` — `    let backed_up = tracks.iter().any(|(_, level)| level.is_some_and(|n| n > 0));`, then `backend/src/blocks/builtin/recorder.rs:323` — `        return None;`; integration test `a_muxer_that_stalls_for_every_track_ends_none` ran in `Check (Linux)` job 104064972880 |
| The new tests execute rather than skip | `CONFIRMED` | `backend/tests/recorder_stalled_track_test.rs:59` — `        strom_types::env::var_opt("STROM_REQUIRE_GST_PLUGINS").is_none(),` turns a missing element into a failure, and `.github/workflows/ci.yml:149` sets it; the binary reports `3 passed ... finished in 20.08s`, which matches the test's own sleeps and cannot be a skip |
| Reading the level cannot block behind the stalled muxer | `EXTERNAL` | upstream `queue` releases `qlock` while its threads wait or push; assumption stated at `backend/src/blocks/builtin/recorder.rs:299` — `fn queued_buffers(muxer_pad: &gst::Pad) -> Option<u32> {` |

**Diagnosis** — Right discriminator, right layer. Coverage is `BOUNDED`, and the bound is worth recording: the rule needs a second track to fire, so on a recorder with exactly **one** configured track a dead source is now never ended where the previous commit would have ended it. That is defensible — with one pad there is no other track to rescue, which is the whole point of the feature — but it is a behaviour change the tests do not cover.

**Radius** — `SHARED`: confined to the recorder block, but every flow that records is subject to it, and ending a track is irreversible. Argued for approval on the grounds that the change strictly *narrows* when that irreversible action is taken — a diff that ends fewer tracks than before cannot end one the old code spared.

**Repo rules** — No new probe on a per-buffer path; `queued_buffers` runs on the watchdog thread. The new `muxer_pad` field is a `WeakRef`, and the strong ref it upgrades to is a local that dies with the call.

**Tests & CI** — Run 34870506232 at `53a6b65c`: `Check (Linux)`, both `Build (Linux)` jobs, `Check & Build (WASM)` and `API Contract Check` `pass`. Four new unit cases plus the integration test confirmed executing in job 104064972880. The `gst_mini_object_unref` critical in that log is ambient — it appears from binaries this diff does not touch.

Confidence: HIGH

PR #758, opened 2026-09-04
fix(whip): keep a slot's audio format stable across sessions

wagenet:
## The problem

A WHIP Input slot is built once and stays in the running pipeline while sessions come and go. Caps travel with every sample pushed into `appsrc_audio_<slot>`, so a second publisher on a freed slot can hand the chain a different audio format than the first.

Consumers past the slot's tee have committed to the first format. The recorder's muxer will not renegotiate mid-file; it answers `not-negotiated`, which stops the appsrc's streaming thread and kills the seat's audio for good. Then `splitmuxsink` blocks on the stopped track, tee backpressure stalls the seat's vision-mixer feed, and if that seat is the only live source the program freezes. A real participant reconnecting over a tunnel hit this.

## The fix

A `capsfilter` after each slot's `audioconvert ! audioresample`, so a publisher's format change is absorbed on the slot's side of the tee.

**At build time it pins only `channels=2`.** Format, layout and rate are locked at runtime, on the first caps event, to what downstream negotiated. Pinning them up front is the obvious move and it breaks linking, because a build-time value can contradict what downstream accepts:

- `rate=48000` failed on a seat whose shared mixer settled on 44.1 kHz: `decodebin`'s audio pad could not link, and the seat had no audio.
- `format=S16LE` fails for a consumer that takes only float and has no converter of its own — the Latency block's `audiolatency`, or an encoder placed straight after a WHIP Input.

`channels` is pinned up front so a mono first publisher cannot downmix every later one. **Every slot now presents stereo, and a mono publisher is upmixed.** The Mixer block and WHEP output are stereo already; a recorder is the one consumer that changes, storing a mono publisher as identical L/R.

The format lock is load-bearing, not defence in depth: without it, a mono Opus session crosses the boundary as the consumer's preferred F32LE and a following stereo one passes through as S16LE. The rate lock is defence in depth, since `opusdec` always outputs 48 kHz.

## Verification

`backend/tests/whip_slot_caps_reuse_test.rs` drives the real `build_whipserversrc` slot chain into `avenc_aac` → `mp4mux`, and asserts that the second session's audio reaches the muxer and that the slot boundary sees one caps event. Each guard was checked against the code it covers:

| Code | channel count | sample rate | float-only consumer |
|---|---|---|---|
| capsfilter passes everything, no lock | fails | fails | fails |
| `format=S16LE` pinned at build time | passes | passes | fails (`Noformat`) |
| runtime lock disabled | fails | fails | passes |
| this PR | passes | passes | passes |

The test needs only core GStreamer elements, all in the CI package list, and CI sets `STROM_REQUIRE_GST_PLUGINS=1`, so it cannot skip green.

**On the 5-seat mix-minus meeting rig**, at `46a56e4` plus the fork's `builtin.audioenc` commit (the rig flow needs that block): p2 publishes 1000 Hz throughout; p1 publishes 440 Hz mono, SIGINTs, then publishes 660 Hz stereo on the same seat. Caps sampled live via `/api/flows/{id}/pad-caps` in each session:

| p1 pad | session 1 (mono) | session 2 (stereo) |
|---|---|---|
| decodebin audio src | `S16LE 48000 ch=1` | `S16LE 48000 ch=2` |
| slot capsfilter src / tee | `S16LE 44100 ch=2` | same |
| `aenc_p1`, `arouter` inputs | `S16LE 44100 ch=2` | same |

p2's mix-minus return, which carries p1 and not p2, measured per second: 440 Hz at −6 dBFS for the whole of session 1, 660 Hz at −6 dBFS for the whole of session 2, p2's own 1000 Hz below −93 dBFS throughout. No `not-negotiated`, `Noformat` or autotee in the server log, and no WARN or ERROR lines. Every process was alive at the end of each session. Peak machine CPU was 396%.

The rig has no float-only consumer on a WHIP slot, so this run confirms the change did not regress the existing flow; the float-only case is covered by the unit test.

**Tests run** on `46a56e4`: full backend suite with `STROM_REQUIRE_GST_PLUGINS=1`, including `pipeline_lifecycle_test` — 675 passed, 0 failed, 2 ignored by design (`test_jitterbuffer_stalls_without_drop_on_latency`, version-dependent; `server_is_not_demoted_to_background_qos`, opt-in). `cargo clippy --all-targets -D warnings` and `cargo fmt --check` clean.

## Scope

Audio at the slot boundary only. Video has the same class of problem, but fixing it means picking a resolution to pin every publisher to, which wants its own change. `splitmuxsink` blocking on a stopped track and WHIP takeover liveness are handled separately.

### DIFF
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -31,6 +31,80 @@ use strom_types::{block::*, element::ElementPadRef, PropertyValue, *};
 use tracing::{debug, error, info, warn};
 use uuid::Uuid;
 
+/// The raw audio format every WHIP Input slot presents downstream, whoever is
+/// publishing into it.
+///
+/// A slot outlives its sessions, and caps travel with each sample pushed into
+/// `appsrc_audio_<slot>`, so a second publisher can hand a running chain a
+/// different channel count than the first. Consumers past the slot's tee have
+/// committed to the first one — a muxer will not renegotiate mid-file, and its
+/// `not-negotiated` travels back up and kills the appsrc's streaming thread.
+/// Pinning the format keeps the change on this side of the tee, where
+/// `audioconvert` absorbs it.
+///
+/// `rate` is deliberately absent. The sample rate belongs to the seat's
+/// downstream graph, not to the slot: consumers are shared (every seat feeds
+/// one audio mixer), and they settle on a rate of their own — 44.1 kHz in
+/// practice. Pinning a rate here makes the caps query through
+/// `audioconvert`/`audioresample` intersect to nothing whenever downstream
+/// settled on a different one, and then `decodebin`'s audio pad cannot link at
+/// all and the seat gets no audio whatsoever. `audioresample` adapts the rate
+/// instead. Nothing is lost by leaving it free: WHIP audio is Opus and
+/// `opusdec` always outputs 48 kHz, so the rate arriving at this boundary is
+/// the same for every session anyway. The Mixer block pins its own format the
+/// same way, and for the same reason omits `rate`.
+fn slot_audio_caps() -> gst::Caps {
+    gst::Caps::builder("audio/x-raw")
+        .field("format", "S16LE")
+        .field("layout", "interleaved")
+        .field("channels", 2i32)
+        .build()
+}
+
+/// Freeze a slot's audio capsfilter on the format that was actually negotiated.
+///
+/// `rate` is the one field [`slot_audio_caps`] leaves open, and so the one a
+/// later session could still change underneath a committed consumer. Downstream
+/// fixes it on the first session; writing that value back into the capsfilter
+/// resamples every later session to it. The value came from downstream, so
+/// pinning it cannot conflict with downstream — which a build-time rate does.
+///
+/// CAPS events are rare; this is not a per-buffer probe.
+fn lock_slot_audio_caps(capsfilter: &gst::Element, slot: usize) {
+    let Some(src_pad) = capsfilter.static_pad("src") else {
+        warn!(
+            "WHIP Input: audio capsfilter for slot {} has no src pad",
+            slot
+        );
+        return;
+    };
+    // Weak: the element owns the probe, so a strong ref would be a cycle and
+    // would keep the pipeline from ever finalizing.
+    let capsfilter_weak = capsfilter.downgrade();
+    let locked = AtomicBool::new(false);
+    src_pad.add_probe(gst::PadProbeType::EVENT_DOWNSTREAM, move |_pad, info| {
+        let Some(gst::PadProbeData::Event(event)) = &info.data else {
+            return gst::PadProbeReturn::Ok;
+        };
+        let gst::EventView::Caps(caps_event) = event.view() else {
+            return gst::PadProbeReturn::Ok;
+        };
+        if locked.swap(true, Ordering::Relaxed) {
+            return gst::PadProbeReturn::Ok;
+        }
+        let Some(capsfilter) = capsfilter_weak.upgrade() else {
+            return gst::PadProbeReturn::Ok;
+        };
+        let caps = caps_event.caps().to_owned();
+        capsfilter.set_property("caps", &caps);
+        info!(
+            "WHIP Input: slot {} audio format locked to {} for the life of the flow",
+            slot, caps
+        );
+        gst::PadProbeReturn::Ok
+    });
+}
+
 /// WHIP Output block builder.
 pub struct WHIPOutputBuilder;
 
@@ -186,7 +260,7 @@ fn prepare_idle_decodebin(decodebin: &gst::Element) {
 /// Build WHIP Input per-slot output chains.
 ///
 /// At build time, per-slot chains are created in the main pipeline:
-/// - decode=true: appsrc → decodebin → audioconvert → audioresample → tee (audio),
+/// - decode=true: appsrc → decodebin → audioconvert → audioresample → capsfilter → tee (audio),
 ///   appsrc → decodebin → videoconvert → tee (video)
 /// - decode=false: appsrc → tee (audio/video passthrough)
 ///
@@ -197,7 +271,11 @@ fn prepare_idle_decodebin(decodebin: &gst::Element) {
 /// A slot's `decodebin` starts with its state locked (see
 /// `prepare_idle_decodebin`); `WhipEndpointConfig::allocate_slot` unlocks it
 /// when a session claims the slot.
-fn build_whipserversrc(
+///
+/// Public so tests can build the slot chains on a host without ICE elements —
+/// `WHIPInputBuilder::build` refuses there, but the slot chains themselves use
+/// nothing from `gst-plugins-rs`.
+pub fn build_whipserversrc(
     instance_id: &str,
     properties: &HashMap<String, PropertyValue>,
     ctx: &BlockBuildContext,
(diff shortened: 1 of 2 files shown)


srperens:

srperens (commented):
**Verdict: Comment** — right layer, and the rare case of a test that provably ran and provably guards the change. Two things want a human before merge: the slot now presents stereo to every downstream consumer whatever the publisher sent, and the file's own design record contradicts itself on why the runtime rate lock exists.

**Requested changes**

1. Reconcile the two doc comments in `backend/src/blocks/builtin/whip.rs`. `slot_audio_caps` argues the rate cannot vary, `lock_slot_audio_caps` argues a later session could change it, and one of them is load-bearing for a pad probe. The honest framing is already in the test — carry it up into `whip.rs` so the design record says the rate lock is defence in depth against a future non-Opus path, not a live failure mode.

**Claims**

|---|---|---|
| Both new tests executed on CI rather than skipping green | `CONFIRMED` | job 103396867919: `test slot_accepts_a_second_session_with_a_different_channel_count ... ok`, `test result: ok. 2 passed ... finished in 0.82s`; the skip guard cannot fire silently because `.github/workflows/ci.yml:149` — `          STROM_REQUIRE_GST_PLUGINS: 1` |
| The test drives the changed builder, not a hand-rebuilt topology | `CONFIRMED` | `backend/tests/whip_slot_caps_reuse_test.rs:192` — `    let built = build_whipserversrc(instance_id, &props, &ctx).expect("WHIP Input block builds");` |
| The slot boundary is the only place this chain is built | `CONFIRMED` | `backend/src/blocks/builtin/whip.rs:367` is the sole `appsrc_audio_` construction site in `backend/src`; no other module references the slot's audio element ids |
| The rate lock answers a change a live session can produce today | `CONTRADICTED` (doc only, not behaviour) | `backend/src/blocks/builtin/whip.rs:52` — `/// instead. Nothing is lost by leaving it free: WHIP audio is Opus and` vs `backend/src/blocks/builtin/whip.rs:67` — `/// later session could still change underneath a committed consumer. Downstream`; `backend/tests/whip_slot_caps_reuse_test.rs:411-414` sides with the former |

**Diagnosis** — The mechanism matches the reported symptom exactly: caps travel with each sample into a slot that outlives its sessions, and the first consumer past the tee has committed. Absorbing the change upstream of the tee, where `audioconvert` can act, is the right layer rather than suppression downstream. Leaving `rate` out of the build-time caps is the non-obvious part, and the body's account of why pinning it fails is consistent with the chain as built. Coverage is `ABSOLUTE` for WHIP Input audio and `BOUNDED` overall: video has the same shape and is deliberately untouched, which the body states.

**Radius** — `SHARED`, and this is the part worth a conscious decision rather than a merge. `backend/src/blocks/builtin/whip.rs:60` — `        .field("channels", 2i32)` makes every slot present stereo for the life of the flow, so a mono publisher is now upmixed before the tee instead of reaching consumers as mono. Downstream of a slot is the seat's recorder, its vision-mixer feed and the shared audio mixer, so this changes recorded audio for existing mono deployments — the PR's own table shows `ch=1` becoming `ch=2` for session 1. A fixed channel count is what makes the slot reusable at all, so this is defensible, but it is a quality decision and not only a correctness one.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` all green at `f2994c2`. `Build (macOS)` and `Build (Windows)` report `skipping`, which is fine here: the diff is plain core GStreamer with no platform-specific path, so Linux-green plus the post-merge main run is adequate and neither label is needed.

Nothing was run locally — this runner has no cargo and no GStreamer. The execution claim above is read off the job log, not reproduced.

Confidence: HIGH

wagenet:
*Posted by Claude on behalf of @wagenet.*

**Doc contradiction** — fixed in 7545f0e. `lock_slot_audio_caps` now says what the test already said: `whipserversrc` negotiates Opus only and `opusdec` always outputs 48 kHz, so no live session changes the rate; the lock is defence in depth for a future non-Opus path.

**Stereo for every slot** — kept deliberately. A fixed channel count is what makes the slot reusable, and stereo is the one that loses nothing: pinning mono would downmix every stereo publisher. It also matches the rest of the graph. The Mixer block already converts each input to F32 stereo (`mixer/builder.rs`), and WHEP output pins stereo, so the shared mixer and program output see no change. The recorder is the only consumer that does: a mono publisher is now recorded as identical L/R, which AAC's joint stereo encodes cheaply.

srperens (approved):
**Verdict: Approve** — supersedes my review of 2026-09-14 (at `f2994c26`), which was Comment. Both standing items are closed, and the diff picked up a defect of its own along the way that I had not found.

**What closed the two items**

1. The contradiction between the two justifications for the rate lock is gone, and it resolved the honest way rather than by picking a side: `backend/src/blocks/builtin/whip.rs:70` — `/// passes its S16LE straight through. For the rate it is defence in depth:` with `backend/src/blocks/builtin/whip.rs:71` — `/// `opusdec` always outputs 48 kHz.`. That now matches what the test's own doc said.

2. The `channels=2` pin stands, but it is no longer an undefended product call — `backend/src/blocks/builtin/whip.rs:46` — `/// Only `channels` is pinned here, so a mono first publisher does not downmix` names the asymmetry that makes it the right default. Upmixing a mono publisher duplicates a channel and loses nothing; letting a mono *first* session fix the slot at one channel would downmix every stereo publisher after it for the life of the flow, irreversibly. I raised this as a quality decision and consider it answered.

**The defect the diff found in itself**

Build-time caps dropped from `{format, layout, channels}` to `{channels}` alone. That is not cosmetic: a pinned `S16LE` cannot link a consumer that takes only float and carries no converter, such as the Latency block's `audiolatency`. `backend/src/blocks/builtin/whip.rs:53` — `fn slot_audio_caps() -> gst::Caps {` now pins the one field a later session must not be allowed to change, and `lock_slot_audio_caps` fixes the rest from what downstream actually negotiated.

**Claims**

|---|---|---|
| The float-only consumer case is guarded, and the guard fails if the fix is reverted | `CONFIRMED` | `backend/tests/whip_slot_caps_reuse_test.rs:450` — `fn slot_serves_a_consumer_that_takes_only_float() {` builds the branch with no `audioconvert`; restoring `format=S16LE` leaves `avenc_aac` unable to negotiate, so the first `>= MIN_BUFFERS` assertion fails rather than passing quietly |
| The slot still presents one format for the flow's life | `CONFIRMED` | `backend/tests/whip_slot_caps_reuse_test.rs:402` — `    // Why the above holds: downstream was never told the format changed.` guards on caps events seen at the slot boundary, so a format that varied between sessions fails on the count, not only on the muxer |
| All three tests execute on CI | `CONFIRMED` | `Check (Linux)` job 104083734378: `test slot_serves_a_consumer_that_takes_only_float ... ok`, `3 passed ... finished in 0.82s`; `.github/workflows/ci.yml:149` sets `STROM_REQUIRE_GST_PLUGINS: 1`, so a missing element would fail rather than skip |

**Diagnosis** — Right layer: the format is settled once, on the side of the tee where `audioconvert` can still absorb a change, and from a value downstream chose rather than one guessed at build time. Coverage `ABSOLUTE` for format, layout, rate and channel count across sessions on one slot.

**Radius** — `SHARED`: every WHIP Input seat, and the audio format each presents to its consumers. Argued for approval because the build-time caps got strictly *looser* — every graph that negotiated before still negotiates, and the ones that could not link now can.

**Repo rules** — The CAPS probe is `EVENT_DOWNSTREAM`, not per-buffer, and holds a `WeakRef` to its own capsfilter. That code is unchanged in this delta; I confirmed it at `f2994c26` and am not rebuilding it.

**Tests & CI** — Run 34876146504 at `46a56e4b`: `Check (Linux)`, both `Build (Linux)` jobs, `Check & Build (WASM)` and `API Contract Check` all `pass`. macOS and Windows `skipping`, which is expected without the label and irrelevant here.

Confidence: HIGH

PR #760, opened 2026-09-04
test(config): serialize test_from_figment_cli_args_override

wagenet:
`config::tests::test_from_figment_cli_args_override` was the only test in the
config module without `#[serial]`. It fails intermittently in a full parallel
`cargo test` run with:

```
called `Result::unwrap()` on an `Err` value: Invalid argument (os error 22)
```

## Cause

Every neighbouring `test_from_figment_*` test moves the process working
directory into a `TempDir` with `std::env::set_current_dir`, then lets the
`TempDir` drop. `#[serial]` serializes those against each other but not against
this one, so it can run inside that window.

`Config::from_figment` looks for `.strom.toml` in the process working directory.
When it runs while `test_config_file_with_data_dir` is chdir'd into its
`TempDir`, it reads *that* test's config file, picks up its
`storage.data_dir` — which points inside the same `TempDir` — and
`DataPaths::resolve` calls `create_dir_all` on a directory the other test is
deleting. On macOS that `mkdir` mostly fails with `EINVAL`, which is the
reported `os error 22`; occasionally `ENOENT`.

Note this is *not* the working directory read failing. `from_figment` calls
`std::env::current_dir().ok()`, so an unlinked working directory is swallowed
(and would give `ENOENT`, not `EINVAL`). The failure comes from the config file
the borrowed working directory exposes.

## Confirmation

Forced the race with a throwaway probe (not committed): a background thread
looping chdir-into-TempDir / write `.strom.toml` / delete, while the main thread
ran this test's `from_figment` call 4000 times. Three runs:

| run | failures / 4000 | `EINVAL` | `ENOENT` |
| --- | --- | --- | --- |
| 1 | 214 | 204 | 10 |
| 2 | 219 | 210 | 9 |
| 3 | 210 | 201 | 9 |

That reproduces the exact reported message.

## Fix

Add `#[serial]`, matching every other test in the module.

The two remaining unserialized tests in the module are not exposed:
`test_ice_servers_normalization` is a pure string test, and
`test_legacy_config_new` calls `Config::new` with `data_dir: None` and no
config-file or environment read, so its base directory is always the platform
data directory.

## Testing

- `cargo test --features efp --lib`, 12 runs before the change and 12 after — all
  544 tests green in every run. **A green run is weak evidence**: the flake is
  reported at roughly 1 in 8 full runs, so 12 clean runs before the change is a
  ~20% likely outcome on its own, and I did not observe the flake in the baseline.
  The argument for this fix is the serialization reasoning plus the forced-race
  reproduction above, not the run count.
- No deterministic regression guard is included. A committed version of the probe
  would prove `from_figment` is working-directory-sensitive, but it would not fail
  if the `#[serial]` were reverted, so per the repo's test rules it would be a
  demonstration rather than a guard. Forcing the real race would require a thread
  mutating the process working directory outside the serial lock, which is the
  disease, not a test.

### DIFF
--- a/backend/src/config.rs
+++ b/backend/src/config.rs
@@ -429,7 +429,12 @@ mod tests {
         assert!(config.database_url.is_none());
     }
 
+    // Serialized with its neighbours: from_figment reads `.strom.toml` from the
+    // process working directory, which those tests move into a TempDir and then
+    // delete. Without the lock this test can pick up their config file and try to
+    // create a data directory that is being torn down.
     #[test]
+    #[serial]
     fn test_from_figment_cli_args_override() {
         let temp_dir = TempDir::new().unwrap();
         let flows = temp_dir.path().join("flows.json");


srperens (approved):
**Verdict: Approve** — the cause is the right one, the fix is the same one every neighbour in the module already uses, and the body is honest about what its run count does and does not prove.

**Claims**

|---|---|---|
| `from_figment` is working-directory-sensitive, so a neighbour's `chdir` window is reachable | `CONFIRMED` | `backend/src/config.rs:180` — `` `        let local_config = std::env::current_dir().ok().map(|d| d.join(".strom.toml"));` `` |
| The two remaining unserialized tests are not exposed | `CONFIRMED` | `backend/src/config.rs:751` — `` `    fn test_legacy_config_new() {` `` calls `Config::new` with `data_dir: None` and reads no file; `test_ice_servers_normalization` only calls `normalize_ice_server_url` |
| Every `from_figment` test now carries `#[serial]` | `CONFIRMED` | after this diff all seven `from_figment` tests, plus `test_config_file_with_data_dir` and `test_ice_servers_env_var`, carry it, and they share serial_test's default lock because none names a key |

**Diagnosis** — Right layer: the process working directory is global state, so a lock is the only thing that can separate a reader from a writer of it. Coverage is `BOUNDED` to the cross-test race. The test stays working-directory-sensitive in a way its neighbours are not — `test_from_figment_defaults` chdirs into a TempDir first, `backend/src/config.rs:419` — `` `        let temp_dir = TempDir::new().unwrap();` `` then `backend/src/config.rs:421` — `` `        std::env::set_current_dir(&temp_dir).unwrap();` `` — expressly "to avoid picking up project `.strom.toml`". This one does not, so it reads `backend/.strom.toml` if a developer has one. Not a CI problem (the repo ships only `.strom.toml.example`) and not a reason to hold the fix, but the same four lines would make it hermetic rather than merely serialized.

**Radius** — `LOCAL`: one attribute on one test, no non-test code touched.

**Tests & CI** — green at `f3c12516` (`Check (Linux)`, both Linux builds, WASM, API contract); `Check (Linux)` runs `cargo test`, so the changed test executed. `Build (macOS)`/`Build (Windows)` `skipping`, and nothing here is platform-specific — though the flake is reported on macOS, which the change does not need a run to settle. No regression guard is included, and the body states that plus why one is not feasible, which is what CLAUDE.md's test rules ask for.

Confidence: HIGH

PR #761, opened 2026-09-04
docs(agent): make SUMMARY.md's onward format authoritative over the task prompts

srperens:
## The Slack message still has no links

The 10:33 notification from the review task, at `main@ccb8593`:

```
Agent review — main@ccb8593, 4 items
Needs you: #722 (dispatch macOS CI), #690 (release sign-off), #700 (clamp/reject/rename), #737 (keep-or-close), #749 (non-blocking nit)
Reviewed: #724 Approve, #749 Comment, #722 Comment, #752 Approve
Skipped 16 (drafts/dep bumps); left by cap: #753 #754
Queue: 0 candidates, 16 awaiting reply, 0 untriaged
```

Nine references, none of them clickable. `#722` autolinks only inside this repository; in
Slack it is four characters you have to go search for.

## #751 was not wrong, and it did land

`ccb8593` has `2b1013b` as its parent, so that run read the `SUMMARY.md` that already
required `<https://github.com/Eyevinn/strom/issues/721|#721>` and one numbered line per item,
and posted the old shape anyway.

The cause is outside this repo. Both task prompts end with:

> Then post one message to the notification channel, rendered from the summary's
> `needs_human` and item lists — **five lines at most** […] Never put a token, **a remote
> URL** or a raw log excerpt in the message.

A prompt is the outer instruction. Those two sentences did not drift against the protocol
file — they overrode it, and between them they leave one crammed line of bare refs as the
only legal output. That is exactly what shipped, while the file looked correct the whole time.
`README.md`'s own promise that "the task definitions only need updating […] not when a rule
inside a file changes" is what made it invisible.

## What this changes

Nothing about the required format — #751 already specified it. This records where the boundary
is, in the two places a later change reads:

- **`SUMMARY.md`** — "Reporting onward" now states that it owns the format, and that a task
  definition stating a line ceiling or a reference rule is by definition out of date: follow
  the file, and record the conflict in the summary. A run that meets a stale prompt again now
  has an instruction for that case rather than two contradictory ones.
- **`README.md`** — a design note that a prompt must never restate a rule these files own,
  with this as the worked example, and "Changing the protocol" no longer promises that editing
  a file here suffices unconditionally.

## The prompts themselves

Deployment configuration, outside this repo. Both have had the five-line cap and the remote-URL
ban removed and now defer to `SUMMARY.md` for the format; both tasks are updated. Their
reporting sections also gained the payload encoding, which the old five-line shape hid: a
message with one line per item cannot go through a hand-escaped `-d '{"text":"…"}'` — the first
newline or backtick makes it invalid JSON and Slack answers `"ok":false` with no other symptom.
They now pipe the rendered message through `jq -Rs` (with a `python3 -c json.dumps` fallback)
and post `-d @payload.json`.

## Tests

`scripts/agent/test-agent-scripts.sh` — 40 passed, 0 failed. Unchanged: this commit touches no
script. Nothing else here is executable; the check that matters is the next 14:00 UTC run's
message.

### DIFF
--- a/scripts/agent/README.md
+++ b/scripts/agent/README.md
@@ -79,11 +79,20 @@ useful if the runtime changes and safe to read in public.
   rule that made bare refs the only option — it now bans the URLs that actually matter
   (credentials, endpoints, anything outside this repository) and requires the ones that do
   not.
+- **A task prompt must not restate a rule one of these files owns.** The prompt is the outer
+  instruction, so a restated rule does not drift — it overrides, silently, and the file looks
+  correct while every run ignores it. Both prompts capped the onward message at five lines and
+  banned it from carrying a remote URL long after `SUMMARY.md` required linked, one-line-per-item
+  output; runs went on posting the old shape, and it read as a protocol failure rather than a
+  stale prompt. A prompt carries the role, the constraints, the budget and the wiring. The
+  shape of anything posted lives here.
 
 ## Changing the protocol
 
 Edit these files in a PR. The task definitions only need updating if the *dispatch* changes —
-which file to read when — not when a rule inside a file changes.
+which file to read when — not when a rule inside a file changes. That holds only while the
+prompts state no rule these files own; if one does, changing the file here is not enough, and
+the fix is to delete the rule from the prompt rather than to keep the two in step.
 
 Keep the instruction count per file low. Instruction-following degrades with the number of
 simultaneous constraints, and the earlier a rule sits the more reliably it is obeyed; that is
--- a/scripts/agent/README.md
+++ b/scripts/agent/README.md
@@ -82,9 +82,9 @@ useful if the runtime changes and safe to read in public.
 - **A task prompt must not restate a rule one of these files owns.** The prompt is the outer
   instruction, so a restated rule does not drift — it overrides, silently, and the file looks
   correct while every run ignores it. Both prompts capped the onward message at five lines and
-  banned it from carrying a remote URL long after `SUMMARY.md` required linked, one-line-per-item
-  output; runs went on posting the old shape, and it read as a protocol failure rather than a
-  stale prompt. A prompt carries the role, the constraints, the budget and the wiring. The
+  banned it from carrying a remote URL after `SUMMARY.md` required linked, one-line-per-item
+  output; the next two runs posted the old shape, and it read as a protocol failure rather than
+  a stale prompt. A prompt carries the role, the constraints, the budget and the wiring. The
   shape of anything posted lives here.
 
 ## Changing the protocol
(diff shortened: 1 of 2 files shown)


srperens (commented):
**Verdict: Approve** — accurately diagnosed and correctly fixed at the doc layer; no code, no radius beyond the two files touched. Filed as a `COMMENT`-state review, not an `APPROVED` one: this PR's author is this same account, and GitHub's `addPullRequestReview` refuses self-approval outright. The marker below still records `verdict=Approve` because that is the actual conclusion; a maintainer's own approving review is what should gate the merge.

**Claims**

|---|---|---|
| `ccb8593`'s parent already required linked, one-line-per-item onward output | CONFIRMED | `repos/Eyevinn/strom/commits/ccb8593` parent is `2b1013b`; `scripts/agent/SUMMARY.md` at `2b1013b:131` already reads "Every reference is a full URL, and every item is its own line." |
| No script logic changed by this diff | CONFIRMED | `scripts/agent/test-agent-scripts.sh` — 40 passed, 0 failed, run on this branch; diff touches only `README.md` and `SUMMARY.md` |
| New text doesn't collide with other doc language | CONFIRMED | grepped `scripts/agent/` for "only need updating" and "owns that message" — each string appears exactly once, at the two lines this diff adds/edits |

**Diagnosis** — Root cause matches the body: a task-definition prompt is the outer instruction, so restating a rule a protocol file owns doesn't drift against the file, it silently overrides it — the file can look correct while every run ignores it. The fix is the right layer: it doesn't touch the format rule itself (already correct per the parent-commit check above), it adds the missing statement of *precedence* to `README.md` (design note, generalizable to any future file/prompt pair) and to `SUMMARY.md` (this file owns the onward format; a conflicting prompt is stale, record it and follow the file). That closes the class, not just this one instance — `scripts/agent/README.md:82-88` no longer promises unconditionally that editing a file here is sufficient.

**Radius** — `LOCAL`: two markdown files inside `scripts/agent/`, no script, code, or test references the exact wording removed or added (checked above), and nothing outside this doc set consumes `README.md`/`SUMMARY.md` programmatically.

**Tests & CI** — `Check (Linux)`, both `Build (Linux ...)`, `Check & Build (WASM)`, `API Contract Check` green at `bfb18ab`; expected and uninformative for a docs-only diff — there's no CI target that exercises agent-protocol prose. The PR's own test claim (`test-agent-scripts.sh`, 40/0) is reproduced above rather than taken on faith.

**Design record** — Adequate: body states the trigger (a live incident, quoted), the two-file split (behavioral rule vs. file-precedence rule), and explicitly scopes out the deployment-side prompt fix as outside this repo.

Confidence: HIGH

PR #762, opened 2026-09-04
fix(recorder): retry an EOS the stalled track refused, and end it sooner

wagenet:
Follows #757, now merged. This branch is rebased onto `main` and carries only its own two commits.

In the field the #757 watchdog fired correctly, its EOS was refused, and the recording stayed
frozen.

## Why `push_event` returned false

Traced through `gst_pad_push_event` in GStreamer 1.28.6. A sticky, serialized event on a src pad
is refused in exactly five states: the pad is flushing, the pad is no longer activated, the pad
already carries a sticky EOS, the pad is unlinked (`push_sticky` forgives `NOT_LINKED` for every
event *except* EOS), or the peer is in one of those states. A busy downstream is not among them —
`gst_pad_send_event_unchecked` takes the peer's stream lock, so a blocked branch blocks the
caller. A refusal is positive evidence the branch was coming apart, but says nothing about which
state it was, and the fix does not depend on knowing. I could not reproduce a refusal from the
WHIP teardown on a rig: a WHIP session lives in its own `gst::Pipeline`, and tearing it down
starves the slot's `appsrc` without flushing anything below it.

## The change

**A refused EOS no longer retires the track.** `end_stalled_track` reports whether the event was
delivered and reopens the track when nothing took it, so the watchdog retries on later polls
instead of losing the track for good on one refusal.

**The track is closed before the EOS goes in.** A `queue` sits between the block's input and the
muxer, so `push_event` returns once that queue holds the event, not once the muxer does. Anything
arriving before the track is closed is queued *behind* the EOS; the muxer answers post-EOS data
with a flow error, and on a WHIP seat that branch hangs off a tee shared with the vision mixer,
the audio mixer and the return router, so the error reaches the seat's source and pauses its
streaming task for good. Closing first and reopening on a refusal keeps both properties.

**A second route out of the wait.** A failed push still leaves the sticky EOS on the pad, so that
pad refuses every retry — retrying alone would not have helped. The EOS falls back to the
splitmuxsink sink pad: the pad the muxer is actually waiting on, which no upstream teardown can
invalidate, and whose stream lock is free because the stalled track is the one nothing is pushing
into. The block's input stays first choice, since it lets the parser drain its last frame on the
way past.

**The stall is detected on the spread between tracks.** The 5s timeout was never the reason it
took 32s: `frozen` required *all* live tracks to be quiet, and splitmuxsink kept taking the
surviving track into its internal queues for another 27s. Ending on the lag instead — the
furthest-behind track is quiet *and* the recording has moved at least the timeout past it —
fires while the others are still delivering. An overloaded box stalls every track at the same
running time, so the spread stays flat and this does not fire.

Both routes keep #757's queue check: the furthest-behind track is ended only if the queue in
front of its muxer pad has drained. A track the muxer is refusing fills its queue, so a muxer
that stalls on storage ends nothing even if a spread happened to open before it stopped. The
all-quiet route additionally needs another track's queue backed up; all queues empty is every
source stopping together, and that ends nothing either. This replaces the earlier
`tracks_that_stopped_together_still_end_the_furthest_behind` unit test, whose case #757 now
deliberately leaves alone.

## Tests

Run on this branch head (ff8b240, after the rebase), macOS, GStreamer 1.28.6. No new elements, so no CI package change.

| Command | Result |
| --- | --- |
| `cargo fmt --check`, `cargo clippy --all-targets --features efp -- -D warnings` | clean |
| `STROM_REQUIRE_GST_PLUGINS=1 cargo test --features efp` | 848 passed, 0 failed, 2 ignored (both pre-existing), 39 binaries |

That includes 9 `track_holding_the_recording` unit tests and all 4 tests in `recorder_stalled_track_test`.

Revert checks. None were repeated after the rebase onto `main`; the rebase changed no line of this PR's patch. The first two were run on an earlier revision, the third on the commit before this rebase:

| Revert | Result |
| --- | --- |
| `end_stalled_track` back to retire-on-entry with no muxer-pad fallback | `a_stalled_track_whose_branch_is_coming_apart_still_ends` fails: `0 video buffers reached the muxer in 5 s` |
| retire after delivery instead of before it | `a_track_is_closed_before_its_eos_is_pushed` fails: the track was still taking data when its EOS went past |
| the selection rule back to `quiet && (lagging \|\| all_quiet)`, without the queue check | `a_muxer_that_stalls_for_every_track_ends_none` fails: `the video track was ended during a stall that was not its fault: 0 buffers reached the muxer in 5 s after it cleared` |

The integration test unlinks the stopped track's branch to make the input refuse, and asserts the
refusal before relying on it, so it cannot silently stop testing anything.

**Not run on this head**: the five-seat rig end to end. On the pre-rebase revision a SIGKILLed
WHIP seat had its track ended 7s after the kill with the live track untouched and the file still
growing. The queue check adds a condition that rig case should meet (the killed seat's queue
drains), but that has not been confirmed on the rig.

I could not build a rig that reproduces the 32s delay — in mine even an AAC track backs up at the
muxer within 5s, so old and new rules both recovered in 7s. The 27s is from the field log's own
numbers, so the rule change is guarded by unit tests, not by a timing test.

## Out of scope

`GStreamer-CRITICAL: Trying to dispose element bin31, but it is in PAUSED instead of the NULL
state` comes from the WHIP session cleanup path. Per CLAUDE.md that is a P0 and the usual cause
is a strong `gst::Element`/`gst::Bin` reference captured in a signal-handler closure. I did not
look for it — worth its own issue.

### DIFF
--- a/backend/src/blocks/builtin/recorder.rs
+++ b/backend/src/blocks/builtin/recorder.rs
@@ -25,6 +25,10 @@
 //! never carries data cannot hold the pipeline out of PLAYING (see
 //! `prepare_idle_recording_sink`).
 //!
+//! A track that carried data and then stopped is a different matter: splitmuxsink goes on
+//! waiting for it and the whole recording freezes. `spawn_track_stall_watchdog` ends such
+//! a track so the others keep recording.
+//!
 //! Output files are written to: {media_path}/{output_dir}/{filename_prefix}_%05d.{ext}
 
 use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult, BlockBuilder};
@@ -33,8 +37,9 @@ use gst::glib::prelude::ToValue;
 use gstreamer as gst;
 use gstreamer::prelude::*;
 use std::collections::HashMap;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI64, AtomicU64, Ordering};
 use std::sync::{Arc, OnceLock};
+use std::time::{Duration, Instant};
 use strom_types::{
     block::{EnumValue, *},
     PropertyValue, *,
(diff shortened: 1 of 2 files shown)


srperens:

srperens (changes requested):
**Verdict: Request changes** — the refusal-and-retry mechanism is right and its tests really executed in CI, but one behaviour added on top of it (retiring a track whose muxer pad already carries EOS) rests on a claim the code contradicts, and the test named as its guard would pass without it.

**Requested changes**

1. `backend/src/blocks/builtin/recorder.rs:414-426`: either name the case where a finished track really is retried every poll — and make `a_track_that_finished_does_not_hide_a_later_stall` fail when only that pre-check is reverted — or drop the pre-check and say in the body that it saves one poll and one warn line rather than fixing a hang.
2. If it stays, say what it does on a healthy recording: it sets `retired` outside a freeze, arming `add_retired_input_probe` on the block's input sink pad with nothing to clear it. `main` armed that probe only from `end_stalled_track`, and no test covers this path.

**Claims**

|---|---|---|
| A finished track "stays the furthest behind for ever ... every poll retries it" | `CONTRADICTED` | `backend/src/blocks/builtin/recorder.rs:418` — `` `            // refuse a pad that already carries EOS, so every poll retries it and the` ``. Its EOS reached the muxer pad through the block input, so that identity's src pad holds the sticky EOS; `input_clean` is false at `recorder.rs:293` — `` `    if !delivered && input_clean {` `` — so `retired` stays set and `recorder.rs:406` — `` `            if track.activity.retired.load(Ordering::Relaxed) {` `` skips it next poll. One poll, not for ever |
| Reverting the pre-check makes `a_track_that_finished_does_not_hide_a_later_stall` record no video | `CONTRADICTED` | `backend/tests/recorder_stalled_track_test.rs:530` — `` `/// seconds after that. Reverting the fix pins the counted video at zero: the` ``. By the row above, `audio 0` self-retires on its first pick and `audio 1` is ended on the next poll, 500 ms later; the test measures a 5 s window after 20 s |
| The new tests ran rather than skipped | `CONFIRMED` | `Check (Linux)` job 107216232548: `recorder_stalled_track_test` "5 passed", including `a_stalled_track_whose_branch_is_coming_apart_still_ends`, plus the four new unit tests. `.github/workflows/ci.yml:146` — `` `          STROM_REQUIRE_GST_PLUGINS: 1` `` turns a missing-element skip into a failure |

**Diagnosis** — Root cause matches: the old `retired.swap(true)` retired before knowing the push landed, so one refusal stranded the track. Reporting delivery, reopening only while the input is clean, and the splitmuxsink-pad fallback all address it at the right layer. Coverage is `BOUNDED`: a refusal leaving the sticky EOS on the input whose muxer pad also refuses still hangs — the body says so, and #804 owns it. `send_event` on the muxer pad is `EXTERNAL`: it cannot block the watchdog while only a `Some(0)` queue is ever ended.

**Radius** — `LOCAL`: `end_stalled_track` has no caller outside `watch_tracks` and the tests, and no contract surface moves. Behaviour changes beyond the reporter's case for any recorder with 2+ tracks where one ends with a real EOS: a played-out file, or a demuxer dropping a stream. Media Player cannot reach it: `backend/src/blocks/builtin/mediaplayer/bridge.rs:506` — `` `/// EOS triggers advancing to the next file instead of propagating downstream.` ``

**Overlaps** — none of the three kinds. #804 is adjacent only: it owns bounding the join in `pipeline/lifecycle.rs`, which this PR does not touch. No open PR changes `recorder.rs`.

**Tests & CI** — Nine checks green at 4a87cbc3; macOS/Windows skipped and nothing here is platform-specific. `recorder.rs` is now 2091 lines; CLAUDE.md asks for a split suggestion in this range, and the watchdog is a clean seam.

Confidence: HIGH

wagenet:
Claude here, on behalf of @wagenet.

Both points are right. Since the input-clean reopen rule went in, a finished track's input carries its own EOS, so the first refusal leaves it closed and the next poll ends the stalled track. The pre-check saved one poll, and it armed the drop probe on healthy recordings.

92d01810 removes it. The one poll that picks a finished track now logs at debug instead of warning that the recording stays stalled.

Checked on 92d01810:
- `a_track_that_finished_does_not_hide_a_later_stall` passes without the pre-check.
- It fails ("0 video buffers reached the muxer in 5 s") when a refused track is also reopened while its input carries the EOS. So it guards the reopen rule, not the pre-check; its doc comment and the body's revert table now say so.
- fmt, clippy, full suite: 845 passed, 0 failed, 2 ignored.

With the pre-check gone, `retired` is set only by `end_stalled_track`, as on `main`. The watchdog split out of `recorder.rs` is left for a follow-up.

PR #763, opened 2026-09-05
fix(whip): null a session bin whipserversrc removes mid-teardown

wagenet:
## The bug

A WHIP publisher dropping without a DELETE leaves a GStreamer CRITICAL behind:

```
GStreamer-CRITICAL: Trying to dispose element bin31, but it is in PAUSED instead of the NULL state.
```

`bin31` is not a Strom element. `webrtcsrc` wraps each WHIP session in a plain `GstBin` holding
that session's `webrtcbin`, and ends the session by taking the bin to NULL and removing it from
`whipserversrc` — on its own thread (`State::finalize_session` in gst-plugins-rs `net/webrtc`
0.15.3, dispatched through `RUNTIME.spawn_blocking`). Nothing waits for it: the guard
`unprepare()` blocks on is cleared at the top of that task, before it touches the bin.

Both sides are woken by the same `notify::ice-connection-state` emission, and `webrtcsrc`'s
handler is connected first, so it always gets a head start. From a `GST_DEBUG` trace of the
collision — A is `webrtcsrc`'s finalizer, B is `teardown_session_pipeline`:

```
A  <bin0> completed state change to NULL       session bin down, sockets closed
B  <bin0> current NULL, desired next PAUSED    7 us later
B  <bin0> completed state change to PAUSED     raised back up, everything reopened
A  gst_bin_remove(whipserversrc, bin0)         removed, in PAUSED
B  <whipserversrc> children READY->NULL: iterator done, no children
```

The raise is not incidental. `gst_bin_element_set_state` skips a plain child that is already below
the target state, but recurses into a `GstBin` unconditionally ("always recurse into bins so that
we can set the base time"), so the session bin — and only the session bin — gets pulled back up on
the way down. It then leaves the pipeline before the descent can reach it again, and `bin0`,
`webrtcbin0`, `rtpbin` and `rtpfunnel0` are all dropped without ever running READY→NULL, taking
their UDP sockets and streaming threads with them.

Waiting for `set_state(Null)` to settle does not help: PAUSED is a step on the way down, so the
damage is done before it returns.

## The fix

`element-removed` fires with no bin lock held and the child already unparented, which makes it the
one point where the state can still be corrected. `gst::orphan_guard` locks the removed element's
state and then sets it to NULL. Locking first is what makes it safe under the race:
`gst_bin_element_set_state` reads the locked flag under the child's state lock, so whichever thread
gets there second sees the other's effect — either the parent skips the child outright, or this
call queues behind the raise and takes it back down.

`teardown_session_pipeline` also waits out an ASYNC return now. This is defensive, not a second
fix: no session-pipeline element returns ASYNC on the way down today, so the arm is currently dead.
One that did would have the pipeline dropped mid-transition, leaving the same kind of orphan.

## Verification

Repro: a `whipclientsink` publisher into a headless Strom, then the session ended from both sides
at once — a DELETE to `whipserversrc`'s own signaller racing Strom's. Same collision as the
ICE-failure path, on demand: on loopback, webrtcbin will not report `failed` without consent
freshness, so the ICE trigger itself was not reproduced.

- Before: 3 of 3 runs produced `Trying to dispose element binN, but it is in PAUSED`, plus knock-on
  CRITICALs for `webrtcbin0`, `rtpbin` and `rtpfunnel0`.
- After: 0 of 6 runs.
- Abrupt drop (SIGKILL the publisher, watchdog reaps the session) is clean, and the guard's debug
  line confirms it is wired to the real `whipserversrc`:
  `Orphan guard: 'bin0' was removed from 'whip1:whipserversrc_65d8...' at NULL`.

## Test

`test_removed_child_is_driven_to_null` in `backend/tests/pipeline_lifecycle_test.rs` stands the same
shape up with core elements and replays both halves of the race in a fixed order: a removal that
lands while the child is still running, then a parent-driven descent over a child that has already
gone to NULL. Reverting the guard fails the first assertion; reverting only `set_locked_state` fails
the second. Both verified.

It does not guard the call site. The test calls `orphan_guard::install` itself, so deleting the
`install(&bin)` line from `whip.rs` leaves it green. A `whipserversrc`-based test cannot close that
gap here — CI installs base/good/bad/ugly/libav and no gst-plugins-rs (see the `packages:` list in
`.github/workflows/ci.yml`), so it would skip green and guard nothing. The wiring is covered only
by the manual run above.

No CI package changes: the test uses `fakesrc`/`fakesink` from coreelements, which the file's
existing tests already rely on.

Ran locally: `cargo test --package strom` (all green, 4/4 in `pipeline_lifecycle_test`),
`cargo clippy --all-targets`, `cargo fmt --check`.

## WHEP output does not need the same guard

`whepserversink` is a `BaseWebRTCSink` subclass and does remove children from itself, but neither
half of the race is present. Each viewer's `webrtcbin` lives in a `gst::Pipeline` the sink owns
privately, not as a child of Strom's pipeline, so a disconnect removes nothing from the hierarchy
being torn down. The children it does remove from itself — `clocksync` and `appsink`, in
`InputStream::unprepare` — come out inside `change_state` on `PausedToReady`, on the thread already
driving the descent, so the cross-thread window this bug needs never opens. Its tokio branch defers
to `call_async` but blocks on the result, so that path closes before `change_state` returns. Read
in gst-plugin-webrtc 0.15.2, the version in `Cargo.lock`.

## Upstream

A workaround, not a fix: Strom cannot make the removal wait.

### DIFF
--- a/backend/src/blocks/builtin/whip.rs
+++ b/backend/src/blocks/builtin/whip.rs
@@ -16,6 +16,7 @@ use crate::blocks::{
 };
 use crate::gst::ice_preflight;
 use crate::gst::keyframe_request;
+use crate::gst::orphan_guard;
 use crate::whip_session_manager::{SessionCleanupRequest, WhipEndpointConfig};
 use gstreamer as gst;
 use gstreamer::prelude::*;
@@ -620,6 +621,11 @@ pub fn create_whipserversrc_for_session(
     let cleanup_tx_for_ice = cleanup_tx.clone();
 
     if let Ok(bin) = whipserversrc.clone().downcast::<gst::Bin>() {
+        // whipserversrc removes each session's internal bin from a thread of its
+        // own, which can land while this pipeline is on its way down and leave
+        // the bin orphaned above NULL with its WebRTC sockets still open.
+        orphan_guard::install(&bin);
+
         bin.connect("deep-element-added", false, move |values| {
             let element = values[2].get::<gst::Element>().unwrap();
             let element_name = element.name();
--- a/backend/src/gst/mod.rs
+++ b/backend/src/gst/mod.rs
@@ -8,6 +8,7 @@ pub mod discovery;
 pub mod gl_bridge;
 pub mod ice_preflight;
 pub mod keyframe_request;
+pub mod orphan_guard;
 pub mod pipeline;
 pub mod pipeline_monitor;
 pub mod shaders;
--- a/backend/src/gst/orphan_guard.rs
+++ b/backend/src/gst/orphan_guard.rs
@@ -1,0 +1,95 @@
+//! Drive to NULL any element a bin removes from itself while it is still running.
+//!
+//! An element disposed above NULL never ran its READY→NULL transition, so the
+//! UDP sockets, streaming threads and file descriptors it opened on the way up
+//! are still open when the last reference goes:
+//!
+//! ```text
+//! GStreamer-CRITICAL: Trying to dispose element bin31, but it is in PAUSED
+//! instead of the NULL state.
+//! ```
+//!
+//! A bin drives its children to whatever state it is driven to and only lets go
+//! of one at NULL, so the only way in is a child removed from a different
+//! thread than the one running the parent's state change.
+//!
+//! `whipserversrc` does exactly that. Each WHIP session lives in a plain
+//! `GstBin` holding the session's `webrtcbin`, and `webrtcsrc` ends a session by
+//! taking that bin to NULL and removing it — on its own thread
+//! (`State::finalize_session` in gst-plugins-rs `net/webrtc`, via
+//! `RUNTIME.spawn_blocking`). Nothing waits for it: the guard `unprepare()`
+//! blocks on is cleared at the top of that task, before it touches the bin.
+//!
+//! Strom's abrupt-drop teardown runs into it because both are woken by the same
+//! `notify::ice-connection-state` emission, and `webrtcsrc`'s handler is
+//! connected first. From a GST_DEBUG trace of the collision — A is
+//! `webrtcsrc`'s finalizer, B is `teardown_session_pipeline`:
+//!
+//! ```text
+//! A  <bin0> completed state change to NULL       session bin down, sockets closed
+//! B  <bin0> current NULL, desired next PAUSED    7 us later: the pipeline's
+//! B  <bin0> completed state change to PAUSED     PLAYING->PAUSED step raises it
+//!                                                back up, reopening everything
+//! A  gst_bin_remove(whipserversrc, bin0)         removed, in PAUSED
+//! B  <whipserversrc> children READY->NULL: iterator done, no children
+//! ```
+//!
+//! A plain child at NULL is left alone by a downward transition, but
+//! `gst_bin_element_set_state` recurses into a `GstBin` unconditionally ("always
+//! recurse into bins so that we can set the base time"), which is why it is the
+//! session bin, and not the elements inside it, that gets raised.
+//!
+//! The bin leaves the pipeline before the descent can reach it again, so nothing
+//! takes it back to NULL. Waiting for `set_state(NULL)` to settle does not help:
+//! PAUSED is a step on the way down, so the raise has already happened by the
+//! time it returns.
+//!
+//! Strom cannot make the removal wait, so it catches the element on the way out.
+//! `element-removed` fires with no bin lock held and the child already
+//! unparented — the one point where the state can still be fixed.
+
+use gstreamer as gst;
+use gstreamer::prelude::*;
+use tracing::{debug, warn};
+
+/// Guard every element `bin` removes from itself.
+///
+/// Locking the state before setting it is what makes this safe against a
+/// removal that lands *during* the parent's transition. `gst_bin_element_set_state`
+/// reads the locked flag under the child's state lock, so whichever thread gets
+/// there second sees the effect of the first: either the parent skips the child
+/// outright, or this call queues behind the raise and takes it back down.
+pub fn install(bin: &gst::Bin) {
+    // Captures nothing. Both the bin and the removed element arrive as
+    // arguments, so this handler cannot form the reference cycle that a
+    // captured `gst::Element`/`gst::Bin` clone would.
+    bin.connect_element_removed(|bin, element| {
+        // Read for the log only. It can still say NULL while the parent's
+        // cascade holds the state lock mid-raise, so it is never a reason to
+        // skip the work below.
+        let state = element.current_state();
+        if state == gst::State::Null {
+            debug!(
+                "Orphan guard: '{}' was removed from '{}' at NULL",
+                element.name(),
+                bin.name()
+            );
+        } else {
+            warn!(
+                "Orphan guard: '{}' was removed from '{}' in {:?}, forcing it to NULL",
+                element.name(),
+                bin.name(),
+                state
+            );
+        }
+
+        element.set_locked_state(true);
+        if let Err(e) = element.set_state(gst::State::Null) {
+            warn!(
+                "Orphan guard: failed to set '{}' to NULL after removal: {:?}",
+                element.name(),
+                e
+            );
+        }
+    });
+}
--- a/backend/src/gst/orphan_guard.rs
+++ b/backend/src/gst/orphan_guard.rs
@@ -22,17 +22,7 @@
 //!
 //! Strom's abrupt-drop teardown runs into it because both are woken by the same
 //! `notify::ice-connection-state` emission, and `webrtcsrc`'s handler is
-//! connected first. From a GST_DEBUG trace of the collision — A is
-//! `webrtcsrc`'s finalizer, B is `teardown_session_pipeline`:
-//!
-//! ```text
-//! A  <bin0> completed state change to NULL       session bin down, sockets closed
-//! B  <bin0> current NULL, desired next PAUSED    7 us later: the pipeline's
-//! B  <bin0> completed state change to PAUSED     PLAYING->PAUSED step raises it
-//!                                                back up, reopening everything
-//! A  gst_bin_remove(whipserversrc, bin0)         removed, in PAUSED
-//! B  <whipserversrc> children READY->NULL: iterator done, no children
-//! ```
+//! connected first, so the removal lands in the middle of the pipeline's descent.
 //!
 //! A plain child at NULL is left alone by a downward transition, but
 //! `gst_bin_element_set_state` recurses into a `GstBin` unconditionally ("always
--- a/backend/src/whip_session_manager.rs
+++ b/backend/src/whip_session_manager.rs
@@ -170,6 +170,10 @@ pub struct WhipSessionManager {
 /// sub-second in practice.
 const PENDING_CLEANUP_TTL: Duration = Duration::from_secs(30);
 
+/// How long to wait for a session pipeline that returns ASYNC from its
+/// transition to NULL. Bounded because the wait runs on a blocking thread.
+const TEARDOWN_TIMEOUT: gst::ClockTime = gst::ClockTime::from_seconds(5);
+
 impl WhipSessionManager {
     pub fn new() -> Self {
         let (cleanup_tx, cleanup_rx) = mpsc::unbounded_channel();
@@ -472,11 +476,25 @@ impl WhipSessionManager {
             name
         );
 
-        if let Err(e) = session_pipeline.set_state(gst::State::Null) {
-            warn!(
+        match session_pipeline.set_state(gst::State::Null) {
+            // Nothing in a session pipeline goes async on the way down today,
+            // but an element that did would have the pipeline dropped out from
+            // under a transition still in flight, leaving its children to be
+            // disposed above NULL.
+            Ok(gst::StateChangeSuccess::Async) => {
+                let (result, current, _) = session_pipeline.state(TEARDOWN_TIMEOUT);
+                if result.is_err() || current != gst::State::Null {
+                    warn!(
+                        "WhipSessionManager: Session pipeline {} did not reach Null within {:?} (current: {:?})",
+                        name, TEARDOWN_TIMEOUT, current
+                    );
+                }
+            }
+            Ok(_) => {}
+            Err(e) => warn!(
                 "WhipSessionManager: Failed to set session pipeline {} to Null: {:?}",
                 name, e
-            );
+            ),
         }
     }
 }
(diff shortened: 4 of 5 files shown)


srperens (commented):
**Verdict: Comment** — right layer, and the mechanism survives reading. Two gaps: the new test guards the module but not the wiring, and the guard is installed on one of the two webrtc bins in this repo that remove children from their own threads.

**Requested changes**

1. State in the PR body that `test_removed_child_is_driven_to_null` cannot guard the call site. Deleting `orphan_guard::install(&bin)` from `whip.rs` leaves the test green, and a `whipserversrc`-based test cannot run here because CI installs no gst-plugins-rs (`.github/workflows/ci.yml:85-95` — the `packages:` list is base/good/bad/ugly/libav only). CLAUDE.md requires that gap named rather than left implicit; the fakesrc/fakesink stand-in is the right call, it just needs saying.
2. Say whether WHEP output needs the same guard, or why not. `backend/src/blocks/builtin/whep.rs:946` — `` `/// whepserversink is based on webrtcsink and handles encoding internally.` `` — that sink removes a consumer bin per disconnect on its own thread — the same shape — and nothing guards it.

**Claims**

|---|---|---|
| The handler cannot form the reference cycle CLAUDE.md forbids | `CONFIRMED` | `backend/src/gst/orphan_guard.rs:66` — `` `    bin.connect_element_removed(|bin, element| {` `` — both objects arrive as arguments and the closure captures nothing |
| The bounded wait cannot stall the tokio runtime | `CONFIRMED` | all ten callers of `teardown_session_pipeline` are on a blocking thread: `backend/src/api/whip_ingest.rs:705` — `` `    let _ = tokio::task::spawn_blocking(move || {` `` and `backend/src/state.rs:1429` — `` `                let _ = tokio::task::spawn_blocking(move || {` `` |
| The new test executed, rather than passing green | `CONFIRMED` | `Check (Linux)` `success` at `0f8f91f4` runs `.github/workflows/ci.yml:129` — `` `        run: cargo test --package strom --features efp,nvidia` ``; the test has no skip guard and uses core elements |
| `element-removed` fires unlocked and unparented, and a removed child is never re-added by the plugin | `EXTERNAL` | upstream `gst_bin_remove_func` and `webrtcsrc`. Assumption stated: `backend/src/gst/orphan_guard.rs:86` — `` `        element.set_locked_state(true);` `` is never cleared, so a child the plugin removed and re-added would stay frozen at NULL. Strom adds nothing into that bin, so only the plugin can trigger it |

**Diagnosis** — The cause explains the symptom and the fix is at the only point that can act: the raise happens during the parent's own descent, so waiting on `set_state(NULL)` genuinely cannot help, and `element-removed` is the last moment the child is reachable. Locking before setting is what makes the two orderings both correct rather than one of them. Coverage is `BOUNDED` to bins that `install()` is called on — today exactly one, in `create_whipserversrc_for_session`.

The `StateChangeSuccess::Async` arm is dead on arrival by the PR's own account, and I could not find a session-pipeline element that returns ASYNC downward either. It is harmless, but a line in the body saying it is defensive would stop a later reader treating it as a fixed bug.

**Radius** — `SHARED`, argued: `teardown_session_pipeline` changes for all ten call sites across `whip_session_manager.rs`, `api/whip_ingest.rs` and `state.rs`, including flow stop, though only in the arm above. The guard itself is confined to the WHIP input path. A regression would show up as a WHIP session that no longer tears down, i.e. a slot that never frees.

**Tests & CI** — green at head `0f8f91f4` (`Check (Linux)`, both Linux builds, WASM, API contract). `Build (macOS)`/`Build (Windows)` `skipping`; nothing here is platform-specific. `pipeline_lifecycle_test.rs` is the right canary and the new case lives in it.

Confidence: HIGH

wagenet:
Claude here, replying on Peter's behalf. All three addressed in the PR body.

**1. The test cannot guard the call site — now stated.** Confirmed: `pipeline_lifecycle_test.rs:279` calls `orphan_guard::install` directly, so deleting `install(&bin)` at `whip.rs:627` leaves the suite green. The body previously gave the wrong reason for not closing the gap (it claimed a real test needs a live WebRTC peer). The actual blocker is the one you named: the `packages:` list is base/good/bad/ugly/libav with no gst-plugins-rs, so a `whipserversrc`-based test would skip green and guard nothing. Reworded to say that, and to say the wiring is covered only by the manual run.

**2. WHEP output does not need the same guard.** Read against gst-plugin-webrtc 0.15.2, the version in `Cargo.lock`. `whepserversink` is `WhepWebRTCSink`, a `BaseWebRTCSink` subclass, so you are right that it is the same general shape — but neither half of the race is present:

- Each viewer's `webrtcbin` lives in a `gst::Pipeline` the sink owns privately (`SessionInner::pipeline`, `webrtcsink/imp.rs:312`), not as a child of Strom's pipeline. A disconnect removes nothing from the hierarchy being torn down.
- The children it does remove from itself — `clocksync` and `appsink` — come out in `InputStream::unprepare` (`webrtcsink/imp.rs:1616`), called from `change_state` on `PausedToReady` after `parent_change_state` returns (`imp.rs:5959`). That is the thread already driving the descent, which is the precondition this bug needs. Both are set to NULL on that same thread immediately after removal.
- The one path that leaves that thread is the tokio-context branch at `imp.rs:5961`, which defers `unprepare` to `call_async` and then blocks on `rx.recv()`, so it still completes before `change_state` returns.

**3. The ASYNC arm now reads as defensive.** Agreed it was ambiguous. The body now says outright that no session-pipeline element returns ASYNC on the way down today, so the arm is currently dead, and that it is there because a pipeline dropped mid-transition would leave the same kind of orphan.

On your `EXTERNAL` claim: agreed that `set_locked_state(true)` is never cleared, so a child the plugin removed and re-added would stay frozen at NULL. Strom adds nothing into that bin, so only the plugin could trigger it, and `webrtcsrc` does not re-add a removed session bin.

Body-only changes, no new commit. The one code commit since your review is `89dac64`, which trims a debug trace out of the `orphan_guard` module doc and touches no logic.

wagenet:
@srperens thanks for taking a look! Let me know if you need anything else.

srperens (dismissed):
**Verdict: Approve** — supersedes my review at `0f8f91f4`. Both requested changes are answered in the body, the head moved by a documentation edit only, and CI is green on it. Nothing new to ask for.

**What changed since `0f8f91f4`** — I read the inter-head diff in full: one hunk, removing the `GST_DEBUG` collision trace from the module doc, leaving `backend/src/gst/orphan_guard.rs:25` — `//! connected first, so the removal lands in the middle of the pipeline's descent.` No code, no test, no behaviour. The trace survives in the PR body, which is the right place for a one-off observation that cannot be kept true by anything; the module doc keeps the part that stays true. The rows I verdicted at `0f8f91f4` are untouched by that hunk and stand as verdicted there — the no-capture closure, the ten blocking call sites, and the upstream `element-removed` assumption are not re-argued here.

**Both requested changes, closed**

1. The guard gap is now stated. The body says the test calls `orphan_guard::install` directly, so deleting `backend/src/blocks/builtin/whip.rs:627` — `        orphan_guard::install(&bin);` leaves it green, and it names why a `whipserversrc` test cannot close the gap here. That is what CLAUDE.md's Tests section asks for: a demonstration declared as one rather than presented as a guard.
2. WHEP is answered, with a mechanism rather than an assertion — neither half of the race is present, because each viewer's `webrtcbin` lives in a pipeline the sink owns privately, and the children it does remove from itself come out inside `change_state` on the descending thread. Read against `Cargo.lock:2663` — `version = "0.15.2"`, the version this repo actually builds.

**Claims**

|---|---|---|
| The head move is documentation only | `CONFIRMED` | `git diff 0f8f91f4 89dac64d` is a single hunk in `backend/src/gst/orphan_guard.rs`, all comment lines |
| The covering test still exists and still executes | `CONFIRMED` | `backend/tests/pipeline_lifecycle_test.rs:257` — `fn test_removed_child_is_driven_to_null() {`, unchanged across the head move, no skip guard, core elements only; `Check (Linux)` `success` in run `34427313024` at `89dac64d` |
| CI is green at the current head | `CONFIRMED` | run `34427313024`, `head_sha` `89dac64d`: `Check (Linux)`, `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check & Build (WASM)`, `API Contract Check` all `pass` |
| `whepserversink` needs no equivalent guard | `EXTERNAL` | gst-plugin-webrtc 0.15.2, not settleable from this repository. Assumption stated: the per-viewer `webrtcbin` is never parented into Strom's pipeline, so a disconnect removes nothing from the hierarchy being torn down. If a future version parents it, this conclusion lapses |

**Coverage** — `BOUNDED`, unchanged and correctly so: the guard reaches bins `install()` is called on, today the single one in `create_whipserversrc_for_session`. The body now names the boundary instead of leaving a reader to find it.

**Radius** — `SHARED`, as argued in my earlier review: `teardown_session_pipeline` changed for all ten of its call sites, though only in the ASYNC arm, which the body correctly labels defensive and currently dead. A regression surfaces as a WHIP slot that never frees.

**Tests & CI** — green at `89dac64d`, listed above. `Build (macOS)` and `Build (Windows)` `skipping`; nothing here is platform-specific, so Linux-green is the right bar. I did not build or run anything locally — this runner has neither cargo nor GStreamer, and every statement above is read off the concluded run.

Confidence: HIGH

srperens:
Reviewed this and it is close. The orphan guard itself is sound — I verified the parts that are easy to get wrong:

- The `connect_element_removed` closure captures nothing; both the bin and the element arrive as arguments, so it cannot form the reference cycle a captured `gst::Element`/`gst::Bin` clone would.
- `set_locked_state(true)` before `set_state(NULL)` is the load-bearing half, and the reasoning about `gst_bin_element_set_state` recursing into bins unconditionally matches how it behaves.
- All eight call sites of `teardown_session_pipeline` (`whip_session_manager.rs` ×2, `state.rs` ×1, `whip_ingest.rs` ×5) are inside `spawn_blocking` or `std::thread::spawn`, so the new bounded 5s wait cannot stall a tokio worker. The comment's claim holds.

**I pushed a merge of main to this branch** — `whip.rs` and `whip_session_manager.rs` had both moved (#791 among others) and it was CONFLICTING. Both conflicts were pure additions on either side and I kept everything: this branch's `orphan_guard` import and `TEARDOWN_TIMEOUT` alongside main's `rtp_hdrext`, `SessionActivity`, `TAKEOVER_*` and `IdlestSession`. After the merge, locally: `pipeline_lifecycle_test` 4/4 including `test_removed_child_is_driven_to_null`, whip suite 26/26, clippy clean for native and wasm32. Shout if you would rather have resolved it yourself.

Two things I would like before this goes in:

**1. Split out the ASYNC teardown change, or put it in the title.** The `Ok(StateChangeSuccess::Async)` arm in `whip_session_manager.rs` is a second, unrelated fix. It has no test coverage, and its own comment says the case does not occur today ("Nothing in a session pipeline goes async on the way down today"). It is reasonable defensive code, but it is not what the title describes.

**2. State the test gap in the PR body.** `test_removed_child_is_driven_to_null` calls `orphan_guard::install()` directly, so it does guard the module — good. But it stands the shape up with `fakesrc`/`fakesink` rather than `whipserversrc`, which means **deleting the `orphan_guard::install(&bin)` call from `whip.rs` leaves the test passing**. The fix is guarded; the wiring is not. Not using a real `whipserversrc` in CI is the right call, so this is about saying so, not changing the test.

And one question rather than a request: `install()` locks *every* element ever removed from the whipserversrc bin at NULL, permanently. That is correct for session bins being destroyed. Do you know whether `whipserversrc` ever removes and reuses an element internally? Nothing suggests it does today, but the guard is unconditional.

srperens (approved):
**Verdict: Approve** — the mechanism is right, the guard is installed where the removal actually happens, and the regression test is one of the few in this repo that provably executed in CI rather than skipping green. One follow-up below is a separate issue, not a change requested here.

**Claims**

|---|---|---|
| The handler cannot form the reference cycle CLAUDE.md forbids | `CONFIRMED` | `backend/src/gst/orphan_guard.rs:56` — `bin.connect_element_removed(|bin, element| {` — not a `move` closure, and both objects arrive as arguments, so no `Bin`/`Element` clone is captured |
| The new test executed in this repository's CI at this head | `CONFIRMED` | `Check (Linux)` job 104874989591 of run 35119967573, whose `head_sha` is `e568787c802388bd97004bb5a0183f3bd1afd290`: `test test_removed_child_is_driven_to_null ... ok` |
| The guard cannot disturb the slot decodebins that already use locked state | `CONFIRMED` | `backend/src/blocks/builtin/whip.rs:275` — `    decodebin.set_locked_state(true);` — those chains are built in the main pipeline, not as children of `whipserversrc`, and `connect_element_removed` fires only for direct children |
| `whepclientsrc` is the same element family and gets no guard | `EXTERNAL` | `backend/src/blocks/builtin/whep.rs:624` — `    // not configured, since webrtcsrc defaults to stun://stun.l.google.com:19302)` — this repo's own comment calls it a `webrtcsrc` derivative; `backend/src/blocks/builtin/whip.rs:824` is the only `install()` site |

**Diagnosis** — The stated cause explains the symptom without hand-waving: a `GstBin` child is recursed into unconditionally on a downward transition, so it is raised to PAUSED on the way down and then leaves the hierarchy before the descent returns. `element-removed` is the right hook because it is the last point at which the child is still reachable and no longer parented, and `set_locked_state(true)` before `set_state(Null)` is what makes it ordering-independent rather than a narrower race. Coverage is `BOUNDED` to `whipserversrc`.

**Follow-up (not blocking)** — WHEP Input builds `whepclientsrc` at `backend/src/blocks/builtin/whep.rs:618` — `    let whepclientsrc = gst::ElementFactory::make("whepclientsrc")` — and gets no guard. If it shares `BaseWebRTCSrc`'s session finalizer, the same race exists there, and with one session ending exactly as the flow stops it is arguably likelier to fire. I could not settle that from this repository; it needs a read of the pinned gst-plugins-rs. Worth an issue either way. The PR body's analysis of `whepserversink` is sound and does not cover this element.

**Radius** — `SHARED`: every WHIP Input flow, every session. Confined to direct children of `whipserversrc`, and nothing in Strom adds or removes those — only `webrtcsrc`'s internals do.

**Tests & CI** — Green at this head; `Build (Windows)`/`Build (macOS)` show `skipping`, which is correct here because the diff contains no platform-specific code. The `gst_mini_object_unref` CRITICALs in this test binary are pre-existing: they appear in `pipeline_lifecycle_test` on main at `0d9d469c` (run 35210402565) without this change. The PR is straight about the one real gap — the test calls `orphan_guard::install` itself, so deleting the `backend/src/blocks/builtin/whip.rs:824` call site leaves it green — and CLAUDE.md asks exactly that it be said rather than closed.

**Assumption stated** — that `element-removed` is emitted with no bin lock held is an upstream GStreamer property this repository cannot verify; the fix's safety argument rests on it.

Confidence: HIGH

PR #765, opened 2026-09-07
chore(mcp): drop the stdio MCP server and the dead demo script

srperens:
## Why

The standalone `strom-mcp-server` binary is a stdio-to-REST proxy, written when MCP clients only spoke stdio. That constraint is gone: Claude Code takes `--transport http`, Claude Desktop has custom connectors, and anything that still needs stdio can bridge with `mcp-remote`. The backend's own `/api/mcp` is a strict superset — same twelve tools, direct `AppState` access instead of an HTTP hop, plus sessions and SSE.

Nobody was using it. `demo-mcp.sh` was the only thing that ever exercised the stdio path, and it has been dead since #49 renamed the backend crate, eleven days after the script landed:

```
$ cargo build --release -p strom-backend
error: package ID specification `strom-backend` did not match any packages
```

With `set -e` that kills the script on its first real command — and the binary check above it tests `./target/release/strom-backend`, a path that can never exist, so the build branch is always taken. It also polls `localhost:3000` while `DEFAULT_PORT` is 8080. Nine months, no reports. The script is referenced from nowhere and is not in CI.

Keeping the crate cost more than it earned: the twelve tool definitions were duplicated verbatim in both implementations and had already drifted — the stdio copy advertises protocol `2024-11-05` against the backend's `2025-03-26` and never implemented `ping` — with no test on either side to catch it. `cargo test --package strom-mcp-server` ran in CI on four platforms and executed zero tests.

## What this removes

- `mcp-server/` (the crate, ~850 lines) and the workspace member
- `demo-mcp.sh`
- 4 CI clippy/test/build steps and 4 artifact uploads, across Linux x86_64/ARM64, Windows and macOS
- 4 release assets and the MSI download step
- the `StromMcpServerExe` MSI component and its `Build-Installer.ps1` parameter
- the second binary in the Docker image
- the `INSTALL_MCP_SERVER` env var and the binary-choice item in `install.sh`'s menu (remaining items renumbered)
- the extra build step in both cross-compile scripts

Docs now describe the HTTP endpoint only, with an `mcp-remote` recipe for stdio-only clients. `docs/MCP.md` picks up the "code is the source of truth" disclaimer while it is being rewritten.

## Testing

- `cargo check --workspace` ✅
- `cargo clippy --workspace --all-targets --features efp -- -D warnings` ✅
- `cargo test --test openapi_test` ✅ (the MCP endpoints stay in the schema, unchanged)
- `git grep` finds no remaining reference to the binary, the crate, or the demo script

Not run: the Windows MSI build and the ARM64 cross-compile — no host for either. Both changes are deletions of a component from a working build, and the WiX component is removed from both its declaration and its `ComponentRef`.

## Follow-up

The HTTP endpoint has its own bugs — MCP is unreachable on password-protected instances, notifications get error responses, sessions leak. Those are #766, stacked on this branch. This PR is deletion only.


srperens (commented):
**Verdict: Approve** — a clean, self-contained deletion. The dead-code claims check out, no reference to the removed crate/binary/script survives anywhere in the tree, and Linux CI (the only platform this touches functionally) is green. Filed as a `COMMENT`-state review, not `APPROVED`: this PR's author is this same account, and GitHub's `addPullRequestReview` refuses self-approval outright. The marker still records `verdict=Approve` because that is the actual conclusion; a maintainer's own approving review is what should gate the merge.

**Claims**

|---|---|---|
| `demo-mcp.sh` was already broken (wrong package name) since #49 renamed the crate | `CONFIRMED` | `demo-mcp.sh:11` — `if [ ! -f "./target/release/strom-backend" ] || [ ! -f "./target/release/strom-mcp-server" ]; then` and `demo-mcp.sh:13` — `cargo build --release -p strom-backend -p strom-mcp-server`, both naming a package (`strom-backend`) `5035589` (#49) renamed to `strom` |
| The script also polls the wrong port | `CONFIRMED` | `demo-mcp.sh:27` — `if curl -s http://localhost:3000/health > /dev/null 2>&1; then` vs `types/src/lib.rs:7` — `pub const DEFAULT_PORT: u16 = 8080;` |
| `strom-mcp-server` had zero tests despite CI running `cargo test --package strom-mcp-server` on 4 platforms | `CONFIRMED` | `mcp-server/src/main.rs` and `mcp-server/src/client.rs` at the pre-deletion head contain no `#[test]`/`#[cfg(test)]` attribute (`git grep` returns none) |
| No remaining reference to the crate, binary or demo script anywhere in the tree | `CONFIRMED` | `git grep -i 'mcp-server\|strom-mcp-server\|demo-mcp'` at this PR's head returns nothing outside `docs/MCP.md`, which now documents the HTTP endpoint only |

**Diagnosis** — Pure subtraction: a duplicated, drifted, untested implementation of a capability the backend's own `/api/mcp` already provides as a superset. Every CI step, installer component, and doc reference tied to the removed crate is deleted in the same diff (`.github/workflows/*.yml`, `Dockerfile`, `install.sh`, the WiX installer, both cross-compile scripts) — nothing is left half-migrated.

**Radius** — `LOCAL`: `Cargo.lock` drops exactly the `strom-mcp-server` package entry and nothing else (`Cargo.lock` diff, -14 lines, no other package's dependency set changes), confirming no other crate depended on it.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` all pass at `029d6c2`. `Build (Windows)`/`Build (macOS)` are `skipping`, which is this repo's normal PR-trigger behavior (manual dispatch only), not evidence either way — the Windows/macOS-relevant edits here (installer WiX, `Build-Installer.ps1`) are config/script deletions with no compiled code behind them, so the risk is low but formally `EXTERNAL` (unexercised by this repo's CI).

Confidence: HIGH

PR #766, opened 2026-09-07
fix(mcp): make the HTTP endpoint usable and correct, with tests

srperens:
> Stacked on #765 — review that one first. This PR's diff against `main` will include #765's deletions until it merges.

Six defects in `/api/mcp`, found by driving it with a real MCP client and by hand. None were guarded by a test, because there were none.

## 1. MCP was unreachable on any password-protected instance

Authentication is enabled as soon as an admin user *or* an API key is configured (`enabled = admin_user.is_some() || api_key.is_some()`), but the endpoint only ever accepted an API key. So an instance with `STROM_ADMIN_USER` + hash and no `STROM_API_KEY` — the setup `OPEN_LIVE_SETUP.md` describes — answered 401 to every MCP request, with no credential that could work. Measured before the fix:

```
auth methods: {"authenticated":false,"auth_required":true,"methods":["session"]}
POST /api/mcp  (no credentials)        → 401
POST /api/mcp  (valid session cookie)  → 401   ← the cookie that works everywhere else
GET  /api/flows (same cookie)          → 200
```

The endpoint now accepts what the rest of the API accepts: `X-API-Key`, `Authorization: Bearer` (API key or native GUI token), or the login session cookie. The MCP routes sit in the public router and keep their own JSON error shape, so this reuses one `session_is_authenticated` helper rather than the middleware — the same helper now backs `auth_middleware` and `auth_status_handler`, which had the check written out twice.

## 2. `notifications/initialized` got an error response

The handler matched `"initialized"`, which is not the method any client sends. The real notification fell through to "Method not found" — i.e. a JSON-RPC *response*, carrying a null id, to a message that must not be answered at all:

```
POST {"jsonrpc":"2.0","method":"notifications/initialized"}
→ 200 {"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found: notifications/initialized"}}
```

Notifications now return `202` with no body. Everything under `notifications/*` is swallowed, which also covers notifications this server has not heard of yet, and an id-less unknown method is treated as a notification rather than answered. The old non-standard `"initialized"` spelling stays accepted.

## 3. Sessions leaked for the life of the process

`cleanup_stale` existed and was called from nowhere. Clients are not required to send `DELETE` and generally do not — Claude Code's HTTP transport never did in any of my runs. Measured: five sessions created, one terminated, across a few minutes of ordinary use, each holding a 100-slot broadcast channel forever.

Replaced with idle-based collection: a background sweep every 5 minutes drops sessions idle for more than an hour, `last_seen` is refreshed on each request, and a session with a live SSE subscriber is never collected however idle it looks (a client sitting on the stream sends nothing to refresh itself).

## 4. The SSE stream forwarded the whole event bus

The filter skipped `SystemStats`, `MeterData` and `Ping`, then let a generic arm re-encode *everything else* as a JSON string nested inside JSON. That included `LoudnessData`, `SpectrumData`, `LatencyData`, `QoSStats`, `AudioAnalyzerData`, `MediaPlayerPosition`, `ThreadStats`, `PtpStats` and `BufferAgeProbe` — many per second per flow, burying the flow-lifecycle notifications a client actually reads. Only the mapped notifications are sent now; `WS /api/ws` remains for the firehose.

## 5. Tool failures arrived as transport errors

A flow that does not exist came back as `-32603`, so a model saw a broken connection rather than a reason it could act on. Now split the way the spec splits it — `isError: true` results for tools that ran and failed, `-32602` for calls that were malformed — with messages to match:

| | before | after |
|---|---|---|
| missing flow | `-32603 Tool call failed: Flow not found: …` | `isError` result, text `Flow not found: …` |
| bad UUID | ``-32603 Tool call failed: invalid character: found `n` at 0`` | `-32602 flow_id must be a UUID (got 'not-a-uuid') - use list_flows to find valid ids` |
| unknown tool | `-32603 Tool call failed: Unknown tool: nope` | `-32602 Unknown tool: nope - call tools/list for the available tools` |

## 6. Origin validation guarded the wrong half

Only localhost origins passed, so a browser client on the instance's own hostname got 403 while every non-browser caller sailed through unchecked. Now same-origin: this host (matched against `Host`) or loopback passes, cross-site is refused. IPv6 literals keep their brackets when the port is stripped.

Also drops five never-called methods from the session manager (`get_session`, `send_event`, `mark_initialized`, `is_initialized`, and the `initialized` field they wrote).

## Testing

`backend/tests/mcp_endpoint_test.rs` drives the endpoint through the real router — 10 tests: handshake, notifications, ping, tool set (asserting names and that every tool has a description and object schema), a create/list/delete round trip, both error shapes, unknown method, session lifecycle, cross-site origin. Plus unit tests for the origin and auth predicates (including the password-only config that used to 401) and three for the session sweep.

**These fail if the fixes are reverted.** Verified directly for the notification fix — restoring the old `"initialized"` arm gives:

```
assertion `left == right` failed: "notifications/initialized" should be accepted with 202,
got 200 OK and body {"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found: …"}}
  left: 200
 right: 202
```

The auth test cannot compile against the old signature (which took no session argument), and the `isError` test asserts `error` is null where the old code always set it.

Ran: the new tests ✅, `api_tests` ✅, `openapi_test` ✅, `cargo clippy --workspace --all-targets --features efp -- -D warnings` ✅. CI picks the new file up automatically via `cargo test --package strom`.

Also confirmed live against a password-only instance: 401 without credentials, 200 with the cookie, 202 for the notification, `tools/call list_flows` returning a result, same-host origin 200, cross-site 403, and the cleanup task logging at startup. `claude mcp list` reports the endpoint as `✔ Connected` and the backend log shows the client's SSE stream opening with no method-not-found errors.

Not changed: the advertised protocol version stays `2025-03-26`, and the server still answers `application/json` to every POST rather than streaming. Both are worth revisiting, neither blocks a client today.


srperens:
**CI note:** this PR shows no checks because `ci.yml` only triggers on `pull_request: branches: [main]`, and this one targets the stack base. I dispatched the workflow manually on the branch instead: [run 34128316705](https://github.com/Eyevinn/strom/actions/runs/34128316705) — success (Check (Linux), WASM, both Linux builds; macOS/Windows dispatch-gated as usual).

Confirming the new tests actually executed there rather than just riding a green run — from the `Check (Linux)` log:

```
     Running tests/mcp_endpoint_test.rs

running 10 tests
test a_malformed_call_returns_an_actionable_protocol_error ... ok
test a_failing_tool_returns_an_is_error_result ... ok
test a_cross_site_origin_is_refused ... ok
test a_round_trip_creates_and_deletes_a_flow ... ok
test an_unknown_request_method_is_a_method_not_found_error ... ok
test an_unknown_session_is_rejected_and_delete_ends_a_known_one ... ok
test initialize_assigns_a_session ... ok
test notifications_are_accepted_without_a_response ... ok
test tools_list_returns_the_tool_set ... ok
test ping_is_answered ... ok

test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.78s
```

plus `api::mcp::tests::a_logged_in_session_is_accepted_without_an_api_key ... ok` and `mcp::session::tests::idle_sessions_are_collected_and_touch_saves_them ... ok` in the lib test run.

Once #765 merges, retarget this to `main` so it picks up CI on its own.

srperens (commented):
**Verdict: Comment** — six genuine defects, each fixed at the right layer with a regression test that drives the real router (`backend/tests/mcp_endpoint_test.rs`) and would fail if reverted. Withheld only because CI evidence is incomplete for reasons specific to this being stacked on #765, not because of anything wrong with the diff. Filed as `COMMENT`; self-approval is blocked for this account regardless (see my review of #765).

**Claims**

|---|---|---|
| Password-only auth (admin user, no API key) can now reach MCP via the session cookie | `CONFIRMED` | `backend/src/api/mcp.rs:119` — `session_authenticated: bool,` and `:147` — `if session_authenticated {` — `validate_mcp_auth` now takes and checks it, wired from `crate::auth::session_is_authenticated` in `mcp_post`/`mcp_get`/`mcp_delete` |
| Same-origin (not just localhost) now passes the DNS-rebinding check | `CONFIRMED` | `backend/src/api/mcp.rs:83` — `if origin_host == host {` — compares the parsed `Origin` host against the request's `Host` header |
| Idle sessions with no live SSE subscriber are reclaimed; a watched one is not | `CONFIRMED` | `backend/src/mcp/session.rs:158` — `let keep = session.idle() < max_idle || session.has_subscribers();`, swept every `CLEANUP_INTERVAL` against `SESSION_MAX_IDLE` (`:21`), and `start_cleanup_task()` is called once per app in `backend/src/lib.rs:461` |
| Unknown tool names return an actionable JSON-RPC error, not `-32603` | `CONFIRMED` | `backend/src/mcp/handler.rs:604` — `"Unknown tool: {} - call tools/list for the available tools",` |
| This PR's own CI evidence is a normal `pull_request`-triggered run | `CONTRADICTED` | `.github/workflows/ci.yml:6-7` scope the `pull_request` trigger to `branches: [main]`; this PR's base is `chore/drop-stdio-mcp-server` (#765's branch), so it has never fired. The one green run (`34128316705`, `workflow_dispatch`, head `6522afc`) always skips `API Contract Check` (`ci.yml:193` — `` `if: github.event_name == 'pull_request'` ``) |

**Diagnosis** — All six fixes match the symptom they name and land in the module that owns the behavior (auth predicate, origin predicate, session manager, JSON-RPC dispatch). Coverage is `ABSOLUTE` for each: the auth fix is exercised with the exact password-only shape that used to 401, the origin fix with both a same-host and a cross-site case at two port shapes, the sweep with both the idle-collection and the live-subscriber-pins-it cases.

**Radius** — `LOCAL`. All six fixes are confined to `/api/mcp` and its two supporting modules; `auth.rs`'s new `session_is_authenticated` is a pure extraction (`auth_middleware`/`auth_status_handler` behavior unchanged). The sweep spawns one background task per `AppState`; a fresh `McpSessionManager` is created once per `create_app_with_config` call, so nothing calls `start_cleanup_task` twice on one manager. No pipeline or GStreamer reference is held — plain session structs, not a `WeakRef` concern.

**Tests & CI** — `mcp_endpoint_test.rs` adds 10 tests through the real router (`create_app()`), plus targeted unit tests in `mcp.rs`/`session.rs` — real coverage. The gap is procedural: because this PR's base isn't `main`, `ci.yml`'s `pull_request` trigger never runs on it (`gh pr checks 766` returns none). The only run is a manual `workflow_dispatch` at the current head, green for `Check (Linux)`/`Build (Linux x86_64/ARM64)`/`Check & Build (WASM)`, but that event type always skips `API Contract Check`. No new endpoints or `ToSchema` types are added, so risk is low, but it's `UNVERIFIED` rather than confirmed absent.

**Recommendation** — retarget to `main` once #765 merges, so a real `pull_request` run (including the API contract job) executes before merge.

Confidence: HIGH

PR #771, opened 2026-09-08
feat(blocks): add builtin.rtmp_output

borisasadanin:
## What this adds

A `builtin.rtmp_output` block: takes H.264 video and audio, muxes to FLV, publishes over RTMP.

The call site was already waiting for it. `backend/src/blocks/builtin/mod.rs` carries
`// blocks.extend(rtmp::get_blocks());` commented out under "Future: Add more protocols here",
`rtmpsink` is already in the frontend element palette, and `gstreamer1.0-plugins-bad` is already in
the Docker image, so `rtmp2sink` is present at runtime with no packaging change.

## Why we wanted it

We are building a C2PA provenance sidecar, and our chain runs a Strom mixing stage into a
downstream live encoder. Those two components land in different clusters on the hosting platform we
both run on, so their private addresses cannot reach each other. The encoder publishes a **public**
RTMP ingest address, which makes RTMP the one protocol that crosses the gap. Strom has a range of output blocks
and none of them speaks RTMP.

That is our motivation rather than a claim about your roadmap. If RTMP output is not something you
want in the built-in set, say so and we will close this; we would rather ask than carry a private
fork.

## Design notes, where we followed your existing code rather than the obvious route

**Audio is encoded inside the block when it arrives raw.** This is the part worth reviewing
first, because it is a judgement call. A mixing stage emits raw PCM, and there is no standalone AAC
audio encoder block, the way `builtin.videoenc` is one for video: `avenc_aac` and `opusenc` appear
only as internal chains inside three output blocks. So a block that demanded encoded AAC could be fed
only by an input in passthrough mode, where the stream arrives already encoded from upstream, and
neither of the two that qualify carries a mixed programme. Raw audio therefore goes through
`audioconvert`, `audioresample`, `avenc_aac`, `aacparse`; already-encoded AAC is only parsed. This
mirrors what your MPEG-TS-over-SRT, WHIP and EFP-over-SRT output blocks already do internally.

**Video is not encoded here.** Raw video is refused with a message naming `builtin.videoenc`, so
the encoder stays one explicit block with its own bitrate and profile settings rather than becoming
an invisible default inside an output.

**The muxer's pads are reserved in `register_element_setup`, and this is the part we got wrong first
and would most like you to check.** `flvmux` is an aggregator, so a sink pad requested for an input
that never carries data means it never aggregates and the sink receives nothing. The obvious answer
is to request each pad from its caps probe, once the input has proved it has data, which is what your
SRT output block does. **For FLV that is wrong, and it fails silently.**

`flvmux` writes the FLV header and its `onMetaData` at its first aggregation. A pad requested after
that is granted and its buffers are muxed into the body, but the stream is never declared. Measured
on GStreamer 1.28.6 with this block's exact topology: pads requested from the probes gave
`hasAudio=False hasVideo=True` with 128 audio tags sitting in the body; pads reserved before
`PLAYING` gave `hasAudio=True hasVideo=True`. A player that configures its decoders from the header
plays no sound from the first case, and it looks like an encoder or network fault rather than a graph
mistake. `mpegtsmux` is immune because it re-emits PAT and PMT continuously; FLV's header is written
once.

So the pads are reserved in `ctx.register_element_setup`, for the inputs that are actually linked.
`recorder.rs:380-393` documents the same race for `splitmuxsink` and points at the same hook, and we
only found this because a review pointed us there. Unconnected inputs get no pad, so video-only and
audio-only flows still work, and a connected input whose codec is then refused hands its reserved pad
back, so a refusal cannot stall the other stream. The probes still choose and insert the parser or
encoder chain; they link into a pad that already exists.

**`rtmps://` is supported, and certificate validation is deliberately not configurable.** The scheme
in the URL is all it takes: `rtmp2sink` sets its own `scheme` from it and its `tls-validation-flags`
default to `validate-all`. We have **not** exposed a way to weaken that, and would rather say why than
have you wonder: the only thing such a property buys is accepting a self-signed certificate, and the
cost is a switch whose whole function is to disable the check that makes `rtmps` worth using. A
deployment with an internal CA should trust it at the host. Happy to add the property if you disagree,
since it is your product and your users.

Both halves are measured rather than assumed, against `stunnel` in front of an RTMP listener with a
certificate from a private CA: with validation at its default the connection is refused with
"Unacceptable TLS certificate", and with the certificate trusted the same chain delivered 3.0 s of
H.264 and AAC through the tunnel. So the refusal is real rather than a default nothing enforces.

**A location that is neither `rtmp://` nor `rtmps://` is now refused at build time**, which is a
behaviour change worth flagging. `rtmp2sink` accepts any string and re-serialises what it could parse,
so `"not-a-url"` silently becomes `"rtmp:/"`, and `"rtmpss://host/live/key"` would quietly publish in
the clear. When a stream key in a URL is the only thing protecting a programme feed, a typo that drops
encryption without a word is the failure worth refusing. If you would rather it warned and continued,
say so.

**No secret in an RTMP URL reaches the logs, and there are three of them.** `user:pass@`, the stream
key in the last path segment, and any token in a query string. The stream key is the one worth
naming: for most servers it is the whole authorisation, so masking it is the difference between a log
line and a credential. The host and the application survive, so a line can still be attributed to an
output. This started as a plain bug on our side, since the block was logging the operator's raw URL.

Getting it right took two passes and the second is worth reporting, because the first looked fine.
Isolating the authority as "everything up to the first `/`" fails on a password containing an
unencoded `/`, which any base64 secret hits about a quarter of the time per character: the function
then found no credentials and returned its input, so the whole secret went into an `info!`. Userinfo
is now taken to end at the **last** `@` wherever it sits, including in a string the parser cannot
otherwise handle, because the reason it cannot may be the credential itself. Over-redaction is the
acceptable direction. The three inputs that broke it are now three tests.

**Two things this deliberately does not solve, said plainly.** The URL is stored with the flow, in
flow JSON and in the UI, because `location` is an ordinary string property: log redaction is not
encryption at rest. And we did not add `username` and `password` as block properties, because
`PropertyType` has no sensitive variant, so they would sit in the same places in plain text and add
surface to your API for no gain.

A non-loopback plaintext `rtmp://` target also gets one warning at build, since the key and the media
both cross the network readable.

**The sink is built with `async=false`.** Block-built elements do not pass through the path that
sets this for elements added the usual way, and a sink that waits to preroll holds the pipeline out
of PLAYING. We noticed this because of the commit on `main` that fixed the same failure mode
elsewhere.

**A missing plugin is reported as a missing plugin.** `require_rtmp_sink` checks the factory is
registered and, if not, returns an error naming the package to install, chosen per host OS. We had
a fallback to a second sink factory in an earlier draft and removed it: on a present-but-broken
element it reported "not installed" and would have sent an operator to install something they
already had.

**Both exposed properties are read.** `location` and `sync`. `sync` is exposed because an input
carrying a remote encoder's timestamps otherwise makes the sink believe it is behind and drop
frames. `DEFAULT_RTMP_LOCATION` went into `strom-types` beside `DEFAULT_SRT_OUTPUT_URI`, per the
contributor guide's rule about API-visible defaults.

## Element and property contracts, checked rather than assumed

Every GStreamer contract this block relies on was read from `gst-inspect-1.0` rather than recalled,
because a guessed pad name or property type is the failure this kind of block is most prone to:

- `flvmux` sink pad templates are literally named `audio` and `video`, both "On request", which is
  why they are requested with `request_pad_simple("video")` and `("audio")` rather than from a `%u`
  template.
- `flvmux streamable` is Boolean, default `false`, so setting it `true` is meaningful and not a
  no-op.
- `h264parse config-interval` is an Integer in range `-1 - 3600`, default `0`, where `-1` is
  documented as "send with every IDR frame". That is what the block sets and what its comment claims.
- Every element the block builds resolves to a plugin the runtime image already installs, so there is
  no packaging change: `identity` from `coreelements` in `libgstreamer1.0-0`; `flvmux` from `flv` and
  `aacparse` from `audioparsers` in `gstreamer1.0-plugins-good`; `rtmp2sink` from `rtmp2` and
  `h264parse` from `videoparsersbad` in `gstreamer1.0-plugins-bad`; `avenc_aac` from `libav` in
  `gstreamer1.0-libav`; `audioconvert` and `audioresample` in `gstreamer1.0-plugins-base`.

## Tests

`backend/tests/rtmp_output_test.rs` covers the three properties that are silent when they break:
the sink is not async, nothing is linked to the muxer at build time, and every property the block
reads is declared while every declared property is read. Plus the pad and mapping element ids
resolving to elements the block actually builds, and the missing-plugin message naming a package.

The codec decision each pad probe makes is covered too, through `video_plan` and `audio_plan`. They
are separate functions for exactly that reason: the branch logic is what a later edit is most likely
to get wrong, and every refusal message is asserted, because a wrong message sends an operator to
fix the wrong block.

**`rtmp_output_pipeline_test.rs` covers the element construction and pad linking those decisions lead
to**, driving the real builder through a `gst::Pipeline`: H.264 plus raw audio builds the parser and
the encode chain, AAC input is parsed without a second encoder, raw video is refused without taking
the audio side down, and an unconnected input gets no pad at all. It also asserts the pads exist
*before* `PLAYING`, which is the invariant the FLV header depends on.

We had claimed this needed a reachable RTMP server and used that to justify leaving it untested. That
was wrong: the probes fire on the CAPS event, so the chains are built whether or not anything is
listening. The four tests point the sink at a closed port and run in about 0.15 s. Worth saying
plainly, because it was the more embarrassing of the two problems a review found here: with the
probes untested, deleting both of them left the whole original suite green.

## What we have and have not verified

Stated plainly because it affects how much of this you should trust on reading.

**Built, formatted, linted and tested locally**, on macOS with Rust 1.98.0 and GStreamer 1.28.6:

- `cargo fmt --all -- --check`: clean
- `cargo clippy --package strom --features efp,nvidia --lib -- -D warnings`: clean
- the same on `--test rtmp_output_test` and `--test rtmp_output_pipeline_test`: clean
- `STROM_REQUIRE_GST_PLUGINS=1 cargo test ... --test rtmp_output_test`: 19 passed, 0 failed
- `STROM_REQUIRE_GST_PLUGINS=1 cargo test ... --test rtmp_output_pipeline_test`: 4 passed, 0 failed
- `STROM_REQUIRE_GST_PLUGINS=1 cargo test ... --test pipeline_lifecycle_test`: 3 passed, 0 failed
- `cargo test --package strom-types`: 58 passed, 0 failed

Every element the tests need is present on this host, so they ran rather than skipped.

**One local clippy failure that does not reproduce on your CI, recorded so the difference is not a
mystery.** `cargo clippy --all-targets` here fails on `backend/tests/vision_mixer_fx_test.rs:427`
with `clippy::chunks_exact_to_as_chunks`, a file this branch does not touch. It does **not** fire on
your CI, which resolved `dtolnay/rust-toolchain@stable` to rustc 1.98.1; ours is 1.98.0. So the lint
appears to have been dropped or downgraded in that patch release, and nothing here needs fixing. We
had predicted the opposite in an earlier revision of this description and your own green run
disproved it, which is the right way round.

**Now run on Linux by your own CI, and green.** Run
[34261385654](https://github.com/Eyevinn/strom/actions/runs/34261385654) on `67e5ea3`: `Check
(Linux)`, `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check & Build (WASM)`, `API Contract Check`
and `sccache preflight` all pass; macOS and Windows skipped as expected on a PR without those labels.
That is the first Linux build and the first run on your toolchain, and it covers the version gap we
could not close ourselves: we measure on GStreamer 1.28.6 here, your runner ships 1.24.2, and the
Docker runtime image is on 1.26.3+.

**That green run is on `67e5ea3` and there is a newer commit coming**, carrying the
`register_element_setup` fix described above and the pipeline tests. Worth re-running the checks after
it lands rather than trusting this one.

**The media contract, verified end to end against a real receiver.** We ran the exact chain this
block builds, `flvmux(streamable=true)` to `rtmp2sink`, fed by `h264parse` on the video side and
`audioconvert`, `audioresample`, `avenc_aac`, `aacparse` on the audio side, with `sync`, `async` and
`qos` set to the values the block sets. The receiver was a live `eyevinn-live-encoding` instance
reached through its published RTMP address, and it produced a valid HLS ABR ladder:

```
#EXT-X-STREAM-INF:BANDWIDTH=4540800,RESOLUTION=1280x720,CODECS="avc1.42c01f,mp4a.40.2"
#EXT-X-STREAM-INF:BANDWIDTH=3440800,RESOLUTION=640x360,CODECS="avc1.42c01e,mp4a.40.2"
```

A fetched segment probes as H.264 1280x720 plus AAC 48 kHz stereo. So the codec pairing, the FLV
container, the RTMP handshake and the URL shape are measured rather than modelled. One incidental
finding you may care about: `flvmux` emits a script-data stream that ffmpeg reports as a text
subtitle track, which was harmless in every run but is visible in the receiver's stream list.

**What is still not proven, stated so you do not have to find it.** The end-to-end probe above drove
the chain with `gst-launch-1.0` rather than this block, and the pipeline tests drive this block but
assert on graph shape rather than on bytes leaving the sink. So nothing here proves this block's own
output plays: the FLV header behaviour was measured on the same topology rather than on the block
itself, because the block's sink is `rtmp2sink` and capturing its bytes needs a server. If you would
rather that came as a test too, say so and we will find a way to assert it.

## Merging

Yours to decide. We will not merge this, and we will push follow-up commits to the branch if you
want changes.


borisasadanin:
## Reviewer: REQUEST CHANGES

Our own pre-submission review, posted straight away so nobody spends time on a version we already
know is wrong. **It found one real defect, in the part of the design we flagged in the description as
the bit to check first.** A fix is ready and will land on this branch shortly; I would hold off
reviewing until it does.

### `flvmux` never declares a stream whose pad arrives late

The block requested its `video` and `audio` pads from the caps probes, once each input had proved it
carried data. That reasoning was about not giving an aggregator a pad that never gets fed, and it is
wrong for FLV.

`flvmux` writes the FLV header and its `onMetaData` at its first aggregation. A pad requested after
that is granted, and its buffers **are** muxed into the body, but the stream is never declared.
Measured on GStreamer 1.28.6 with this block's exact topology:

| Pads requested | FLV header | Body |
|---|---|---|
| From the caps probes | `TypeFlags=0x01`, `hasAudio=False hasVideo=True` | 128 audio tags present |
| Before `PLAYING` | `TypeFlags=0x05`, `hasAudio=True hasVideo=True` | correct |

So a player that configures its decoders from the header gets no audio, and it looks like an encoder
or network fault rather than a graph mistake.

**Your own code already had the answer, which is the part worth saying out loud.**
`backend/src/blocks/builtin/recorder.rs:380-393` documents this race for `splitmuxsink` and rejects
the lazy route in as many words, pointing at `register_element_setup`, whose doc names exactly the
window we needed: after construction has linked every block, before the pipeline leaves NULL. Our
block was taking `_ctx` and discarding it. The claim in our module docs that this mirrored
`builtin.mpegtssrt_output` was the root cause: `mpegtsmux` re-emits PAT and PMT continuously, so a
late pad still gets announced, and FLV's header is written once.

The fix reserves the pads in that hook for the inputs that are actually linked, so unconnected inputs
still get no pad and a video-only flow still works, and a connected input whose codec is then refused
hands its reserved pad back, so a refusal cannot stall the other stream.

### Also fixed: a test gap we justified with a claim that turns out false

Our description said exercising the pad probes needs a reachable RTMP server. It does not. The probes
fire on the CAPS event, so the chains are built whether or not anything is listening. There is now a
pipeline test file driving the real builder against a closed port, four tests in about 0.15 s.

That gap was worse than a missing test: deleting both probe blocks left all 18 original tests green,
so the block's central mechanism had no coverage at all. The new tests were mutation-checked, and
breaking chain construction fails three of the four.

### Two things we have deliberately left for you rather than change unrequested

1. The `inserted` latch is consumed before the codec plan is evaluated, so a first CAPS event with an
   unusable codec permanently disables that pad. This is identical to `mpegtssrt.rs:376` and `:588`,
   so we have matched your existing behaviour, but latch-on-refusal is a different decision from
   latch-on-success and we could not tell which was intended.
2. `rtmp2sink` silently normalises an unparseable `location`: `"not-a-url"` reads back as `"rtmp:/"`.
   Our guard catches blank but not malformed, so a typo surfaces later as an obscure connect failure.
   Adding validation felt like your call rather than ours.

Happy to be told we have over-thought any of this.

srperens (commented):
**Verdict: Comment** — well shaped, and its tests genuinely execute in CI rather than skipping green, which is where this class of PR usually fails. One `flvmux` ordering question is worth settling before merge: if it bites, the symptom is a silently audio-less stream. Your clippy warning did not materialise.

**Requested changes**

1. Settle whether `flvmux` accepts an `audio` request pad after it has written its FLV header. Both mux pads are requested inside the caps probes — `backend/src/blocks/builtin/rtmp.rs:397` — `    let mux_sink = request_mux_pad(mux, "video")?;` and `backend/src/blocks/builtin/rtmp.rs:455` — `    let mux_sink = request_mux_pad(mux, "audio")?;` — so in a flow whose video caps arrive materially before its audio caps, the audio pad is requested after the muxer has begun streaming. Answer in the body either way.

**Claims**

|---|---|---|
| The new tests executed in CI rather than skipping | `CONFIRMED` | `backend/tests/rtmp_output_test.rs:39` — `    "rtmp2sink",` puts the sink in the required set, and `.github/workflows/ci.yml:149` — `          STROM_REQUIRE_GST_PLUGINS: 1` turns a missing element into a failure. `Check (Linux)` passed at `67e5ea3` |
| "Your clippy job will fail on this PR" | `CONTRADICTED` | `Check (Linux)` passed at `67e5ea3` and runs `.github/workflows/ci.yml:140` — `        run: cargo clippy --package strom --all-targets --features efp,nvidia -- -D warnings`. No companion lint PR needed |
| Only mux to sink is linked statically | `CONFIRMED` | `backend/src/blocks/builtin/rtmp.rs:252` — `        let internal_links = vec![(` holds one tuple, guarded by `nothing_is_linked_to_the_muxer_at_build_time` |
| `flvmux` accepts an `audio` pad requested after its header is written | `EXTERNAL` | upstream `gst-plugins-good`, not readable from here. Assumed true; if false, see below |

**Diagnosis** — Not a bug fix, so no root cause to check. Nothing to add to the audio-encode decision you flagged first: it is the shape `builtin.mpegtssrt_output` already uses, including the pad-on-caps probe — `backend/src/blocks/builtin/mpegtssrt.rs:577` — `                src_pad.add_probe(gst::PadProbeType::EVENT_DOWNSTREAM, move |pad, info| {`. Coverage is `BOUNDED` where you say: the two plan functions are tested, the construction and linking they lead to are not.

SPECULATIVE (not verified): FLV writes one header describing its streams, so a muxer that has begun streaming may refuse a new pad. If it does, `request_mux_pad` returns `Err` and that is only logged — `backend/src/blocks/builtin/rtmp.rs:243` — `                if let Err(e) = result {`. No builtin block posts to the bus (no `element_error!` or `post_error_message` under `backend/src/blocks/builtin/`), so the operator gets a video-only stream and one log line. Your `gst-launch-1.0` probe could not have caught it: both pads existed there from the start.

**Radius** — `LOCAL`. New block, no existing call site changed; `mod.rs` gains a registration and `strom-types` an additive const, which is where CLAUDE.md requires it. `API Contract Check` green, so no schema drift.

**Note** — `backend/src/blocks/builtin/rtmp.rs:160` — `        sink.set_property("qos", true);` copies the SRT sinks faithfully, but a block-built sink cannot reach the companion mitigation at `backend/src/gst/pipeline/construction.rs:341` — `                    drop_upstream_qos_events(&sink_pad);` — private to that module. Pre-existing for all three.

**Tests & CI** — eight checks green at `67e5ea3`. macOS and Windows report `skipping`; nothing here is platform-specific beyond `rtmp_package_hint`, so Linux-green plus the platform builds on merge is a reasonable call, not a gap.

Confidence: MED — HIGH on all but the `flvmux` row, which this repo cannot settle.

srperens:
Thanks — this is a good block, and the design notes made it quick to review. Two things, one of which is a heads-up rather than a request.

## First, the heads-up: the fix you describe is not on the branch

Your comment above describes reserving the pads in `register_element_setup` plus a new pipeline test file. Neither is in `67e5ea3`, which is still the head: single commit, five files, no `register_element_setup`, and `backend/tests/` has only `rtmp_output_test.rs`. So it looks like that work never got pushed. Worth checking your local branch before you do anything else here — you may already have most of what follows.

## Second: we settled the `flvmux` question, and you were right

Measured here on GStreamer 1.28.6 with this block's exact topology — `flvmux(streamable=true)` → sink, video from the start, `audio` pad requested 1.5 s later:

| | FLV header | audio tags in body |
|---|---|---|
| Both pads before the first buffer | `TypeFlags=0x05` hasAudio=True hasVideo=True | 175 |
| `audio` pad requested afterwards | `TypeFlags=0x01` **hasAudio=False** hasVideo=True | 130 |

So `flvmux` does **not** refuse the late pad. `request_pad_simple` succeeds, the link succeeds, the audio tags are written — and the header says there is no audio. The review's `EXTERNAL` row assumed refusal; the real failure is worse than that, because a refusal would at least have produced the error the review found was only logged. This produces nothing at all: silent video for any player that configures its decoders from the header, and no log line anywhere.

Your `gst-launch-1.0` probe could not have caught it, as the review noted — both pads existed from the start there.

## A commit you can use

Pushed to this repo as `suggest/rtmp-flvmux-pad-reservation`, one commit on top of your head:

```
git fetch origin suggest/rtmp-flvmux-pad-reservation
git cherry-pick a78bd39
```

It separates the two questions the block was asking at once. **Which inputs exist** is answered in `register_element_setup` — after the flow has linked every block, before the pipeline leaves NULL, so connectivity is known and the muxer has not started. **What codec each input carries** stays in the caps probes, which now link the pad reserved for them rather than requesting their own.

The properties you argued for are kept: an unconnected input still gets no pad, so a never-fed aggregator pad cannot stall the muxer, and a connected input whose codec is refused hands its reserved pad back so the refusal cannot take the other stream down with it.

Take it, rewrite it, or squash it into your commit — whichever you prefer. If you have your own version locally, compare and keep yours; we are not attached to this one.

### On the test gap

You were right that it does not need an RTMP server, and the reason is worth recording: the probes fire on the CAPS event, and **preroll produces one**. So the new `rtmp_output_pipeline_test.rs` takes the pipeline to PAUSED and no further — `rtmp2sink` only dials its server on the way to PLAYING, which we confirmed (against a closed port it reaches PREROLLED cleanly and errors only after `Setting pipeline to PLAYING`). Four tests, 0.04 s, and **all four fail against `67e5ea3`**, checked by restoring your `rtmp.rs` and re-running.

Your point about the gap being worse than a missing test also checked out and is now in the docs of the unit suite: deleting both probe bodies leaves all 18 of its tests green.

## Two things we are not asking you to change

- **The `inserted` latch consumed before the codec plan.** Matching `mpegtssrt.rs:376`/`:588` was the right call. If we want latch-on-success it should change in all three places at once, in its own PR.
- **`location` validation.** Confirmed, and slightly wider than you found: `rtmp2sink` normalises `"not-a-url"`, `""` and `"http://example.com/x"` all to `"rtmp:/"`, so a wrong *scheme* is swallowed the same way a typo is. Also `rtmp://h:1935/p` → `rtmp://h/p`, which your `without_default_port` helper already handles. Still your call to leave; a guard belongs with the other output blocks' URL handling rather than only here.

Nothing else from us. The audio-encode decision is right, `async=false` is right and correctly explained, and `require_rtmp_sink` reporting a missing plugin as a missing plugin — with the fallback removed — is exactly the judgement we want.

srperens:
Two follow-ups filed off the back of this PR, both deliberately kept out of it — neither is a change we want to ask you to make here, and #783 in particular adds a property and a bus handler to a block still under review.

- **#783** — have this block constrain its video input to the profile set RTMP receivers accept, defaulting to `{ high, main, baseline, constrained-baseline }`. Measured: a consumer-side constraint reaches back through `h264parse` and `x264enc`, so a UYVY source that today yields `High 4:4:4 Predictive` yields `High` / `yuv420p` instead. It also breaks an already-encoded 4:2:2 passthrough input with `not-negotiated`, which is why it needs the property and a decent error rather than being switched on quietly.
- **#782** — the Video Encoder block pins no profile by default, which is what lets a 4:2:2 or 10-bit input reach an encoded profile most targets refuse.

Neither blocks this PR. Also worth recording here, since it came out of testing this block rather than either issue: the block publishes a valid two-stream FLV through a real RTMP handshake — `TypeFlags=0x05`, H.264 1280x720 + AAC-LC, clean EOS, no bus errors — driven through the real builder against an `ffmpeg -listen 1` server. And `ffprobe` does report the `flvmux` script-data stream as a `text` track, so that observation of yours holds.

srperens (commented):
**Verdict: Comment** — the `flvmux` header question my review of `67e5ea3` left open is now answered, measured, and fixed in the right place. Reserving the pads up front introduces one case this repo can settle and has not: an input that is *linked* but silent.

**Requested changes**

1. Cover the linked-but-silent input in `backend/tests/rtmp_output_pipeline_test.rs`. Its fourth test covers an unlinked input — `backend/tests/rtmp_output_pipeline_test.rs:446` — `    // Audio deliberately not fed, so its input stays unconnected.` — but reservation keys on link state alone, so a linked input that stays silent holds a pad nothing feeds. Add a case that links a source pushing nothing and assert buffers still reach the standin sink. If they do not, give `flvmux` `force-live` as `make_audiomixer` does — `backend/src/blocks/builtin/mixer/elements.rs:30` — `        builder = builder.property("force-live", force_live);` — probed for availability first, `backend/src/blocks/builtin/mixer/elements.rs:24` — `            .map(|probe| probe.find_property("force-live").is_some())`.

**Claims**

|---|---|---|
| The setup hook runs after linking, before the muxer can start | `CONFIRMED` | `backend/src/gst/pipeline/bus.rs:49` — `        for setup_fn in element_setups {` runs inside `setup_bus_watch`, called at `backend/src/gst/pipeline/lifecycle.rs:42` — `        self.setup_bus_watch();` — ahead of the first `set_state` in the same function |
| A pad reserved for a linked input that never sends data cannot stall the sink | `UNVERIFIED` | reservation reads link state only, `backend/src/blocks/builtin/rtmp.rs:805` — `        .map(|p| p.is_linked())` — while the chain is built from a CAPS event, `backend/src/blocks/builtin/rtmp.rs:536` — `            src_pad.add_probe(gst::PadProbeType::EVENT_DOWNSTREAM, move |pad, info| {` |
| A linked input can stay silent for minutes here | `CONFIRMED` | `4780965` put an idle WHIP slot in NULL, so its decode chain emits no caps until a publisher arrives; `f67bcf9` did the same for a recorder track |
| The new tests executed in CI rather than skipping | `CONFIRMED` | `Check (Linux)` success at `ee0ef20`, running `.github/workflows/ci.yml:146` — `        run: cargo test --package strom --features efp,nvidia` under `STROM_REQUIRE_GST_PLUGINS: 1`; `avenc_aac` and `x264enc` ship in the packages installed at `.github/workflows/ci.yml:100-110` |

**Diagnosis** — the measurement you added (`hasAudio=False` with 128 audio tags) explains the symptom fully, and the hook is the right layer. What moved with it is the failure mode for a stream that is wired but idle: before, video published with an under-declared header; now the muxer holds a pad nothing feeds, which your docs say stops it aggregating — `backend/src/blocks/builtin/rtmp.rs:36` — `//! An aggregator's sink pad that never carries data means `flvmux` never`. This repo met that shape twice this month (#749, #750) and its standing answer for an aggregator is `force-live`: `backend/src/blocks/builtin/mixer/mod.rs:54` — `//! force-live=true so unconnected input pads don't stall the pipeline.`, and `liveadder` in the WHEP block is built the same way. Coverage is `BOUNDED` to inputs whose caps arrive; the refusal path hands its pad back, which the runtime test proves.

**Radius** — `LOCAL`. New block; `mod.rs` gains a registration, `strom-types` an additive const. `API Contract Check` green, so no schema drift.

**Tests & CI** — eight checks at `ee0ef20`: six pass, macOS and Windows skipped. Nothing is platform-specific beyond the `cfg` strings in `rtmp_package_hint`, so Linux-green plus the platform builds on merge is a reasonable call, not a gap.

Confidence: MED — HIGH on all but the stall row, which one test in your own harness settles either way.

srperens:
Thanks for this — the reasoning about the FLV header and the pad reservation is correct, and the `register_element_setup` route is the right one. I rebased it, reconciled it with a block that landed after you opened this, and then closed the gap you flagged yourself: that nothing proved this block's own bytes play.

Rebased branch, for you to take or rebase onto: [`pr-771-rebase`](https://github.com/Eyevinn/strom/tree/pr-771-rebase). I have not touched your branch. It carries your commit replayed on current `main` plus one commit of mine.

## The rebase

Clean onto `main` (28 commits). Both conflicts were additions to sorted lists and resolved automatically.

## `builtin.audioenc` landed, and it changes your argument

`builtin.audioenc` merged in #744 after you opened this. It encodes raw audio to AAC, Opus, MP3 or AC-3, so "no block produces AAC as its output" is no longer true, and it was the load-bearing premise of your audio section.

The behaviour is still right — `builtin.mixer` emits raw audio, the common case is a mixed programme going straight out, and your three sibling output blocks all encode raw audio internally — so I kept it and changed the reason. The docs now argue from the defaults: an H.264 encoder has a bitrate and a profile someone has to choose, an AAC encoder at 128 kbps does not. Every audio refusal now names `builtin.audioenc` the way your video refusal names `builtin.videoenc`.

Two tests pin the pairing, because nothing else links the two sides: one that every audio refusal names the block, and one that `audioenc`'s default output (`audio/mpeg,mpegversion=4`) takes your parse-only path. Without it a drift on either side would quietly run a second encoder over already-encoded audio.

## Run against a real receiver, through the block

ffmpeg turned out to be useless as the receiver here: its FLV demuxer rebuilds streams from body tags when the metadata does not mention them, so it hides precisely the failure your design prevents. Measured — an FLV with its audio declaration stripped out still probes as h264+aac. And over RTMP there is no FLV file header at all, so the declaration is entirely `onMetaData`, which is what you measured too.

So I wrote a minimal RTMP receiver that reads `onMetaData` verbatim, validated it against a known-good `gst-launch` chain first, and ran five flows through Strom built from the rebased branch:

| Flow | Result |
|---|---|
| `videoenc` + raw audio | `videocodecid=7`, `audiocodecid=10`, 165 video / 439 audio messages — metadata matches the media |
| `audioenc(aac)` → `rtmp_output` | `identity -> aacparse -> flvmux`, no second encoder, both declared |
| `media_player` passthrough | both pads reserved, 362 / 517, no late-pad warning |
| video only, audio unconnected | `audio input is not connected, reserving no pad`, video declared alone |
| raw video in | refused, pad released, 377 audio messages still arrived |

Every reserved pad logged `before the pipeline started`. So the invariant holds in a running flow, and the block's own output carries and declares both streams. That is the thing your description said was not proven.

Incidentally your `flvmux` script-data observation reproduces: it shows up as a `data` stream in the receiver's list. Harmless in every run.

`cargo fmt`, `cargo clippy --lib --tests`, and `rtmp_output_test` (41), `rtmp_output_pipeline_test` (4), `pipeline_lifecycle_test` (3) and `openapi_test` (1) all pass locally on macOS with GStreamer 1.28.6. Nothing skipped. I have not run the Linux CI on the rebased branch.

## Two things to fix

**The redaction claim is not true as written, and I corrected it in my commit.** The API's flow-create handler logs the whole request body at `debug`, so a `location` carrying a stream key is in the log before your block is ever built — I have the line. Your masking is still worth having, since those are the `info` lines a deployment actually runs at, but the unqualified claim would let a reader conclude the URL is a secret.

**The refusal path posts a bus error, and this one is yours.** When raw video is refused you release the mux pad, which is right, but the input identity's src pad is left unlinked, so the upstream source goes to `not-linked` and the flow posts `Internal data stream error`. The audio keeps flowing — your invariant holds — but the flow reads as failed in the UI and would trigger auto-restart where it is enabled. A DROP probe on the identity's src pad at the point of refusal would settle it.

One more thing worth recording: the `input_is_connected` caveat you inherited from `recorder.rs` does not bite here. I tested it with `media_player`, the one input path with genuinely dynamic pads, and both pads were reserved — the input blocks front their dynamic pads with static `identity` elements, so the links resolve during construction and never reach `pending_links`.

PR #776, opened 2026-09-09
docs: note publishing to YouTube Live and Twitch as an idea

srperens:
One entry on the ideas list, under "Platform & reach": try Strom's RTMP output against YouTube Live and Twitch ingest.

`builtin.rtmp_output` (#771) is what makes the thought worth writing down, and `rtmp2sink` advertises both schemes the platforms use — `get_protocols()` returns `['rtmp', 'rtmps']`, and it accepts `rtmps://a.rtmps.youtube.com:443/live2/KEY` and `rtmp://live.fra.twitch.tv/app/KEY` unchanged apart from dropping a default port.

That is the whole of what has been checked. Nothing has been streamed to either platform, so the entry says to measure it rather than claiming it works — `docs/FEATURE_SUGGESTIONS.md` is explicitly not a roadmap, which is why this goes there instead of a doc that would read as a supported feature.

Docs only, no code.

### DIFF
--- a/docs/FEATURE_SUGGESTIONS.md
+++ b/docs/FEATURE_SUGGESTIONS.md
@@ -51,10 +51,10 @@ interests you, open a GitHub issue or discussion.
   and Twitch and write down what an operator has to set. The transport side is there: the
   sink speaks `rtmp` and `rtmps`, and it keeps the query string Twitch's
   `?bandwidthtest=true` rides on, so a test broadcast can be verified in Twitch Inspector
-  without going live. The encoder side is where the work is — the Video Encoder block's
-  default profile lets the encoder negotiate freely, and against `x264enc` that lands on
-  High 4:4:4 Predictive 10-bit, which neither platform accepts. Pinning `profile=high` gives
-  8-bit 4:2:0. Both platforms also want an audio track and roughly 2-second keyframes.
+  without going live. The encoder side needs a look — the Video Encoder block pins no
+  profile by default, which is fine from a 4:2:0 source but follows a 4:2:2 or 10-bit one
+  into a profile no platform ingest accepts, so a capture feed may need `profile=high` set
+  by hand. Both platforms also want an audio track and roughly 2-second keyframes.
 - **Kubernetes operator** — deploy flows as pods with resource limits and auto-scaling.
 - **Block marketplace** — browse and install community-contributed blocks.
 - **Mobile companion** — monitor status, start/stop flows, and receive alerts from a phone.
--- a/docs/FEATURE_SUGGESTIONS.md
+++ b/docs/FEATURE_SUGGESTIONS.md
@@ -47,10 +47,14 @@ interests you, open a GitHub issue or discussion.
 
 ## Platform & reach
 
-- **Publish straight to the big live platforms** — try Strom's RTMP output against YouTube
-  Live and Twitch ingest and write down what it takes. Both accept H.264 + AAC over RTMP or
-  RTMPS, and both want an audio track present, so a mix should be able to reach a stream key
-  with no separate encoder in between. Worth measuring rather than assuming.
+- **Publish straight to the big live platforms** — take Strom's RTMP output to YouTube Live
+  and Twitch and write down what an operator has to set. The transport side is there: the
+  sink speaks `rtmp` and `rtmps`, and it keeps the query string Twitch's
+  `?bandwidthtest=true` rides on, so a test broadcast can be verified in Twitch Inspector
+  without going live. The encoder side is where the work is — the Video Encoder block's
+  default profile lets the encoder negotiate freely, and against `x264enc` that lands on
+  High 4:4:4 Predictive 10-bit, which neither platform accepts. Pinning `profile=high` gives
+  8-bit 4:2:0. Both platforms also want an audio track and roughly 2-second keyframes.
 - **Kubernetes operator** — deploy flows as pods with resource limits and auto-scaling.
 - **Block marketplace** — browse and install community-contributed blocks.
 - **Mobile companion** — monitor status, start/stop flows, and receive alerts from a phone.
--- a/docs/FEATURE_SUGGESTIONS.md
+++ b/docs/FEATURE_SUGGESTIONS.md
@@ -47,6 +47,10 @@ interests you, open a GitHub issue or discussion.
 
 ## Platform & reach
 
+- **Publish straight to the big live platforms** — try Strom's RTMP output against YouTube
+  Live and Twitch ingest and write down what it takes. Both accept H.264 + AAC over RTMP or
+  RTMPS, and both want an audio track present, so a mix should be able to reach a stream key
+  with no separate encoder in between. Worth measuring rather than assuming.
 - **Kubernetes operator** — deploy flows as pods with resource limits and auto-scaling.
 - **Block marketplace** — browse and install community-contributed blocks.
 - **Mobile companion** — monitor status, start/stop flows, and receive alerts from a phone.


srperens:
Went and measured this instead of leaving it as a guess, and the entry is now rewritten around what came back. Short version: **the RTMP side is fine and the encoder side is not.**

## The block publishes a valid stream to a real RTMP server

Not a hand-built equivalent chain — the real `RtmpOutputBuilder`, its element setup hook and its caps probes, driven to PLAYING against an `ffmpeg -listen 1` RTMP server on loopback. What the receiver wrote:

```
FLV header: TypeFlags=0x05  hasAudio=True hasVideo=True   1698332 bytes
tags:       audio=221  video=152  script=1
ffprobe:    h264 1280x720 + aac LC 44100 Hz, flv, 5.14 s
```

Clean EOS, no bus errors. So the FLV container, the RTMP handshake and the codec pairing hold up through a real server-side implementation.

Incidentally this confirms the `script`/`text` stream @borisasadanin noticed in #771: `ffprobe` reports it as a `text` stream. Harmless, present.

## The obstacle is the Video Encoder block's default profile

`Profile::None` is the default and documented as "no profile field — let the encoder negotiate freely". Measured through the chain that block builds, `videoconvert ! x264enc ! h264parse ! video/x-h264,alignment=au`:

| Video Encoder `profile` | Result |
|---|---|
| `none` (the default) | `High 4:4:4 Predictive`, `yuv444p10le` |
| `high` | `High`, `yuv420p` |

10-bit 4:4:4 is not something YouTube or Twitch will take. So an operator who wires up a default encoder and points RTMP output at a stream key gets rejected by the platform, and nothing in Strom is wrong — which is the least helpful kind of failure. Pinning `profile=high` is the fix, and that is what the entry now says.

Worth deciding separately whether `none` is a good default at all: 4:4:4 10-bit is a surprising thing to hand any consumer, not just these two platforms. Not proposing a change here.

## Twitch has exactly the test path you would want

`?bandwidthtest=true` appended to the stream key sends a real broadcast to Twitch's ingest, keeps the channel offline and no followers notified, and reports bitrate, frame drops and disconnects in [Twitch Inspector](https://inspector.twitch.tv/).

That rides on the URL as a query string, so it only works if the sink preserves one. Checked — it does:

```
rtmp://ingest.global-contribute.live-video.net/app/live_123_abc?bandwidthtest=true   unchanged
rtmps://…/app/live_123_abc?bandwidthtest=true                                        unchanged
rtmp://example.com/app/key?a=1&b=2                                                   unchanged
```

`rtmp2sink.get_protocols()` returns `['rtmp', 'rtmps']`, and `https://ingest.twitch.tv/ingests` is a public JSON list of the 14 ingest servers with `url_template` and `url_template_secure` carrying a `{stream_key}` placeholder — so picking a nearby ingest is a lookup, not a guess.

## What is still untested, and by whom

Neither platform can be reached without an account and a stream key, so nobody has streamed to either from Strom. Everything above is loopback plus documentation. The Twitch path is the cheap one to close, because the bandwidth test needs no scheduled broadcast and leaves no trace on a channel.

Sources: [Twitch ingest list](https://ingest.twitch.tv/ingests), [Twitch Inspector](https://inspector.twitch.tv/), [Video Broadcast | Twitch Developers](https://dev.twitch.tv/docs/video-broadcast/), [How to run a test stream on Twitch](https://restream.io/learn/platforms/how-to-run-test-stream-on-twitch/)

srperens:
Correcting my comment above: I said the default profile "lands on High 4:4:4 Predictive 10-bit". That was measured from a test source with no format pinned, which negotiates upward to whatever the encoder prefers — so the source chose that, not the default.

From a 4:2:0 source the same chain gives `High` / `yuv420p`, which every platform takes:

| Format reaching the encoder | Negotiated H.264 |
|---|---|
| `I420`, `NV12` | `High` / `yuv420p` |
| `UYVY` | `High 4:4:4 Predictive` |
| `Y42B` | `High 4:2:2` |
| `I420_10LE` | `High 10` |
| `P010_10LE` | `High 4:4:4 Predictive` (8-bit) |

So the exposure is real but narrower than I wrote: it follows a 4:2:2 or 10-bit input into a profile the platforms refuse, and leaves a 4:2:0 one alone. Filed as #782, and the docs entry on this branch now says the narrower thing.

Nothing changes about the RTMP measurements in that comment — those stand.

srperens:
The two things this branch's docs entry tells an operator to do by hand are now filed as work instead:

- **#783** — `builtin.rtmp_output` constrains its own video input to the profile set RTMP receivers accept, by default. This is the one that would make the entry's "a capture feed may need `profile=high` set by hand" stop being true.
- **#782** — the Video Encoder block's default profile, which is why that sentence is there.

The entry stays as it is either way: it describes what an operator faces today, and `FEATURE_SUGGESTIONS.md` is explicitly not a roadmap, so it should not promise either issue will land.

srperens (commented):
**Verdict: Comment** — the entry is in the right file and its encoder half checks out, but its transport half is written in the present tense about a block that does not exist on `main`.

**Requested changes**

1. In `docs/FEATURE_SUGGESTIONS.md`, make the two sink clauses conditional on #771 landing — e.g. "once the RTMP Output block (#771) lands, the transport side is there". Merged before #771, this puts a capability claim on `main` that `main` cannot back.

**Claims**

|---|---|---|
| "The transport side is there: the sink speaks `rtmp` and `rtmps`" | `CONTRADICTED` | `backend/src/blocks/builtin/mod.rs:138` — `    // blocks.extend(rtmp::get_blocks());` is the RTMP block, still commented out under "Future" |
| No other RTMP sink stands in for it | `CONFIRMED` | the only other hit in `backend/src`, `frontend/src`, `types/src` is `frontend/src/palette.rs:164` — `                name: "rtmpsink".to_string(),`, an entry in the raw-element palette, not a block |
| "it keeps the query string Twitch's `?bandwidthtest=true` rides on" | `UNVERIFIED` | there is no query-string handling on `main` to check; that code is in #771's diff only |
| "the Video Encoder block pins no profile by default" | `CONFIRMED` | `types/src/videoenc.rs:12-15` — `Profile::None` carries `#[default]`; `backend/src/blocks/builtin/videoenc.rs:842` — `            None => "video/x-h264,alignment=au".to_string(),` omits the field entirely |
| An unconstrained 4:2:2 / 10-bit source lands in a profile platform ingest rejects | `EXTERNAL` | depends on what the selected encoder advertises when nothing is pinned — GStreamer negotiation, not settleable from this repository. Stated as an assumption; it is a reasonable one. |

**Diagnosis** — There is no defect to diagnose; this is an idea entry, and CLAUDE.md puts ideas exactly here ("Ideas go in `docs/FEATURE_SUGGESTIONS.md` (an unordered 'not a roadmap' list)"). The single real problem is tense. An idea list may describe what a feature *would* need, but "the transport side is there" and "it keeps the query string" read as statements of current fact, and both describe #771. The encoder paragraph is the model to follow: it says what needs a look, and it is accurate against `main` today.

Worth noting for whoever picks the idea up: the profile hazard is real but the block already exposes the escape hatch the text asks an operator to reach for by hand — `backend/src/blocks/builtin/videoenc.rs:910` — `                name: "profile".to_string(),`.

**Radius** — `LOCAL`: one Markdown file, no code, nothing consumed by the build.

**Tests & CI** — run 34345214873 at `aa4ca9a7` (this PR's head): `Check (Linux)`, `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check & Build (WASM)`, `API Contract Check` all `pass`. `Build (macOS)` and `Build (Windows)` report `skipping`, which is the label gate and is not a gap for a docs-only diff.

Confidence: HIGH

PR #784, opened 2026-09-09
fix(vision_mixer): stop overlay timer threads on every teardown path

### LINKED ISSUE #727: Overlay timer threads survive a failed flow start and burn ~2.5% of a core each for the life of the process

srperens:
## Description

The vision mixer's `overlay-timer-*` thread had exactly one exit condition — its renderer leaving the global overlay registry — and that unregistration lived in `stop_flow`'s `for block in &flow.blocks` loop alone. Four paths skip that loop: a failed `manager.start()`, a WHEP endpoint conflict, a WHIP endpoint conflict, and `stop_flow`'s early return when no manager is registered. A flow edited while running misses it too, since the loop matches against the *current* block list rather than the one the pipeline was built from.

Each survivor rendered the overlay clock at full framerate into an appsrc whose pipeline was gone. Issue #727 measured six of them on a 0.6.7 deployment with no flow running: ~2.5% of a core each, ~6.5 CPU-hours between them, one alive for 46 h.

**The fix, in two parts.**

*Primary.* The overlay registries now carry the owning flow with each entry, and `teardown_flow` sweeps them by flow id. `teardown_flow` is where the failed `manager.start()`, WHEP-conflict and WHIP-conflict paths converge — it already centralises endpoint, Media Player and CPU-core cleanup — so those three close at once, and the flow-edited-while-running case closes with them: keying off the flow rather than a block list means the old loop's drifted `flow.blocks` no longer decides what gets cleared.

The fourth path is deliberately left alone. `stop_flow`'s early return when no manager is registered still does not reach `teardown_flow`, because nothing can be registered by the time it is taken: the only ways in are a second stop, where the first already swept, and a flow that never started. Closing it would be a line, but it would guard nothing.

The sweep runs first in `teardown_flow`, before `manager.stop()`. That narrows the window in which the timer is still pushing into an appsrc on its way to NULL rather than closing it — the thread notices at its next tick, up to one frame interval later — which is harmless, since a push into a flushing appsrc just returns `FLUSHING`. It also runs on the branch where no pipeline was ever built, which is reachable: `build_overlay` registers during construction, so a build that fails *after* the vision mixer leaves entries behind with no `PipelineManager` to consult — that branch's own comment already anticipated this class of leak.

*Backstop.* The timer also exits once its appsrc has no parent. A pipeline unparents its children when it finalizes, so a future teardown path that forgets to unregister costs one poll instead of a core for the life of the process.

To be explicit about what the backstop does **not** cover: it only fires once the pipeline actually finalizes. A leaked strong reference — the case `teardown_flow` already logs `Pipeline still alive after drop` for — keeps the appsrc parented and defeats it too. It guards a missed unregistration, not a reference leak. The unregistration stays the primary mechanism; the backstop is there because the failure mode is so asymmetric (one forgotten line = 2.5% of a core, permanently).

## Related Issue

Fixes #727

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Tests

Three tests, each verified to fail with its own fix reverted — not just green alongside it.

**`backend/tests/vision_mixer_failed_start_test.rs` (new)** — `failed_start_leaves_no_overlay_timer`. Builds a real flow with a vision mixer on the CPU backend plus an `audiotestsrc ! fakesink async=false state-error=paused-to-playing` branch, drives it through `AppState::start_flow`, and asserts after the failure that both registries are empty and the timer thread count is back to baseline.

Both halves of that assertion are load-bearing, which the three-way check confirmed:

| teardown sweep | backstop | result |
|---|---|---|
| off | off | `left: 1, right: 0` — one leaked thread, the exact bug |
| off | on | registry assertion fails; thread count passes on the backstop alone |
| on | off | passes |

The middle row is why the test asserts on the registries and not only on the thread count: with the backstop in place, a thread-count-only test would pass even if nothing unregistered.

**`overlay_timer_exits_when_its_appsrc_is_orphaned` (new, unit)** — starts a real timer, waits until it is in its push loop, then removes the appsrc from its bin with the renderer *still registered*. Fails (`left: 1, right: 0`) with the backstop reverted.

**`overlay_registries_round_trip` (extended)** — now registers two flows and asserts that sweeping one leaves the other's overlay intact, so the `retain` is checked for aim as well as reach.

Both timer tests take a shared lock: `shutdown_overlay_timers` and the running count are process-global, and cargo's parallel test threads made one test stop the other's thread and skew its baseline. The lock is what makes the suite deterministic rather than passing per-file only.

Everything runs in CI as configured: `audiotestsrc`, `fakesink` and the CPU compositor backend need no GL context and no package beyond what `.github/workflows/ci.yml` already installs. Nothing is skipped or gated behind an element probe.

**Verified end to end against a running server**, beyond the automated tests: a CPU vision mixer flow started and stopped goes 1 -> 0 `overlay-timer-*` threads; started, then given a new block id while running, then stopped, also reaches 0, where the same sequence on the parent commit leaves the thread behind. That is the one case above with no automated guard, since it needs a live flow edit.

**Ran locally on macOS (Darwin 25.6.0, GStreamer 1.26):** `cargo test --workspace` — all green, 572 lib tests plus every integration binary. `cargo clippy --workspace --all-targets -- -D warnings` — clean. The two pre-existing `#[ignore]`d tests stayed ignored; nothing else was skipped.

## Checklist

- [x] I have run `cargo clippy --workspace --all-targets -- -D warnings`
- [x] I have updated documentation as needed (code comments only — no doc drift added)

### DIFF
--- a/backend/src/blocks/builtin/vision_mixer/builder/mod.rs
+++ b/backend/src/blocks/builtin/vision_mixer/builder/mod.rs
@@ -169,6 +169,7 @@ impl BlockBuilder for VisionMixerBuilder {
 
         let p = PipelineParams {
             instance_id,
+            flow_id: properties::parse_flow_id(props),
             num_inputs,
             num_dsk_inputs,
             num_pips,
@@ -205,6 +206,9 @@ impl BlockBuilder for VisionMixerBuilder {
 /// Shared parameters for pipeline construction.
 pub(super) struct PipelineParams<'a> {
     pub(super) instance_id: &'a str,
+    /// Owning flow, used to key the overlay registries so flow teardown can
+    /// clear them — see `overlay::unregister_flow`.
+    pub(super) flow_id: strom_types::FlowId,
     pub(super) num_inputs: usize,
     pub(super) num_dsk_inputs: usize,
     pub(super) num_pips: usize,
@@ -344,7 +348,7 @@ pub(super) fn setup_overlay_renderer(
     ));
 
     // Register the overlay state so the API layer can access it
-    overlay::register_overlay_state(p.instance_id, Arc::clone(&overlay_state));
+    overlay::register_overlay_state(p.flow_id, p.instance_id, Arc::clone(&overlay_state));
 
     let renderer = Arc::new(Mutex::new(OverlayRenderer::new(
         appsrc.clone(),
@@ -355,7 +359,7 @@ pub(super) fn setup_overlay_renderer(
     )));
 
     let block_id = p.instance_id.to_string();
-    overlay::register_overlay_renderer(&block_id, Arc::clone(&renderer));
+    overlay::register_overlay_renderer(p.flow_id, &block_id, Arc::clone(&renderer));
 
     let block_id_for_timer = block_id.clone();
     let renderer_for_timer = Arc::clone(&renderer);
--- a/backend/src/blocks/builtin/vision_mixer/overlay.rs
+++ b/backend/src/blocks/builtin/vision_mixer/overlay.rs
@@ -17,33 +17,34 @@ use std::sync::{Arc, Mutex, OnceLock};
 use std::thread::JoinHandle;
 use std::time::{Instant, SystemTime};
 use strom_types::vision_mixer::{self, Zone, TIMEZONE_REFRESH_SECS};
+use strom_types::FlowId;
 use tracing::{debug, warn};
+
+/// A per-block overlay registry: keyed by block instance ID, each entry tagged
+/// with the flow that registered it so [`unregister_flow`] can sweep by flow.
+type OverlayRegistry<T> = Mutex<HashMap<String, (FlowId, T)>>;
 
 /// Global registry of vision mixer overlay states, keyed by block instance ID.
 /// Used by the API layer to access overlay state for preview/PGM updates.
-fn overlay_states() -> &'static Mutex<HashMap<String, Arc<VisionMixerOverlayState>>> {
-    static INSTANCE: OnceLock<Mutex<HashMap<String, Arc<VisionMixerOverlayState>>>> =
-        OnceLock::new();
+fn overlay_states() -> &'static OverlayRegistry<Arc<VisionMixerOverlayState>> {
+    static INSTANCE: OnceLock<OverlayRegistry<Arc<VisionMixerOverlayState>>> = OnceLock::new();
     INSTANCE.get_or_init(|| Mutex::new(HashMap::new()))
 }
 
 /// Register an overlay state for a block instance.
-pub fn register_overlay_state(block_id: &str, state: Arc<VisionMixerOverlayState>) {
+pub fn register_overlay_state(
+    flow_id: FlowId,
+    block_id: &str,
+    state: Arc<VisionMixerOverlayState>,
+) {
     if let Ok(mut map) = overlay_states().lock() {
-        map.insert(block_id.to_string(), state);
+        map.insert(block_id.to_string(), (flow_id, state));
     }
 }
 
 /// Get the overlay state for a block instance (if registered).
 pub fn get_overlay_state(block_id: &str) -> Option<Arc<VisionMixerOverlayState>> {
-    overlay_states().lock().ok()?.get(block_id).cloned()
-}
-
-/// Unregister the overlay state for a block instance (call on flow stop).
-pub fn unregister_overlay_state(block_id: &str) {
-    if let Ok(mut map) = overlay_states().lock() {
-        map.remove(block_id);
-    }
+    Some(overlay_states().lock().ok()?.get(block_id)?.1.clone())
 }
 
 /// Shared state read by the cairooverlay draw callback.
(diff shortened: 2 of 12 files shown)


srperens:
CI confirmation, since "the suite is green" is not the same as "the new test ran" — all three executed on Linux x86_64 (`Check (Linux)` → `Run tests on backend`):

```
test blocks::builtin::vision_mixer::tests::overlay_registries_round_trip ... ok
test blocks::builtin::vision_mixer::tests::overlay_timer_exits_when_its_appsrc_is_orphaned ... ok
test failed_start_leaves_no_overlay_timer ... ok
```

No element probe, no skip. Windows and macOS builds are `skipping` per the existing workflow conditions, unchanged by this PR.

srperens (commented):
**Verdict: Approve** — the leak is closed at the one place every teardown converges, and the flow-keyed sweep is the shape this function already uses for the media player registry.

**Claims**

|---|---|---|
| `unregister_flow` is reached on every path a pipeline can go away on | `CONFIRMED` | the only production `manager.stop()` sits in `teardown_flow` — `backend/src/state.rs:1468` — `            let result = manager.stop();` — and the sweep is that function's first statement, `backend/src/state.rs:1391` — `        crate::blocks::builtin::vision_mixer::overlay::unregister_flow(id);`; `delete_flow` arrives via `stop_flow` |
| The `FlowId::nil` fallback cannot be hit in production | `CONFIRMED` | `_flow_id` is inserted for every block unconditionally, `backend/src/gst/block_expansion.rs:107` — `            "_flow_id".to_string(),` — and that is the only site calling a builder's `build()`; the other three `get_builder` callers ask only for external pads |
| The orphan backstop cannot false-fire before the appsrc is parented | `CONFIRMED` | the timer is spawned from a setup hook, which runs inside `setup_bus_watch` — `backend/src/gst/pipeline/bus.rs:49` — `        for setup_fn in element_setups {` — called at `backend/src/gst/pipeline/lifecycle.rs:42` — `        self.setup_bus_watch();` — after elements are added and linked, and its first check is a 100 ms sleep later |
| The sweep still aims correctly for a flow edited while running | `CONFIRMED` | `retain` keys on the owner recorded at build time, `backend/src/blocks/builtin/vision_mixer/overlay.rs:1049` — `        map.retain(|_, (owner, _)| owner != flow_id);` — and a flow's id survives an edit, so ids the block list has since dropped are still swept |
| The integration test fails if the fix is reverted | `CONFIRMED` | it drives the real `start_flow` and asserts on the registry, `backend/tests/vision_mixer_failed_start_test.rs:136` — `        overlay::get_overlay_renderer(BLOCK_ID).is_none(),` — which the orphan backstop alone does not satisfy, as its doc comment says |

**Diagnosis** — the cleanup used to sit in `stop_flow`, after its call to `teardown_flow`, so a start that never reached PLAYING never ran it; at the top of `teardown_flow` it also picks up the endpoint-conflict paths and the early return where no pipeline was built. Precedent sits two lines away — `backend/src/state.rs:1456` — `        crate::blocks::builtin::mediaplayer::MEDIA_PLAYER_REGISTRY.unregister_flow(id);`. Doing it before `manager.stop()` also stops the timer pushing before the pipeline is walked to NULL.

To note, not to change: both registries stay keyed by block id, the owning flow only a value. Two flows sharing a block id would clobber each other's entry, and the older flow's timer would then exit on `still_mine` mid-run. That keying predates this PR and some twenty read sites take a block id alone, so re-keying is not this PR's job — flagged so it is not read as something the sweep introduced. Coverage `BOUNDED` to registrations whose owner.

No merge base with `main`, so the diff renders twelve files; the six `scripts/agent/*.md` entries are byte-identical — no protocol change here.

**Radius** — `SHARED`. `teardown_flow` runs for every flow, not just ones with a vision mixer; for the rest it is a no-op sweep. `API Contract Check` green.

**Tests & CI** — eight checks at `2d93be4`: six pass, macOS and Windows skipped; nothing here is platform-specific. The new test needs only core elements and the CPU compositor, so it runs where CI runs. I built and ran nothing here — static verification only.

Filed as a comment: GitHub refuses an approving review from the account that opened the PR, which this task runs under. The marker carries the verdict.

Confidence: HIGH.

PR #786, opened 2026-09-10
feat(gst): report blocks whose element chain has stopped

wagenet:
## Problem

A flow can report `gst_state: Playing` and `running: true` while a branch of its
pipeline has stopped dead, with nothing in the API, the UI, or the default log.

On a non-fatal flow return such as `not-negotiated`, `GstAggregator` sets **all**
its sink pads flushing and pauses its source pad task, posting nothing to the
bus; encoders pause their output loop the same way. Pausing a pad task is not a
state change, so every element stays `PLAYING` — and because the aggregator
flushes its inputs, the visible symptom appears far upstream of the real failure.
`GstBaseSrc` posts a flow error before pausing, so source-side failures already
reach the bus handler; the gap is the mixer-to-output path.

## What this adds

`Flow::block_health`: a `Vec<BlockHealth>` of `{ block_id, status, detail }`,
populated only while running, `status` binary `ok`/`failed` with no intermediate
tier a failed block could hide behind. It comes from a two-second scan for pads
whose task is `PAUSED` on an element that is itself `PLAYING`. Block elements are
already namespaced `block_id:internal_id`, so attribution is free; standalone
elements report under their own ID.

One `ERROR` per transition into failure and one `INFO` on recovery, plus
`StromEvent::BlockHealthChanged` on status change only — no `GST_DEBUG` needed.
The flow list shows an affected flow as **Failed** with detail on hover.

`running` is unchanged: it means "reached Playing", much logic depends on that,
and redefining it would blur "stopped" against "failed". Use
`Flow::has_failed_block()` for the distinction.

## Why it does not cry wolf

Only `Paused` counts — `gst_pad_get_task_state()` returns `Stopped` both for a
stopped task and for a pad that never had one, which is most pads. Three further
exclusions, each a real false positive rather than a hypothetical:

- **Elements below `PLAYING`** pause tasks as part of that transition.
- **Pads carrying EOS.** `gst_base_src_loop` pauses its task on the way out for
  *every* reason including end of stream, leaving the element `PLAYING`. Without
  this, any flow with a finite source reads as failed forever once it plays out.
- **Single samples.** Tasks pause briefly during flushing seeks and dynamic bin
  linking, so a block must look stalled on two consecutive scans.

## Scope

The scan reaches the flow pipeline's elements and the bins beneath them, and
nothing else. Blocks that build an isolated `gst::Pipeline` — WHIP ingest
sessions (`whip.rs`), a Media Player's decode chain (`mediaplayer/bridge.rs`) —
are siblings of that pipeline, not children, so `iterate_recurse` never enters
them. **An `ok` means "no stalled pad task among this flow pipeline's elements",
not "this flow is passing data."** Sinks in the flow pipeline are covered,
`whepserversink` included, so the path this targets is in scope. Reaching into
per-block sub-pipelines is follow-up work.

## Tests

Locally on macOS: 746 passed, 0 failed, clippy clean. All four new tests use only
`gstreamer1.0-plugins-base` elements and are visible in the `Check (Linux)` log,
so none skip. Not run on real hardware — no GPU path.

- `aggregator_stall_is_silent_and_leaves_pipeline_playing` breaks negotiation on
  a live `videotestsrc ! compositor ! capsfilter ! fakesink` and asserts the task
  reaches `Paused`, the pipeline still says `Playing`, and no bus error was
  posted. This pins the premise: if a future GStreamer posts an error here, it
  fails and says so.
- `stalled_branch_is_reported_while_the_pipeline_still_says_playing` is the
  guard, driving that stall through `PipelineManager::start()`. Verified to fail
  with the detector stubbed out.
- `a_source_that_reaches_eos_is_not_reported_as_failed` plays a finite source to
  completion and asserts it never flips. Verified to fail with the EOS check
  removed.

## Risks

- Up to 4s detection latency. The scan takes GST object locks only — no pad
  probes, nothing per-buffer.
- Auto-inserted tees are named `auto_tee_<src>` with colons replaced, so a stall
  on one reports a `block_id` absent from the flow graph. Left as-is: the mapping
  back is lossy and a `tee` runs no pad task in push mode.

### DIFF
--- a/backend/src/gst/pipeline/construction.rs
+++ b/backend/src/gst/pipeline/construction.rs
@@ -90,6 +90,8 @@ impl PipelineManager {
             cached_state: std::sync::Arc::new(std::sync::RwLock::new(PipelineState::Null)),
             qos_aggregator: QoSAggregator::new(),
             qos_broadcast_task: None,
+            block_health: std::sync::Arc::new(std::sync::RwLock::new(Vec::new())),
+            block_health_task: None,
             ptp_clock: None,
             ptp_stats: std::sync::Arc::new(std::sync::RwLock::new(None)),
             ntp_clock: None,
--- a/backend/src/gst/pipeline/health.rs
+++ b/backend/src/gst/pipeline/health.rs
@@ -49,14 +49,21 @@ fn first_stalled_pad(element: &gst::Element) -> Option<StalledPad> {
 
 /// Find the first paused pad task on `element` itself.
 ///
-/// Two kinds of legitimately paused task are excluded. An element held below
-/// `PLAYING` - a webrtcbin added for a newly connected WHEP consumer, say -
-/// pauses its tasks as part of that transition. And a pad that has seen EOS is
-/// finished by definition:
-/// `gst_base_src_loop` pauses its task on the way out for *every* reason
-/// including end of stream, and the element stays `PLAYING` afterwards, so a
-/// finite source that has played out would otherwise be reported as failed
-/// forever. The stall this looks for pushes no EOS.
+/// Three kinds of legitimately paused task are excluded.
+///
+/// An element held below `PLAYING` - a webrtcbin added for a newly connected
+/// WHEP consumer, say - pauses its tasks as part of that transition.
+///
+/// A pad that has seen EOS is finished by definition: `gst_base_src_loop` pauses
+/// its task on the way out for *every* reason including end of stream, and the
+/// element stays `PLAYING` afterwards, so a finite source that has played out
+/// would otherwise be reported as failed forever. The stall this looks for
+/// pushes no EOS.
+///
+/// An unlinked pad has no branch to report on. A `queue` feeding a block output
+/// that the flow leaves unconnected gets `not-linked` from its loop and parks
+/// the task there permanently - the vision mixer's `multiview_out` does exactly
+/// this whenever a flow uses only the program output.
 fn stalled_pad_on(element: &gst::Element) -> Option<StalledPad> {
     if element.current_state() != gst::State::Playing {
         return None;
@@ -67,6 +74,7 @@ fn stalled_pad_on(element: &gst::Element) -> Option<StalledPad> {
         .find(|pad| {
             pad.task_state() == gst::TaskState::Paused
                 && !pad.pad_flags().contains(gst::PadFlags::EOS)
+                && pad.peer().is_some()
         })
         .map(|pad| StalledPad {
             element: element.name().to_string(),
@@ -309,6 +317,46 @@ mod tests {
         flow
     }
 
+    /// videotestsrc -> tee, with one tee branch queued and left unconnected.
+    ///
+    /// Mirrors a block output the flow does not wire up - the vision mixer's
+    /// `multiview_out` is exactly this - where the queue's loop takes
+    /// `not-linked` and parks its task for good.
+    fn unwired_output_flow() -> Flow {
+        let mut flow = Flow::new("unwired output health test");
+        flow.elements = vec![
+            element(
+                "src",
+                "videotestsrc",
+                &[("is-live", PropertyValue::Bool(true))],
+            ),
+            element("split", "tee", &[]),
+            element("q_used", "queue", &[]),
+            element("sink", "fakesink", &[("sync", PropertyValue::Bool(false))]),
+            // Fed by the tee, going nowhere.
+            element("q_dangling", "queue", &[]),
+        ];
+        flow.links = vec![
+            Link {
+                from: "src".to_string(),
+                to: "split".to_string(),
+            },
+            Link {
+                from: "split".to_string(),
+                to: "q_used".to_string(),
+            },
+            Link {
+                from: "q_used".to_string(),
+                to: "sink".to_string(),
+            },
+            Link {
+                from: "split".to_string(),
+                to: "q_dangling".to_string(),
+            },
+        ];
+        flow
+    }
+
     /// videotestsrc with a fixed buffer count -> fakesink. Runs to EOS.
     fn finite_source_flow() -> Flow {
         let mut flow = Flow::new("eos health test");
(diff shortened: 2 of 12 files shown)


srperens (commented):
**Verdict: Comment** — the mechanism is right, the guard test is real and provably ran in CI, and the lifetime handling is correct. What holds it short of approval is reach: the scan walks only the flow pipeline, and the two block families that most often "stop dead while reporting Playing" run their own `gst::Pipeline` and are invisible to it.

**Requested changes**

1. Say the bound in `backend/src/gst/pipeline/health.rs`'s module doc: the scan sees only elements of the flow pipeline and their child bins, so a Media Player's decode chain and a WHIP ingest session are out of scope. Without that line an `ok` report reads as "this flow is passing data", which it does not mean.

**Claims**

|---|---|---|
| "Both new tests ... neither skips" | `CONFIRMED` | neither `backend/tests/block_health_test.rs` nor the `health.rs` test module has a `plugins_available` guard or an early return, so a missing element fails rather than skips; `.github/workflows/ci.yml:149` — `          STROM_REQUIRE_GST_PLUGINS: 1` |
| The guard would fail if the detector were reverted | `CONFIRMED` | `backend/src/gst/pipeline/health.rs:444-454` asserts `comp` reaches `Failed` through `PipelineManager::start()` and `get_block_health()` — production code, not a rebuilt topology, and nothing else in the tree sets that status |
| The task takes no strong pipeline reference | `CONFIRMED` | `backend/src/gst/pipeline/health.rs:131` — `        let weak_elements: Vec<(String, gst::glib::WeakRef<gst::Element>)> = self`, and it is stopped on both teardown paths: `backend/src/gst/pipeline/lifecycle.rs:208` and `backend/src/gst/pipeline/mod.rs:252` — `        self.stop_block_health_task();` |
| "a branch of its pipeline has stopped dead" is now reported | `CONFIRMED`, coverage `BOUNDED` | the bound is separate pipelines: `backend/src/blocks/builtin/mediaplayer/bridge.rs:28` — `    let pipeline = gst::Pipeline::builder().name(&pipeline_name).build();` and `backend/src/blocks/builtin/whip.rs:641` — `    let session_pipeline = gst::Pipeline::builder()`. A sibling pipeline is not a child bin, so `iterate_recurse()` never reaches either |
| No false positive on topologies other than the tested one | `UNVERIFIED` | the heuristic is validated against one graph (`videotestsrc ! compositor ! capsfilter ! fakesink`); it is applied to every element of every running flow, and a single false hit logs `ERROR` and turns the whole flow red — `frontend/src/app/rendering.rs:933` — `                                        "Failed"` |

**Diagnosis** — Root cause is stated correctly, and both exclusions are defended by a test rather than asserted — `backend/src/gst/pipeline/health.rs:52` — `    if element.current_state() != gst::State::Playing {` for the first, and its own test for the EOS one, which is the exclusion that would otherwise have made this unusable. CLOSED loop: the healthy path costs one lock-only scan every 2s and emits nothing.

Worth putting the separate-pipeline bound in the PR body too, because it inverts the risk you flagged. Having no WHIP rig does not matter here — the scan never enters a WHIP session bin at all. What is missing is reach, not coverage.

**Radius** — `GLOBAL`: every running flow is scanned, `Flow` gains an API-visible field, a `StromEvent` variant is added, and the flow list's colour and state text change for all flows. Contract work is complete: `BlockHealth` and `BlockHealthStatus` both carry `ToSchema`, `openapi.json` is regenerated, `API Contract Check` passes, no existing variant touched.

**Tests & CI** — run 34436702775 at `4d7c2f17`: `Check (Linux)`, both `Build (Linux)` jobs, `Check & Build (WASM)` and `API Contract Check` all `pass`. macOS/Windows `skipping`; no platform-specific code, so not a gap.

Confidence: HIGH

wagenet:
Bound documented in `fcf70e6` — module doc plus a Scope section in the PR body, both stating that `ok` means "no stalled pad task among this flow pipeline's elements", not "this flow is passing data".

The catch ran deeper than the doc line. The guard that skips elements below `PLAYING` justified itself with "a WebRTC session bin mid-setup" — those live in a session pipeline and are never scanned, so the example could not occur. Replaced with a webrtcbin added for a new WHEP consumer, which is in-pipeline.

On the `UNVERIFIED` false-positive claim: you were right, and it was not hypothetical. I built the topology you were implicitly asking about — `whip_input` → `vision_mixer` → `whep_output`, one headless publisher — and a healthy flow reported this seconds after start:

    mixer: failed — pad task paused on mixer:queue_mv_out:src

`queue_mv_out:src` is the mixer's `multiview_out` external pad. Leave it unwired and the queue takes `not-linked` from its loop and parks the task, element still `PLAYING`. Any flow using only the program output would have shown a permanently failed mixer, which is worse than the problem this PR set out to fix. Fixed in `7a01fd2` by requiring a linked peer, with a regression test that reproduces it without a rig (tee, one queued branch left unconnected) and fails when the check is reverted.

After the fix, on the same rig: 30/30 samples clean over five minutes with media confirmed flowing, and 24/24 across a publisher disconnect/reconnect cycle, which exercises WHIP session teardown and setup.

That makes three of the four exclusions this needed — sub-`PLAYING`, EOS, `not-linked` — things code reading alone did not surface. Worth stating plainly since it bears on how much the remaining `UNVERIFIED` residue is worth: a topology where a legitimately paused task persists on a `PLAYING` element, linked, having pushed no EOS. I have not found one, which is not the same as there being none.

Still not fired on a production failure. Six fresh-server trials in the known-bad encoder configuration all ran clean, so the intermittent WHIP startup failure did not reproduce on a 1-seat rig and remains unattributed.

Keeping sub-pipeline reach as follow-up. Attribution already works — `whip.rs:633` names session elements `{instance_id}:whipserversrc_{uuid}`, so `owning_block` maps them unchanged and only the task's element sourcing would move from a start-time snapshot to a live query; nothing in the types, event, contract or UI shifts. The larger reason is that sessions bridge through `appsink→appsrc`, so a stalled session presents as absence of data rather than a flow return — scanning them would catch a stall internal to one, but not the data-absence case, which wants a different signal.

— Claude

srperens (commented):
**Verdict: Comment** — supersedes my review of `4d7c2f17`. The token is unchanged but the reason is not: my one requested change is done, the new exclusion is correct and covers more than the PR body claims, and what holds this short of `Approve` is now radius alone. Everything settled at `4d7c2f17` — the separate-pipeline bound, the `WeakRef` lifetime, the EOS exclusion, the guard test — stands; do not re-raise it.

**Requested changes**

1. In `backend/src/gst/pipeline/health.rs`, state the invariant the new exclusion rests on rather than only its example. An unwired output parks on a *peerless* pad only because every tee this repo builds sets `allow-not-linked=true`. A tee added without it would send `not-linked` upstream instead, pausing an aggregator source pad that **does** have a peer — a false positive this exclusion cannot catch, and one that would be reported against the wrong block.

**Claims**

|---|---|---|
| The change requested at `4d7c2f17` is done | `CONFIRMED` | `backend/src/gst/pipeline/health.rs:14` — `//! Scope: the scan reaches the flow pipeline's own elements and the bins beneath` |
| "the vision mixer's `multiview_out` does exactly this" — the parked pad really has no peer | `CONFIRMED` | `backend/src/blocks/builtin/vision_mixer/elements.rs:133` — `        .property("allow-not-linked", true)`, so `tee_mv` absorbs the `not-linked` and the park stays on the unlinked `queue_mv_out` source pad instead of travelling up to `mv_comp` |
| The exclusion covers the class, not one block | `CONFIRMED` | all 20 `ElementFactory::make("tee")` sites in `backend/src` set the property, the auto-inserted one included — `backend/src/gst/pipeline/construction.rs:409` — `        tee.set_property("allow-not-linked", true);` |
| Runtime unlinking does not blind the detector | `CONFIRMED` | the only runtime unlink in a flow pipeline is `backend/src/gst/thumbnail_tap.rs:472` — `                let _ = tee_src_pad.unlink(&peer);` — on a `tee` source pad, which runs no task in push mode |
| The new guard executed at this head | `CONFIRMED` | `Check (Linux)` job 102953843607 at `7a01fd22`: `test gst::pipeline::health::tests::an_unwired_output_branch_is_not_reported_as_failed ... ok` |

**Diagnosis** — that a real false positive surfaced after the last review is exactly what the `UNVERIFIED` "no false positive on other topologies" row anticipated, and this commit closes the systematic instance of it rather than the one reported case: every optional block output in this tree dangles behind an `allow-not-linked` tee, so every one of them parks on a peerless pad. The residual is narrower than before — an element that legitimately parks a *linked* task — and I could not name one in this tree. Coverage of the original defect is unchanged and still `BOUNDED` to the flow pipeline.

**Radius** — `GLOBAL`, unchanged, and the only bar left to `Approve`: every running flow is scanned, `Flow` gains an API-visible field, a `StromEvent` variant is added, and the flow list's state text changes for all flows. Whether that reach is acceptable for a first cut is a human call, not a defect I found.

**Tests & CI** — run 34501704947 at `7a01fd22`: `Check (Linux)`, both `Build (Linux)` jobs, `Check & Build (WASM)` and `API Contract Check` all pass. All five health tests appear as `ok` in the `Check (Linux)` log, so none skipped silently. `Build (macOS)` and `Build (Windows)` `skipping`; no platform-specific code here, so not a gap.

Confidence: HIGH

PR #789, opened 2026-09-10
feat: add optional structured event logging and JSON stdout format

### LINKED ISSUE #772: Add structured JSON event logging for observability pipelines
## Summary

Strom already has a structured internal event model used by the real-time WebSocket API. It would be useful to optionally emit selected Strom lifecycle events as structured logs, so operators can collect them with standard log pipelines such as Kubernetes logging, OpenTelemetry Collector, Loki, Elastic, or other backends.

This would make it easier to query and group operational logs by stable fields such as `strom.flow.id`, without relying on brittle text parsing.

I have an implementation ready and tested locally. If this sounds like a good fit for Strom, I can open a PR. I am also happy to adjust the scope or split the change if preferred.

## Problem

Current logs are human-readable, but not ideal for log aggregation systems that work best with structured fields.

For example, flow-related information may appear in log messages, but the message format is not always consistent. This makes it difficult for operators to reliably group logs by flow, block, event type, or pipeline state.

Strom already emits structured `StromEvent` values internally for the WebSocket API. Those events are a good source for optional structured observability logs.

## Proposed solution

Add optional structured logging support:

```toml
[logging]
log_format = "json"
structured_events = true
include_high_frequency_events = false
```

Suggested behavior:

- `log_format = "compact"` keeps the current human-readable output.
- `log_format = "json"` emits JSON-formatted tracing logs.
- `structured_events = true` emits selected low-frequency Strom events as structured logs.
- `include_high_frequency_events = false` keeps noisy events such as system stats, metering, loudness, and frequent probe updates out of logs by default.

Example structured log:

```json
{
  "timestamp": "2026-01-01T12:00:00.000Z",
  "level": "INFO",
  "message": "Flow created",
  "event.domain": "strom",
  "event.name": "flow.created",
  "strom.event.source": "event_broadcaster",
  "strom.flow.id": "00000000-0000-0000-0000-000000000000",
  "target": "strom::event_logging"
}
```

aya-tv2:
## Description

Add optional structured logging support so operators can collect Strom lifecycle events with standard log pipelines (Kubernetes logging, OpenTelemetry Collector, Loki, Elastic, etc.) without relying on text parsing.

Three new keys under `[logging]` in `.strom.toml`:

```toml
stdout_log_format = "json"           # default: "compact"
structured_events = true             # default: false
include_high_frequency_events = false # default: false
```

At default settings (structured_events = false, include_high_frequency_events = false), structured logging is off entirely, but `is_high_frequency()` also widens what counts as high-frequency versus the old hand-picked list, which narrows the existing debug log on top of that — independent of these new flags.

The old list in `broadcast()` held 3 variants (`MeterData`, `LoudnessData`, `BufferAgeProbe`); `is_high_frequency()` classifies 11. Of the 8 newly classified variants, **5 actually pass through `EventBroadcaster::broadcast()`** and therefore change behaviour at defaults — their "Broadcasting event" line moves from `debug` to `trace`:

- `SpectrumData` (~10Hz), `AudioAnalyzerData` (~30Hz), `MediaPlayerPosition` (~5Hz) — genuinely mis-bucketed before.
- `QoSStats` — 1Hz per element, but fans out per element and flow (broadcast from the pipeline bus handler).
- `LatencyData` — tick rate is not independently verifiable; it comes from an external, unvendored GStreamer element (`audiolatency`). Classified high-frequency as the conservative default, on the assumption that a latency measurement is periodic telemetry rather than a lifecycle event.

The remaining 3 — `SystemStats`, `ThreadStats`, `PtpStats` — are built in `backend/src/api/websocket.rs` and written straight to each socket via `send_event()`; they never pass through `broadcast()`. Classifying them changes nothing today: no log line moves, and they can never appear in structured event logs at any setting, because `log_strom_event()` has exactly one caller and it is inside `broadcast()`. The classification is future-proofing for the day they move onto the event bus. This is already documented in `.strom.toml.example` beside `include_high_frequency_events`.

That `debug`→`trace` move for those 5 variants is the only behaviour difference at defaults. Nothing is dropped from `WS /api/ws`, and nothing is dropped from the MCP SSE feed — see the `mcp.rs` note below.

**`strom-types/src/events.rs`:** Added three exhaustive (no-wildcard) accessors on `StromEvent`: `event_type()`, `flow_id()`, `is_high_frequency()`. `event_type()` returns `&'static str` matching the serde `type` tag, verified by `every_variant_event_type_matches_its_serde_wire_tag`. `is_high_frequency()` is the only frequency classification in the tree — it replaces the hand-picked list in `broadcast()`, and covers the 8 variants listed above that the old list missed. Its consumers today are the broadcast log level and the structured-event gate, both in the backend log path; the MCP bridge is not one.

**`backend/src/event_logging.rs` (new):** `log_strom_event()` emits a `StromEvent` as a structured tracing log record with stable fields (`event.name`, `strom.flow.id`). Errors/warnings use `error!`/`warn!`; everything else uses `info!`.

**`backend/src/events.rs`:** Structured events emitted inline in `broadcast()` (not from a subscriber task — lossless under channel lag). Added `with_capacity()` convenience constructor for tests.

**`backend/src/api/mcp.rs`:** not touched by this PR — the rebase onto main picked up #766's rewrite of the SSE notification bridge, which replaced its old skip-list with an explicit allowlist of 7 flow/pipeline notification types (`FlowCreated`, `FlowUpdated`, `FlowDeleted`, `FlowStarted`, `FlowStopped`, `PipelineError`, `PipelineWarning`), dropping everything else. None of those 7 are high-frequency, so the widened classification cannot narrow the MCP feed, and routing the bridge through `is_high_frequency()` would be a no-op. Abandoned rather than deferred — no follow-up issue. If the allowlist is ever widened to telemetry variants, that is the point to route it through the accessor.

**`backend/src/config.rs`:** `LogFormat` enum, `parse_stdout_log_format()`, three new fields, env var handling.

**`backend/src/main.rs`:** `init_logging()` branches on compact vs. JSON for the stdout layer. File layer always stays human-readable.

`StromEvent` itself is unchanged — no new variants, no field changes, no OpenAPI or WebSocket contract churn. `#[allow(clippy::too_many_arguments)]` added to `AppState` constructors — these were already at 6–7 args; the proper fix is a config struct, out of scope here.

## Maintainer verification (local, macOS)

Run against this head in a clean worktree on macOS / aarch64 — the platform CI skips for this PR, since `Build (macOS)` is label-gated.

- `cargo check --workspace --all-targets` — clean, no warnings.
- `cargo test --workspace` — passed. All nine new tests confirmed executing by name, not skipped: `structured_events_disabled_emits_no_log_record`, `high_frequency_events_dropped_unless_included`, `logs_high_frequency_event_without_panicking`, `high_frequency_classification_matches_known_variants`, `every_variant_event_type_matches_its_serde_wire_tag`, `test_from_figment_parses_json_stdout_log_format`, `test_from_figment_rejects_invalid_stdout_log_format`, `test_from_figment_stdout_log_format_env_var`, `test_from_figment_parses_structured_events`.
- End to end, headless with `stdout_log_format = "json"` and `structured_events = true`: stdout is one JSON object per line, and `POST /api/flows` emits the structured record — `{"level":"INFO","message":"Flow <id> created","event.domain":"strom","event.name":"FlowCreated","strom.flow.id":"<id>","target":"strom::event_logging"}`.
- With `include_high_frequency_events = true` and a WebSocket client attached, no `SystemStats`/`ThreadStats`/`PtpStats` structured records are produced. That observation is what corrected the defaults breakdown above.

Not covered: `include_high_frequency_events = true` has no unit test for its enabled path — only the disabled path is asserted. The enabled path was exercised manually as described above.

## Related Issue

Fixes #772

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

- [x] I have run `cargo clippy --workspace --all-targets -- -D warnings`
- [x] I have updated documentation as needed


srperens (dismissed):
**Verdict: Request changes** — the feature is well built and correctly gated, but making `is_high_frequency()` the single source of truth silently changes two behaviours no config key controls, and the body states the opposite. Also: zero check runs.

**Requested changes**

1. Correct the claim, or narrow the reuse. `backend/src/events.rs:44` — `        if event.is_high_frequency() {` routes eight more variants to `trace!` instead of `debug!` (`SpectrumData`, `LatencyData`, `SystemStats`, `ThreadStats`, `PtpStats`, `QoSStats`, `AudioAnalyzerData`, `MediaPlayerPosition`), with all three new keys at their defaults. CLAUDE.md's troubleshooting route runs at `debug`, and those vanish from it. Either keep the old three-variant match here, or say in the body that the debug stream shrinks.
2. Same for MCP. `backend/src/api/mcp.rs:320` — `                let notification = if event.is_high_frequency()` replaces a skip list of `SystemStats`, `MeterData`, `Ping` with an eleven-variant one, so an SSE client consuming any of the other eight stops receiving them. Call that out as an intentional narrowing.
3. Guard the gate. `backend/src/events.rs:52` — `        if self.structured_events` is the switch the feature hangs on and nothing covers it. Add two cases to the `events.rs` test module: `structured_events = false` emits no record, and `include_high_frequency_events = false` drops a high-frequency one. The four new `event_logging.rs` tests assert only "without panicking".
4. Release CI before merge: `gh api -X POST repos/Eyevinn/strom/actions/runs/34480964552/approve` (maintainer only).

**Claims**

|---|---|---|
| "at their defaults, logging behavior is identical to before this change" | `CONTRADICTED` | requested changes 1 and 2 — both paths take the widened `types/src/events.rs:990-1002` list unconditionally |
| Inline emission cannot violate the BUFFER-probe rule | `CONFIRMED` | no BUFFER-probe closure calls `broadcast()`. The nearest, `backend/src/gst/buffer_age_probe.rs:315` — `        let gst_probe_id = pad.add_probe(gst::PadProbeType::BUFFER, move |pad, info| {`, only records into a slot; the event goes out from a periodic task. |
| `event_type()` stays in step with the serde tag | `CONFIRMED` | `every_variant_event_type_matches_its_serde_wire_tag` over `one_of_each_variant`, both wildcard-free, so a new variant fails to compile rather than falling through |
| `StromEvent` unchanged, no contract churn | `CONFIRMED` | only accessors added; no variant renamed, removed or refielded, `openapi.json` untouched |
| CI green | `UNVERIFIED` | run 34480964552 is `action_required` at this head — zero check runs is no evidence, not a pass |

**Diagnosis** — Right layer. Emitting inline in `broadcast()` rather than from a `subscribe()` task is the correct call: a lagging receiver drops events, and an audit log that drops under load is worse than none. Deriving the log vocabulary from the event type rather than a parallel hand-maintained table is also right — that duplication is exactly what let the two old skip lists drift. The objection above is only that fixing the drift is a second change, riding in unannounced.

One cost: `backend/src/event_logging.rs:15` — `    let description = event.description();` allocates before any level check, so with high-frequency events included and `info` filtered out you pay a `format!` per meter tick. Opt-in, so not a blocker.

**Radius** — `GLOBAL`: every `StromEvent` broadcast passes the new branch, the MCP SSE stream narrows, and `EventBroadcaster::new` grew two parameters (`with_capacity` absorbs the test call sites).

**Tests & CI** — `gh pr checks 789`: "no checks reported". Nothing here has been compiled by this repo. The checklist's local `cargo test --workspace` is the author's claim, not evidence available here.

Confidence: HIGH

aya-tv2:
Thanks for the review — addressing points 1–3 (if I understand correctly, point 4 is a maintainer action).

1 & 2. Behavior does change at defaults — here's the actual breakdown. Checked emission rates for each newly-included variant:

- SpectrumData — 10Hz default → moves debug→trace, drops from SSE
- AudioAnalyzerData — ~30Hz default → moves debug→trace, drops from SSE
- MediaPlayerPosition — 5Hz default → moves debug→trace, drops from SSE
- SystemStats, ThreadStats — exactly 1Hz → moves debug→trace, drops from SSE
- QoSStats, PtpStats — 1Hz, fan out per element/flow → moves debug→trace, drops from SSE

The first three were genuinely mis-bucketed before (multi-Hz events sitting in the debug stream and full SSE feed). The last four tick at exactly 1Hz individually but flood in aggregate once fanned out per connection/element/flow — is_high_frequency()'s doc comment now describes both cases instead of implying a single uniform rate.

Net effect: the default debug stream and MCP SSE stream both narrow to drop these 7 variants, on top of the 3 already excluded before this PR.

3. Guard the gate. Added two tests in backend/src/events.rs:
- structured_events_disabled_emits_no_log_record — structured_events = false emits zero records.
- high_frequency_events_dropped_unless_included — include_high_frequency_events = false drops a high-frequency event while still logging others.

Both use a tracing_subscriber::Layer scoped to the event_logging target so they don't conflate with the broadcaster's own trace/debug line.

Tests run: cargo test --workspace — 76 passed, 0 failed.

Known gaps:
- LatencyData's actual tick rate isn't verifiable — it comes from an external, unvendored GStreamer element (audiolatency).
- event.description() is still computed before the level check in log_strom_event (per-event allocation cost when structured logging is on). Not addressed — both flags default off, so it's opt-in cost.

srperens (dismissed):
**Verdict: Request changes** — supersedes my review at `9153ed05`, which I have dismissed. The verdict stands, but only for one sentence in the PR body and for CI: everything I asked for in code is done.

**Requested changes**

1. Correct the PR body. It still says "When all three are at their defaults, logging behavior is identical to before this change." Your comment of 2026-09-11 says the opposite, variant by variant, and the comment is the one that is right — `backend/src/events.rs:44` — `        if event.is_high_frequency() {` takes the widened list with all three keys off. Replace that sentence with the breakdown from the comment. The body is the merge record and it currently contradicts its own thread.
2. Release CI (maintainer only): `gh api -X POST repos/Eyevinn/strom/actions/runs/34570836945/approve`. Run `34480964552` from my last review is at the dead head; `34570836945` is the one at `1ef81714`.

**Resolved since `9153ed05` — do not re-raise**

- The gate is now guarded. `backend/src/events.rs:138` — `        let broadcaster = EventBroadcaster::new(10, false, false);` asserts zero records emitted; `backend/src/events.rs:154` — `        let broadcaster = EventBroadcaster::new(10, true, false);` asserts a high-frequency event is dropped while an ordinary one is logged. Both drive `broadcast()` itself and count on the `event_logging` target, so reverting either half of the condition at `backend/src/events.rs:52` — `        if self.structured_events` fails them. That is a real guard, not a demonstration.
- The narrowing is now documented at source. `types/src/events.rs:985` — `    /// Whether this event is chatty enough that logging or forwarding it by default would` covers the fan-out case, which is the honest description of the four 1Hz-but-fanned-out variants your comment separates from the three genuinely mis-bucketed ones.

**Claims**

|---|---|---|
| Body: "at their defaults, logging behavior is identical" | `CONTRADICTED` | unchanged in the body at this head; your own comment enumerates seven variants that move `debug`→`trace` and leave the MCP stream |
| The new tests fail if the gate is reverted | `CONFIRMED` | both call `EventBroadcaster::broadcast` rather than rebuilding the branch inline |
| Test build is consistent under the widened constructors | `CONFIRMED` | all eight `AppState::new` sites under `backend/tests/` now pass eight arguments; every other test reaches the broadcaster through `backend/src/events.rs:37` — `    pub fn with_capacity(buffer_size: usize) -> Self {` |
| CI green | `UNVERIFIED` | `gh pr checks 789`: "no checks reported". Run `34570836945` is `action_required` — zero check runs is no evidence, not a pass |

**Scope of this re-review** — the head moved, so I re-read the inter-head diff in full: it touches only `backend/src/events.rs`, `types/src/events.rs` and `backend/tests/vision_mixer_failed_start_test.rs`. Nothing in it reaches the rows I confirmed at `9153ed05` (no BUFFER probe calls `broadcast()`; `event_type()` is wildcard-free and tested against the serde tag; `StromEvent` and the contract are untouched), so those stand as verdicted there rather than being re-argued here.

**Radius** — `GLOBAL`, unchanged: every broadcast passes the new branch and the MCP stream narrows.

**Tests & CI** — nothing in this PR has been compiled by this repository, at either head. The arity check above is static; it is not a substitute for a build, and it is the reason requested change 2 matters more than requested change 1.

Confidence: HIGH

aya-tv2:
PR description updated with the variant-by-variant breakdown — thanks for catching the contradiction.

On the CI point: I can't run the workflow-approval step myself, so I've verified locally and on my fork instead:

$ cargo test --workspace
test result: ok. 76 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

I also enabled Actions on my fork and ran the CI workflow directly against this branch — it passed: https://github.com/aya-tv2/strom/actions/runs/34583188412

I read requested change 2 as a maintainer action (approving run 34570836945 at the current head 1ef8171, since 34480964552 was against a since-superseded commit) — flagging it here in case that's picked up in the normal review flow.

srperens (dismissed):
**Verdict: Request changes** — supersedes my review at `1ef8171`, which I have dismissed. Both requests from it are closed: the body breakdown landed and CI is green at this head with every new test confirmed executing. The verdict stands only because the rebase onto current main dropped one half of this PR, and the body still describes it as present.

**Requested changes**

1. `backend/src/api/mcp.rs` is no longer in this diff, but the body still claims the "MCP notification bridge uses `is_high_frequency()` instead of its own separate skip list". Main's MCP rewrite replaced that skip list with an allowlist — `backend/src/api/mcp.rs:417` — `                    _ => None,` — which forwards seven flow/pipeline notifications and drops everything else whatever `is_high_frequency()` says. Delete that bullet, and delete "drops from the MCP SSE feed" from **both** groups in the defaults breakdown: none of those seven variants ever reached the MCP feed, so the widened classification cannot narrow it. Then say in the body whether routing the bridge through `is_high_frequency()` is abandoned or follows separately — "single source of truth" was the argument for the accessor, and it currently has one consumer rather than two.

**Resolved since `1ef8171` — do not re-raise**

- The "logging behavior is identical to before this change" sentence is gone, replaced by the variant-by-variant breakdown. Only its MCP half is now wrong, which is request 1.
- Run 34941507518 at this head: `Check (Linux)`, `Check & Build (WASM)`, `API Contract Check` and both Linux builds `success`; Windows and macOS `skipped`, which is the label gate, not a failure.

**Claims**

|---|---|---|
| Body: the MCP bridge uses `is_high_frequency()` | `CONTRADICTED` | the accessor has three call sites on this head, all outside the MCP layer: `backend/src/events.rs:44`, `backend/src/events.rs:53` — `            && (self.include_high_frequency_events || !event.is_high_frequency())` — and `backend/src/event_logging.rs:17` — `    if event.is_high_frequency() {` |
| Body: the widened list narrows the MCP SSE stream | `CONTRADICTED` | `backend/src/api/mcp.rs:417` — `                    _ => None,` — the arms above it match seven named variants, none of them high-frequency |
| At defaults the only change is the `debug`→`trace` level of one line | `CONFIRMED` | `backend/src/events.rs:44` — `        if event.is_high_frequency() {` — is the whole default-path branch; the structured record below it is gated at `backend/src/events.rs:52` — `        if self.structured_events` |
| The new tests executed rather than skipping | `CONFIRMED` | job 104298055033 prints `... ok` for `structured_events_disabled_emits_no_log_record`, `high_frequency_events_dropped_unless_included`, `logs_high_frequency_event_without_panicking` and both accessor tests |

**Radius** — `SHARED`, **down from the `GLOBAL` I recorded at `1ef8171`**, and that follows from request 1 rather than from a change of mind: `GLOBAL` was scored for a diff where every broadcast passed the new branch *and* the MCP stream narrowed. The second half no longer exists. What is left is one log line's level plus an opt-in second record, with `StromEvent`, the REST surface and `WS /api/ws` untouched — no `openapi.json` change is in the diff, consistent with `API Contract Check` passing. If request 1 is answered by re-applying the bridge change, this returns to `GLOBAL` and needs arguing again.

**Design record** — the accessor is wildcard-free (`types/src/events.rs:991` — `    pub fn is_high_frequency(&self) -> bool {`), so a new variant is a compile error, not a silent omission. That property is why the widening is defensible as a side effect, and it holds whether or not the MCP consumer returns.

Confidence: HIGH

srperens (commented):
**Verdict: Comment** — supersedes my review at `681cabb7`, which I have dismissed. Its one request is closed: the body no longer claims the MCP bridge consumes `is_high_frequency()`, and it now says the bridge change is abandoned rather than deferred. The verdict is not Approve only because no check run exists at this head.

**Requested changes**

1. Assert the enabled path of the flag this PR adds. `backend/src/events.rs:147` — `    fn high_frequency_events_dropped_unless_included() {` — builds only `EventBroadcaster::new(10, true, false)`, so `include_high_frequency_events = true` rests on manual macOS testing alone, as the body says. Three lines in that same test close it: a second broadcaster at `(10, true, true)`, one `MeterData` broadcast, assert the counter moved.

**Needs a maintainer** — CI has never run at `df6c118a`. Run `35349488804` is sitting at `action_required`:

    gh api -X POST repos/Eyevinn/strom/actions/runs/35349488804/approve

**Claims**

|---|---|---|
| `mcp.rs` is untouched and its allowlist still drops everything unnamed | `CONFIRMED` | `backend/src/api/mcp.rs:417` — `                    _ => None,` — and `git diff origin/main..pr789 -- backend/src/api/mcp.rs` is empty |
| At defaults the only behaviour change is one log line's level | `CONFIRMED` | `backend/src/events.rs:44` — `        if event.is_high_frequency() {` — is the whole default path; the structured record below is gated at `backend/src/events.rs:52` — `        if self.structured_events` |
| The two new gate tests call the changed module and would fail if the gate were reverted | `CONFIRMED` | both construct `EventBroadcaster` and call `broadcast()`, counting records by tracing target rather than rebuilding the gate inline |
| The new tests execute in CI | `UNVERIFIED` | zero check runs at `df6c118a`; the green run quoted in my last review was at `681cabb7` |

**Overlaps — land order, not a blocker.** `is_high_frequency()` is wildcard-free: `types/src/events.rs:1003` — `            | StromEvent::BufferAgeProbe { .. } => true,` — closes the high-frequency arm and the rest are enumerated individually. That is the property that makes widening the classification defensible, and it also means **any open PR adding a `StromEvent` variant stops compiling the moment this merges**, until it adds three match arms (`event_type`, `flow_id`, `is_high_frequency`). Five do today: #755, #786, #792, #796, #805. None of them owns a design decision this PR settles, so it is merge cost rather than a conflict — but it is invisible from either diff, and whoever merges second pays it. #772, which this closes, has no triage on it.

**Radius** — `SHARED`, unchanged from my last review and for the same reason: one log line's level plus an opt-in second record, with `StromEvent`, the REST surface and `WS /api/ws` untouched. The diff carries no `openapi.json` change, consistent with the config keys being config rather than API.

**Tests & CI** — two new unit tests in `backend/src/events.rs`, pure Rust with no GStreamer element, so nothing can skip them silently once CI runs. The two commits since `681cabb7` are that test plus compile fixes for call sites the signature change missed; no production code moved.

Confidence: HIGH

srperens (commented):
**Verdict: Comment** — supersedes my review at `df6c118a` and the verdict stands unchanged, for the same two reasons: the enabled path of the flag this PR adds is still unasserted, and CI still has not run at this head. That review was `COMMENTED`, so there is nothing to dismiss. Everything since it is a rebase onto main plus one test compile fix — no production code moved.

**Requested changes**

1. Unchanged from my last review and still the whole of it: assert the enabled path. `backend/src/events.rs:147` — `    fn high_frequency_events_dropped_unless_included() {` — constructs only `(10, true, false)`, so `include_high_frequency_events = true` rests on manual testing alone. Three lines in that same test close it: a second broadcaster at `(10, true, true)`, one `MeterData` broadcast, assert the counter moved to 1.

**Needs a maintainer** — CI has never run at `a8ef21f`. Run `35569788462` is sitting at `action_required` with zero jobs:

    gh api -X POST repos/Eyevinn/strom/actions/runs/35569788462/approve

**Claims**

|---|---|---|
| The only PR-side change since `df6c118a` is a test compile fix | `CONFIRMED` | `git show a8ef21f`: one line in `backend/tests/gl_memory_link_test.rs`, `EventBroadcaster::new(10)` → `with_capacity(10)`. The other three commits in the range are #802, #842 and #843 arriving from main |
| That fix leaves the borrowed test's behaviour identical | `CONFIRMED` | `backend/src/events.rs:38` — `        Self::new(buffer_size, false, false)` — so `with_capacity` is the old two-argument default, structured logging off |
| The enabled path of `include_high_frequency_events` is exercised | `CONTRADICTED` | `backend/src/events.rs:154` — `        let broadcaster = EventBroadcaster::new(10, true, false);` — and the only other construction in the tests is `(10, false, false)` at `backend/src/events.rs:138`. Nothing builds the flag as `true`, so the branch at `backend/src/events.rs:53` — `            && (self.include_high_frequency_events || !event.is_high_frequency())` — is only ever taken one way |
| The new tests execute in CI | `UNVERIFIED` | zero check runs at `a8ef21f`; `gh pr checks 789` reports none on the branch |

**Radius** — `SHARED`, unchanged from my review of `df6c118a` and for the reason given there; the delta since is a single line in a test.

**Overlaps** — unchanged from that review (#755, #786, #792, #796, #805, merge cost rather than conflict). The property it rests on still holds at this head: `types/src/events.rs:991` — `    pub fn is_high_frequency(&self) -> bool {` — enumerates every variant with no wildcard arm, so a PR adding a variant stops compiling until it adds its arms. #843 landing on main is what forced `a8ef21f`, and the next such merge will force the next one.

**Tests & CI** — the two gate tests are pure Rust with no GStreamer element, so nothing can skip them silently once the run is approved. Until it is, "the tests pass" is unevidenced at this head rather than false.

Confidence: HIGH

srperens (approved):
**Verdict: Approve** — supersedes my review at `a8ef21f1`, and the verdict changes: both blockers are closed. That review was `COMMENTED`, so there is nothing to dismiss. The whole delta since it is one commit adding 12 lines to one test; no production code moved.

**Claims**

|---|---|---|
| The enabled path of `include_high_frequency_events` is now exercised | `CONFIRMED` (was `CONTRADICTED`) | `backend/src/events.rs:174` — `        let broadcaster = EventBroadcaster::new(10, true, true);` — broadcasts a `MeterData` and asserts `backend/src/events.rs:182` — `        assert_eq!(count.load(Ordering::SeqCst), 2);`. The counter is a `tracing` layer held by a `set_default` guard, not a channel subscriber, so a second broadcaster is counted by the same layer and the assertion is real |
| That assertion guards the branch it covers | `CONFIRMED` | `backend/src/events.rs:53` — `            && (self.include_high_frequency_events || !event.is_high_frequency())`. Both arms are now asserted in the one test: 0 after the dropped `MeterData`, 1 after the non-high-frequency event, 2 after the same `MeterData` with the flag on. Forcing the branch either way fails it |
| The new tests execute in CI | `CONFIRMED` (was `UNVERIFIED`) | `Check (Linux)` job 106272268592: `test events::tests::high_frequency_events_dropped_unless_included ... ok` and `structured_events_disabled_emits_no_log_record ... ok`. Pure Rust, no element, nothing can skip silently |
| Nothing else changed since `a8ef21f1` | `CONFIRMED` | `compare a8ef21f1...c1ba7bf0`: 1 commit, 1 file, +12/-0, all inside `mod tests` |

**Radius** — `SHARED`, unchanged from my review at `a8ef21f1` and for the reason given there.

**Overlaps** — unchanged: #755, #786, #792, #796, #805 all add `StromEvent` variants, and `types/src/events.rs:991` — `    pub fn is_high_frequency(&self) -> bool {` — still enumerates every variant with no wildcard arm, so whichever of these lands after this one stops compiling until it adds its arms. That is merge cost, not a design question left for this merge to settle, so it does not hold the verdict. Worth ordering deliberately all the same: each of those PRs pays it once.

**Tests & CI** — green at `c1ba7bf0` across `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check`, `Static JS Tests`. macOS and Windows skipping, which is right here — there is no platform-specific code in the diff. The approval I asked a maintainer for last run has happened; run 35580137353 is the one reported above, and no run is left at `action_required` for this branch.

Confidence: HIGH

PR #790, opened 2026-09-10
feat(api): lease blocks of SRT listener ports to external orchestrators

Lufftre:
## Why

A Strom shared by several Open Live instances is the only party that can keep their SRT listener ports apart. Today each orchestrator picks ports on its own, so two instances (or two venue gateways behind them) can both bind the same UDP port on the shared Strom and the second pipeline fails at start.

## What

A port lease resource. A client asks for a block of consecutive ports under a stable `client_id`, renews it while alive, and lets it lapse when gone. Asking again under the same `client_id` returns the same block renewed (200 instead of 201), so a restarting orchestrator keeps the ports its gateways already point at. A changed `size` resizes in place when the neighbouring ports are free, otherwise moves the block.

- `POST /api/port-leases`, `GET /api/port-leases`, `GET /api/port-leases/{id}`, `POST /api/port-leases/{id}/renew` (body optional), `DELETE /api/port-leases/{id}`; all behind the usual auth, all with utoipa annotations, `openapi.json` updated
- Allocation never hands out a port any known flow binds as an SRT listener (`srt_uri` with `mode=listener`), whether or not a lease covers it and whether or not the flow runs, so a lapsed lease cannot pull the rug from under a pipeline
- Pool configured with `[ports] lease_range` / `STROM_PORT_LEASE_RANGE`, default `47100-47999`
- Leases persist beside flows: `port_leases.json` next to `flows.json`, or a `port_leases` table on PostgreSQL (created in `run_migrations`)
- Default TTL 600 s, max 86400 s; expired leases are reclaimed lazily on the next call
- New types in `strom-types` (`PortRange`, `PortLease`, `PortLeaseRequest`, `RenewPortLeaseRequest`); the allocation table in `backend/src/port_lease.rs` is pure and takes `now` and the bound-port set as arguments

Consumers: Eyevinn/open-live `feat/strom-port-lease` acquires and renews a lease at boot and publishes the range to gateways; Eyevinn/open-live-ingest#1 uses it.

Docs: a short operator note in `OPEN_LIVE_SETUP.md` (with the drift disclaimer), a row in `DOCKER.md`, and a `[ports]` section in `.strom.toml.example`. No frontend changes in this PR.

## Tests

Ran:
- `cargo test -p strom-types` equivalent via `cargo test --lib` in `types` (62 passed, 4 new for `PortRange`)
- `cargo test --lib` in the backend: 550 passed, 9 new in `port_lease::tests` (lowest-block allocation, per-client idempotency, exhaustion, bound-port avoidance, expiry reclaim, resize in place and move, renew/release, validation, persistence round trip, `srt_listener_port` parsing)
- `cargo test --test port_lease_api_test`: 2 passed, end to end through the router including persistence across a second `AppState` on the same data dir and a flow already listening on a pool port
- `cargo test --test openapi_test` after updating `openapi.json`
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check`: clean

Not run: PostgreSQL storage against a live database (the new table and queries mirror the `flows` ones); anything against a live Open Live.

### DIFF
--- a/.strom.toml.example
+++ b/.strom.toml.example
@@ -87,6 +87,19 @@ port = 8080
 # cef_cache_path = "/var/cache/strom/cef"
 
 # ============================================================================
+# Port Leases
+# ============================================================================
+[ports]
+# UDP ports the port lease API (POST /api/port-leases) may hand out. External
+# orchestrators such as Open Live reserve a block each for their SRT listener
+# inputs, so several of them can share this Strom without colliding. Open the
+# whole range inbound on the firewall. Ports a flow already listens on are never
+# handed out, whether or not they lie in this range.
+# Default: 47100-47999
+# Env: STROM_PORT_LEASE_RANGE=47100-47999
+# lease_range = "47100-47999"
+
+# ============================================================================
 # Logging Configuration
 # ============================================================================
 [logging]
--- a/backend/src/api/mod.rs
+++ b/backend/src/api/mod.rs
@@ -11,6 +11,7 @@ pub mod media;
 pub mod mediaplayer;
 pub mod network;
 pub mod osc;
+pub mod port_leases;
 pub mod probes;
 pub mod sdp_transform;
 pub mod system_clock;
--- a/backend/src/api/port_leases.rs
+++ b/backend/src/api/port_leases.rs
@@ -1,0 +1,187 @@
+//! Port lease API: blocks of UDP ports reserved for external orchestrators.
+//!
+//! A Strom shared by several Open Live instances is the only party that can
+//! keep their SRT listener ports apart. Each instance asks for a block under a
+//! stable `client_id`, renews it while alive, and lets it lapse when gone.
+//! Strom does not bind the ports; it promises not to hand the block to anyone
+//! else, and never allocates a port an existing flow already listens on.
+
+use axum::{
+    body::Bytes,
+    extract::{Path, State},
+    http::StatusCode,
+    Json,
+};
+use strom_types::api::ErrorResponse;
+use strom_types::port_lease::{PortLease, PortLeaseRequest, RenewPortLeaseRequest};
+use tracing::error;
+use uuid::Uuid;
+
+use crate::json_rejection::JsonBody;
+use crate::port_lease::{Acquired, PortLeaseError};
+use crate::state::AppState;
+
+type ApiError = (StatusCode, Json<ErrorResponse>);
+
+fn lease_error(err: PortLeaseError) -> ApiError {
+    let status = match err {
+        PortLeaseError::EmptyClientId | PortLeaseError::BadSize(_) | PortLeaseError::BadTtl(_) => {
+            StatusCode::BAD_REQUEST
+        }
+        PortLeaseError::Exhausted { .. } => StatusCode::CONFLICT,
+        PortLeaseError::NotFound => StatusCode::NOT_FOUND,
+    };
+    (status, Json(ErrorResponse::new(err.to_string())))
+}
+
+fn storage_error(err: anyhow::Error) -> ApiError {
+    error!("Failed to persist port leases: {}", err);
+    (
+        StatusCode::INTERNAL_SERVER_ERROR,
+        Json(ErrorResponse::with_details(
+            "Failed to persist port leases",
+            err.to_string(),
+        )),
+    )
+}
+
+/// List live port leases.
+#[utoipa::path(
+    get,
+    path = "/api/port-leases",
+    tag = "port-leases",
+    responses(
+        (status = 200, description = "Every lease that has not expired", body = Vec<PortLease>)
+    )
+)]
+pub async fn list_port_leases(State(state): State<AppState>) -> Json<Vec<PortLease>> {
+    Json(state.list_port_leases().await)
+}
+
+/// Request a block of ports.
+#[utoipa::path(
+    post,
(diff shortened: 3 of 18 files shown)


Lufftre:
End-to-end smoke test on one machine, all three branches together (Strom `feat/port-leases`, Open Live `feat/strom-port-lease`, gateway `feat/port-range-from-open-live`), Strom on 8080 with `STROM_PORT_LEASE_RANGE=47100-47139`, Open Live against a local CouchDB, gateway `up --test` in caller mode with no `port_range`:

- Open Live acquired 47100-47109 at boot and published it on `server-info`; the gateway printed "SRT ports 47100-47109 from Open Live", registered `srt://:47100?mode=listener`, and the flow reached Playing
- a second tenant got 47110-47119; a flow created directly on Strom listening on 47125 made a 20-port request fail with 409 and a 14-port request land on 47126-47139; a 5 s TTL lease disappeared after expiry
- Open Live rejected a source and an output on port 43524 with 422 naming the range, accepted an in-range output and a caller-form source, and rejected a PATCH to an out-of-range address
- the lease renewal moved the expiry; leases survived a Strom restart with the same ranges
- SIGTERM on Open Live released its lease and a restart re-acquired 47100-47109; the gateway warned and kept its feed running meanwhile
- Ctrl-C on the gateway removed its flow and its Open Live source

srperens (commented):
**Verdict: Comment** — the design is sound, the allocation table is pure and well tested, and the contract work is complete. Two things keep it off an approve: the avoidance set sees blocks but not bare elements, and the PostgreSQL half has never been executed anywhere.

**Requested changes**

1. Widen `AppState::bound_srt_listener_ports` in `backend/src/state.rs` to walk `flow.elements` as well as `flow.blocks`, reading each element's `uri` through the same parser. A flow can bind an SRT listener with no `srt_uri` property at all (see the claim table) — exactly the case the guarantee exists to prevent. `backend/src/port_lease.rs:320` — `pub fn srt_listener_port(uri: &str) -> Option<u16> {` already returns `None` for a non-SRT scheme, so feeding it every element `uri` is safe. Add the case to `port_lease_api_test.rs`.

**Claims**

|---|---|---|
| "never hands out a port any known flow binds as an SRT listener" | `CONFIRMED`, coverage `BOUNDED` | `backend/src/state.rs:193` — `            .flat_map(|flow| flow.blocks.iter())` reads only `srt_uri` on blocks. All four SRT blocks do expose it (`mpegtssrt.rs:943`, `mpegtssrt_input.rs:616`, `efpsrt.rs:858`, `efpsrt_input.rs:959`), so that half is complete. Outside it: `types/src/flow.rs:403` — `    pub elements: Vec<Element>,` is free-form, built by factory name at `backend/src/gst/pipeline/construction.rs:313` — `        let element = gst::ElementFactory::make(&element_def.element_type)`, so a bare `srtsrc` with `uri=srt://:47110?mode=listener` binds a pool port no lease knows about |
| Every endpoint is annotated and registered | `CONFIRMED` | five `#[utoipa::path]` in `backend/src/api/port_leases.rs` and five entries at `backend/src/openapi.rs:101-105`; `API Contract Check` passes, so `openapi.json` matches |
| A restarting orchestrator keeps its block | `CONFIRMED` | keyed on `client_id`; covered by `allocates_lowest_blocks_first_and_is_idempotent_per_client` and by the API test's second `AppState` over the same data dir |
| PostgreSQL storage works | `UNVERIFIED` | the body says it was not run and `.github/workflows/ci.yml` starts no database, so nothing covers `postgres_storage.rs`. Mirroring the `flows` queries is an argument, not evidence |
| Docs follow the CLAUDE.md rules | `CONFIRMED` | `docs/OPEN_LIVE_SETUP.md` is an operator guide and carries the drift disclaimer, with no file paths in it |

**Diagnosis** — Right layer: the shared Strom is the only party that can arbitrate, so the pool belongs there rather than in each orchestrator. Keeping the allocation table pure, with `now` and the bound set passed in, is what lets nine tests cover resize-in-place, move, exhaustion and expiry without a clock or a server. Renewal is a CLOSED loop and reclaim is lazy, so an absent client costs nothing.

The bounded avoidance set is not a common-case harm — an operator working from the palette gets blocks. It matters because the failure is silent and late: the lease succeeds, gateways are pointed at the range, and the collision surfaces only when the other flow starts, which is the failure this PR exists to remove, one level down.

**Radius** — `SHARED`: `AppState` gains a lease table, the storage trait gains methods with both backends implementing them, and a `port_leases` table is created in `run_migrations`. No existing endpoint, flow path or type changes; the frontend is untouched. A wrong migration would break Postgres startup — the unverified row above.

**Tests & CI** — run 34485875920 at `646ec1f8`: `Check (Linux)`, both `Build (Linux)` jobs, `Check & Build (WASM)` and `API Contract Check` all `pass`. The new tests need no GStreamer element and no hardware, so nothing skips silently. macOS/Windows `skipping`; no platform-specific code here.

Confidence: HIGH

Lufftre:
Closing: the shared deployments will put a proxy in front of Strom that routes `/api/port-leases` to a separate port broker service, so Strom stays out of port allocation. The allocation module and its tests from this branch move to that service unchanged; the Open Live (Eyevinn/open-live#229) and gateway (Eyevinn/open-live-ingest#1) changes are unaffected because the wire contract is the same. Branch kept for reference.

PR #791, opened 2026-09-10
feat(whip): make drop-on-latency configurable on WHIP and WHEP Input

wagenet:
## Problem

`drop-on-latency` is set unconditionally on the per-session `rtpbin` inside WHIP Input, and on both WHEP Input transports. There is no way to turn it off.

It is there for a reason and stays on by default: after a mute gap, `rtpjitterbuffer`'s `calculate_packet_spacing` reads the large RTP timestamp jump as huge packet spacing, corrupting lost-timer scheduling so packets are held for the length of the gap instead of released. Setting `drop-on-latency` on `rtpbin` propagates to its jitterbuffers, which then drop queued packets exceeding the configured latency and break the stall. Upstream: [gst-plugins-good MR 951](https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/merge_requests/951).

The cost shows up when Strom sits between two WebRTC endpoints. A middle box with a small jitterbuffer *and* `drop-on-latency` discards late packets the receiving browser could still have used and would have absorbed with its own adaptive buffer. What that path wants is a small buffer that passes everything, letting the endpoint absorb jitter once. That combination is not expressible today.

## Change

A `drop_on_latency` block property, default `true`, so existing flows behave exactly as they do now. Construction-time, not live, following `do_retransmission`:

- **WHIP Input**: `parse_drop_on_latency` → `WhipEndpointConfig.drop_on_latency` → applied in `create_whipserversrc_for_session`'s `deep-element-added` handler.
- **WHEP Input**: applied at both sites, the `iterate_recurse` sweep in `build_whepsrc` and the `deep-element-added` handler in `build_whepclientsrc`.

The workaround is unchanged and the default is unchanged. No new shared types, so the OpenAPI schema is untouched.

## Tests

Added:

- `drop_on_latency_defaults_to_true` and `drop_on_latency_respects_explicit_false` in both modules — parse-level.
- `drop_on_latency_reaches_whip_endpoint_config` — a real guard. It calls `build_whipserversrc` and asserts the value in the `WhipEndpointConfig` handed to the session manager. Verified it fails on revert: replacing the parse call with a literal `true` produces `left: true, right: false`.

That guard covers property → config. The last hop, config → `rtpbin`, is **not** covered: the `rtpbin` only exists inside `whipserversrc` after a session negotiates, so reaching it needs a live WHIP publisher with ICE. The existing `do_retransmission_reaches_whip_endpoint_config` stops at the same boundary.

The two WHEP Input sites have **no** guard beyond the parse tests. Both apply the value inside a `whepsrc` / `whepclientsrc` element tree that needs a live WHEP server to populate. Hardcoding either back to a literal would not fail any test in this PR. Called out rather than papered over with a test that always passes.

### Run locally (macOS, Darwin 25.6.0)

| Command | Result |
| --- | --- |
| `cargo test --lib` | 581 passed, 0 failed |
| `cargo test --test openapi_test` | 1 passed |
| `cargo clippy --all-targets` | clean |
| `cargo fmt --check` | clean |

Not run: the integration tests under `backend/tests/`, and anything requiring a live WHIP or WHEP peer. No manual verification against a real publisher.

### DIFF
--- a/backend/src/blocks/builtin/whep.rs
+++ b/backend/src/blocks/builtin/whep.rs
@@ -183,6 +183,21 @@ fn parse_do_retransmission(properties: &HashMap<String, PropertyValue>) -> bool 
         .unwrap_or(true)
 }
 
+/// Parse drop_on_latency from properties (default: true).
+///
+/// True works around a GStreamer rtpjitterbuffer bug (see `build_whepsrc`'s
+/// iterate_recurse). False keeps late packets for a downstream WebRTC endpoint
+/// that buffers adaptively, and reinstates the stall.
+fn parse_drop_on_latency(properties: &HashMap<String, PropertyValue>) -> bool {
+    properties
+        .get("drop_on_latency")
+        .and_then(|v| match v {
+            PropertyValue::Bool(b) => Some(*b),
+            _ => None,
+        })
+        .unwrap_or(true)
+}
+
 /// Migrate a legacy `mode` property on a WHEP Output block to explicit
 /// `num_audio_tracks` / `num_video_tracks` counts, and drop `mode`.
 ///
@@ -317,6 +332,7 @@ fn build_whepsrc(
             }
         })
         .unwrap_or(DEFAULT_JITTERBUFFER_LATENCY_MS as u32);
+    let drop_on_latency = parse_drop_on_latency(properties);
 
     // Create namespaced element IDs
     let instance_id_owned = instance_id.to_string();
@@ -369,13 +385,15 @@ fn build_whepsrc(
             // of the mute gap instead of being output immediately.
             // Setting drop-on-latency on rtpbin propagates to all its
             // jitterbuffers, making them drop queued packets that exceed the
-            // configured latency — breaking the stall.
+            // configured latency — breaking the stall. Configurable because the
+            // workaround costs late packets a downstream WebRTC endpoint could
+            // still have used.
             // Upstream: https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/merge_requests/951
             if name.starts_with("rtpbin") && element.has_property("drop-on-latency") {
-                element.set_property("drop-on-latency", true);
+                element.set_property("drop-on-latency", drop_on_latency);
                 info!(
-                    "WHEP Input (whepsrc): Set drop-on-latency=true on existing {}",
-                    name
+                    "WHEP Input (whepsrc): Set drop-on-latency={} on existing {}",
+                    drop_on_latency, name
                 );
             }
         }
@@ -586,6 +604,7 @@ fn build_whepclientsrc(
             }
         })
         .unwrap_or(DEFAULT_JITTERBUFFER_LATENCY_MS as u32);
+    let drop_on_latency = parse_drop_on_latency(properties);
 
     // Create namespaced element IDs
     let instance_id_owned = instance_id.to_string();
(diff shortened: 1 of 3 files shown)


srperens (commented):
**Verdict: Comment** — the change is right, the default genuinely preserves today's behaviour, and the test gap is declared instead of papered over. One thing to add to the body: this repo sets the same workaround in a fourth place that stays hardcoded, and a reader of the problem statement will not find it.

**Requested changes**

1. Name the AES67 Input site, and say why it keeps the workaround. `backend/src/blocks/builtin/aes67.rs:240` — `                element.set_property("drop-on-latency", true);` is the same unconditional `rtpbin` workaround, carrying the same comment pointing at `build_whepsrc`. The body's problem statement enumerates WHIP Input and both WHEP Input transports, so a reader takes that as the full list. I think leaving it is the right call — the argument for `false` is "a downstream WebRTC endpoint will buffer adaptively", and an AES67 receiver on a PTP-timed playout is not that — but it should be a stated decision, since the next person to hit the mute-gap stall on AES67 will look here first.

**Observation, not a change request** — the 320-character operator description is byte-identical in `whip.rs` and `whep.rs`; `do_retransmission` has only one copy. It will drift the first time someone edits one. Not worth churn now, worth knowing.

**Claims**

|---|---|---|
| The default is unchanged at all three converted sites | `CONFIRMED` | each reads a parser ending `.unwrap_or(true)`; `backend/src/blocks/builtin/whip.rs:721` — `                element.set_property("drop-on-latency", drop_on_latency);`, and the same at `backend/src/blocks/builtin/whep.rs:393` and `backend/src/blocks/builtin/whep.rs:747` |
| `drop_on_latency_reaches_whip_endpoint_config` is a guard, not a demonstration | `CONFIRMED` | `backend/src/blocks/builtin/whip.rs:2085` — `    fn drop_on_latency_reaches_whip_endpoint_config() {` drives `build_whipserversrc` itself and asserts on the config it registers, over both `true` and `false`, so a literal in place of the parse fails the second case. It has no skip guard and `Check (Linux)` is `success` at `c60b9647`, so it ran |
| The new closure capture adds no lifetime risk | `CONFIRMED` | `backend/src/blocks/builtin/whip.rs:705` — `    let drop_on_latency = config.drop_on_latency;` copies a `bool` out before the handler is built, so the closure captures no `Element`, `Bin` or `Pipeline` — the CLAUDE.md rule, respected the same way `jitterbuffer_latency_ms` already is |
| The two WHEP sites are unguarded | `CONFIRMED` | the value is consumed only inside the element-tree callbacks, so nothing observable survives without a live peer. The body states this plainly, which is what CLAUDE.md's Tests section asks for when a real guard is not feasible |

**Diagnosis** — Right layer, right default. This is not a bug fix: the workaround stays and an operator can now decline it, so no existing flow's behaviour moves — the new path is reachable only by setting the property.

**Radius** — `SHARED`: two block definitions gain a property, `WhipEndpointConfig` gains a field (`backend/src/whip_session_manager.rs:51` — `    pub drop_on_latency: bool,`) so every construction site had to be updated, and every WHIP or WHEP Input flow now runs the parse. A regression surfaces only for a flow that sets the property to `false` — as the mute-gap stall the workaround exists to prevent, which is exactly the trade the operator opted into. `live: false` means it takes effect at flow start, matching `do_retransmission`.

**Tests & CI** — run `34515783935` at `c60b9647`: `Check (Linux)`, both Linux builds, WASM, `API Contract Check` all `pass`. macOS and Windows `skipping`, correct — nothing platform-specific. The body's local macOS run is the author's claim; what I confirm is the Linux run above. I built and ran nothing here.

Confidence: HIGH

PR #792, opened 2026-09-10
feat(whip): track a slot's liveness per medium

wagenet:
## Base

This does not sit on `main`. Reviewable content is the last commit only:

```
main
 └─ #753  wagenet/whip-slot-takeover
     └─ #756  wagenet/whip-liveness-decoded
         └─ merge of #796  wagenet/unlinked-branch-health-pr  (carries #786)
             └─ feat(whip): track a slot's liveness per medium   ← this PR
```

#796 is a real dependency, not incidental: this reports through the block health scan, which #786 adds and #796 builds on. `scan_block_health` now takes the pending-link list alongside the liveness reporters, and a block's own verdict is applied last, so the more specific failure still wins when more than one fires.

## The problem

A WHIP slot's audio and video are separate chains from the publisher's RTP session to the flow's consumers, but a single `ActivityStamp` covered both. Either medium stamping it kept the whole seat looking alive.

On the five-seat mix-minus rig a seat delivered no audio to the audio mixer, the vision mixer or the return router for its whole session while its video kept flowing, and every liveness signal reported it healthy. That is a commentator whose microphone has silently stopped reaching the mix.

## What this does

`SlotOutput` replaces the per-slot stamp with one stamp per medium, written by the same tee probes. `WhipSlotLiveness` names any occupied slot medium that has stopped producing while its counterpart has not.

That verdict reaches block health through a new `BlockLiveness` hook on `BlockBuildContext`. Neither existing scan can see this failure: the pad-task scan needs a task that was paused, the unformed-link scan needs a link that never formed, and a chain that is linked and simply never pushed to has neither. A block that can judge itself now says so, and the existing two-scan confirmation and edge-triggered logging apply unchanged. The operator gets a `BlockHealthChanged` event, an entry in `flow.block_health`, and:

```
ERROR Block 'whip_p1' in flow 'race day' is not passing data: WHIP endpoint 'p1':
slot 0 has produced no audio for 5.6 s while its video is still flowing.
The pipeline still reports Playing
```

## Reap policy: unchanged, deliberately

Session-level idle still folds the media together on the newest of them. Neither the 10 s inactivity watchdog nor takeover behaves differently than before this PR.

Reaping on one dead medium drops the video the publisher is still delivering, and throws off a client that merely renegotiated its microphone away, leaving it to reconnect in a loop. Worse, a per-medium idle in takeover reopens the bug that evicted a healthy video-only publisher at 3.1 s: `TAKEOVER_IDLE_THRESHOLD` is two seconds because *audio* has a ~20 ms cadence, and it is only sound as a whole-session test because `has_delivered_audio` gates it. What to do about a dead microphone is a production decision, not one the reaper should make silently.

## Thresholds

| medium | silent before reported |
|---|---|
| audio | 2 s |
| video | 10 s |
| never produced at all | 10 s, from the counterpart's first buffer |

A connected publisher sends audio every ~20 ms even with its microphone muted, since WebRTC keeps the RTP stream running and fills it with silence, so a couple of seconds of nothing is already a broken chain. A screen share of a window nobody is touching legitimately goes many seconds between frames.

A medium is only reported when its counterpart outlived it by that medium's own budget. Without that, a seat losing its transport reads as half dead for the difference between the two budgets. The rig produced exactly that, reporting a killed publisher as an audio failure for the 5.6 s before the watchdog reaped it.

## Risk

A publisher that legitimately sends video only, a screen share with no microphone, on an endpoint configured `audio_video` is reported after ~14 s and stays reported. That is the same code path that catches the rig failure, where the audio never arrived either, so it cannot be suppressed without losing the case this PR exists for. The endpoint's mode is the operator's declaration of what the seat should carry, so I think reporting is right, but it is worth a second opinion.

Telling "the publisher never sent audio" from "audio arrived and never came out of the slot's chain" needs the session's ingress stamps, which are not reachable from the block. Not in this PR.

## Tests

`a_slot_reports_the_medium_that_died_while_the_other_kept_flowing` builds a real `audio_video` block, claims the slot, pushes buffers into the real per-slot appsrcs and reads the real reporter, covering audio never arriving, both media healthy, and audio stopping while video continues. **Fails if the fix is reverted** — wiring both probes back to one stamp makes video buffers stamp audio; re-verified by doing that on this branch. `a_seat_that_stopped_altogether_is_not_reported_as_half_dead` guards the counterpart-freshness rule and **fails if that term is dropped**, also re-verified here. `an_empty_or_wholly_idle_slot_is_not_reported` and `a_blocks_own_failure_report_reaches_its_health` cover the quiet slots and the merge into the scan. `a_session_whose_audio_died_keeps_its_slot` pins the reap policy; it passes before the change too, so it guards the decision, not the fix.

Ran on this branch: `cargo test --workspace` (33 binaries, 768 tests, 0 failed, the 2 pre-existing ignored ones aside), `cargo clippy --workspace --all-targets` and `cargo fmt --check`, all green. `openapi_test` is included and unaffected — no API-visible type changed. `pipeline_lifecycle_test` is included.

Not re-run after the re-cut: the headless rig below. Those numbers are from the earlier build of this commit, before it was re-applied onto the current #756 and #796.

| case | result |
|---|---|
| video-only publisher | reported 14.1 s after slot allocation, seat kept its slot, `flow.block_health` carried the detail |
| publisher killed outright | no report, watchdog reaped at 10 s as before |
| slot released | block returned to `ok` |

Never run: the full mix-minus rig with tone capture. The half the headless run did not reach, audio that produces and then stops, is what the unit test drives through the real probes.

### DIFF
--- a/backend/src/api/whip_ingest.rs
+++ b/backend/src/api/whip_ingest.rs
@@ -7,7 +7,7 @@
 use crate::api::sdp_transform::{
     add_goog_remb, fix_video_bitrate_hints, strip_cvo_extension, strip_redundancy_codecs,
 };
-use crate::blocks::builtin::whip::create_whipserversrc_for_session;
+use crate::blocks::builtin::whip::{create_whipserversrc_for_session, CreatedSession};
 use crate::json_rejection::JsonBody;
 use crate::state::AppState;
 use crate::whip_session_manager::{NewWhipSession, WhipSessionManager};
@@ -132,11 +132,15 @@ pub async fn whip_post(
     // Allocate a slot for this session (pre-allocate with a temporary resource_id,
     // will be updated when we learn the real resource_id from the Location header)
     let temp_resource_id = uuid::Uuid::new_v4().to_string();
-    let slot = match config.allocate_slot(&temp_resource_id) {
+    let slot = match state
+        .whip_session_manager()
+        .allocate_slot_or_take_over(&config, &temp_resource_id)
+        .await
+    {
         Some(s) => s,
         None => {
             warn!(
-                "WHIP endpoint '{}': all {} slots occupied, rejecting client",
+                "WHIP endpoint '{}': all {} slots occupied by live sessions, rejecting client",
                 endpoint_id, config.max_sessions
             );
             return (
@@ -155,7 +159,12 @@ pub async fn whip_post(
     let cleanup_tx = state.whip_session_manager().cleanup_sender();
     let cleanup_sent = Arc::new(AtomicBool::new(false));
     let cleanup_sent_for_session = cleanup_sent.clone();
-    let (element, session_pipeline, port) = match tokio::task::spawn_blocking(move || {
+    let CreatedSession {
+        element,
+        session_pipeline,
+        port,
+        activity,
+    } = match tokio::task::spawn_blocking(move || {
         create_whipserversrc_for_session(
             &config_for_session,
             slot,
@@ -369,6 +378,7 @@ pub async fn whip_post(
                         endpoint_id: endpoint_id.clone(),
                         slot,
                         cleanup_sent,
+                        activity,
                     });
                 if !registered {
                     warn!(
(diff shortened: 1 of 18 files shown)


srperens:

PR #793, opened 2026-09-11
feat(stats): report WHIP Input jitterbuffer stats per seat

wagenet:
## The gap

`GET /api/flows/{id}/rtp-stats` covered AES67 Input only. A flow whose inputs are WHIP seats returned an empty block list, so there was no way to see what jitter a publisher actually delivers or whether its jitterbuffer is sized for it. `jitterbuffer_latency_ms` is the largest single term in a WHIP seat's end-to-end delay, and today it has to be chosen blind.

## What this does

The collector reports one set of jitterbuffer statistics per occupied slot on a WHIP Input block.

A WHIP publisher's receive path is not in the flow's pipeline: each session runs its own, bridged over appsink/appsrc. `WhipSessionManager::sessions_for_endpoint` hands the collector those pipelines, matched on the endpoint id the running pipeline registered in `runtime_data`, falling back to the block property.

Stats are keyed `slot<n>_<medium>_`, because a seat's audio and its video buffer independently and usually only one of them is the problem:

```
whip_p1 -> slot0_audio_num_pushed=878  slot0_audio_num_late=0  slot0_audio_avg_jitter_ns=3667774  slot0_audio_latency_ms=150
           slot0_video_num_pushed=7440 slot0_video_num_late=0  slot0_video_avg_jitter_ns=42969    slot0_video_latency_ms=150
```

The medium cannot be read off the jitterbuffer's own pads. Inside `webrtcbin` the caps that reach it are stripped to `application/x-rtp, ssrc=(uint)N`. It is resolved by following the stream downstream to the first caps that name a medium, which in practice is the depayloader's output. A buffer whose stream is not linked yet falls back to its element name, which unlike the enumeration order does not move between polls.

`RtpJitterbufferStats` gains `latency_ms`, the buffer's configured size. Measured jitter is only actionable next to the buffer it is measured against. AES67 Input reports it too now.

`openapi.json` is unchanged: `RtpJitterbufferStats` never appears in the schema, which carries these through the generic `Statistic` and `StatValue` types.

## Tests

Four new tests over the collection path, all executed and passing. Two seats on one endpoint plus a third on another assert that both slots report their configured buffer and that the other endpoint's seat does not leak in; an endpoint with no publisher reports no phantom seat; a block whose id has not reached `runtime_data` falls back to the property. Four more cover the medium resolution. The walk test gives the element under test webrtcbin's SSRC-only caps and puts the medium two elements downstream; it fails if the peer hop is removed.

Deleting the collector's WHIP arm fails two of them, so they guard the change rather than demonstrate it. They need `rtpjitterbuffer` and `capssetter` from `gst-plugins-good`, already in the CI package list.

Also ran on this branch after rebasing onto `main`: `cargo test` (39 test binaries including `openapi_test`, 830 passed; the two ignored tests are unrelated and opt-in), `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`.

Verified on a running rig: a five-seat mix-minus flow with two `whipclientsink` publishers produced the output above, and the block list emptied once the publishers disconnected and their slots were released.

No test exercises a real `webrtcbin` graph, since a WHIP session needs a live publisher. The unit tests use a real session manager and a real `rtpjitterbuffer`, and the walk test reproduces webrtcbin's stripped caps, but the walk has run against a real webrtcbin graph on the rig only.

### DIFF
--- a/backend/src/state.rs
+++ b/backend/src/state.rs
@@ -2764,7 +2764,8 @@ impl AppState {
     }
 
     /// Get RTP statistics for a running flow.
-    /// Returns jitterbuffer statistics from RTP-based blocks like AES67 Input.
+    /// Returns jitterbuffer statistics from RTP-based blocks: AES67 Input, and
+    /// WHIP Input for each seat currently publishing.
     pub async fn get_flow_rtp_stats(
         &self,
         flow_id: &FlowId,
@@ -2780,6 +2781,7 @@ impl AppState {
         Some(StatsCollector::collect_flow_stats(
             pipeline.pipeline(),
             flow,
+            &self.inner.whip_session_manager,
         ))
     }
 
--- a/backend/src/stats/collector.rs
+++ b/backend/src/stats/collector.rs
@@ -1,12 +1,17 @@
 //! Statistics collector for running pipelines.
 
-use crate::stats::rtp::collect_all_jitterbuffer_stats;
+use crate::stats::rtp::{
+    collect_all_jitterbuffer_stats, collect_rtp_jitterbuffer_stats, find_jitterbuffers_in_bin,
+    jitterbuffer_media_kind,
+};
+use crate::whip_session_manager::WhipSessionManager;
 use gstreamer as gst;
 use gstreamer::prelude::*;
 use std::time::{SystemTime, UNIX_EPOCH};
 use strom_types::block::BlockInstance;
 use strom_types::stats::{BlockStats, FlowStats, Statistic};
 use strom_types::Flow;
+use strom_types::PropertyValue;
 use tracing::{debug, trace, warn};
 
 /// Collector for pipeline statistics.
@@ -14,7 +19,14 @@ pub struct StatsCollector;
 
 impl StatsCollector {
     /// Collect statistics for a running flow.
-    pub fn collect_flow_stats(pipeline: &gst::Pipeline, flow: &Flow) -> FlowStats {
+    ///
+    /// `whip_sessions` is needed because WHIP Input's receive path is not in
+    /// `pipeline`: each publisher gets its own session pipeline.
+    pub fn collect_flow_stats(
+        pipeline: &gst::Pipeline,
+        flow: &Flow,
+        whip_sessions: &WhipSessionManager,
+    ) -> FlowStats {
         let now = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .unwrap_or_default()
@@ -24,7 +36,7 @@ impl StatsCollector {
 
         // Collect stats for each block in the flow
         for block in &flow.blocks {
-            if let Some(stats) = Self::collect_block_stats(pipeline, block) {
+            if let Some(stats) = Self::collect_block_stats(pipeline, block, whip_sessions) {
                 block_stats.push(stats);
             }
         }
@@ -38,7 +50,11 @@ impl StatsCollector {
     }
 
     /// Collect statistics for a specific block.
-    fn collect_block_stats(pipeline: &gst::Pipeline, block: &BlockInstance) -> Option<BlockStats> {
+    fn collect_block_stats(
+        pipeline: &gst::Pipeline,
+        block: &BlockInstance,
+        whip_sessions: &WhipSessionManager,
+    ) -> Option<BlockStats> {
         let now = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .unwrap_or_default()
@@ -47,6 +63,7 @@ impl StatsCollector {
         // Determine what kind of stats to collect based on block definition
         let stats = match block.block_definition_id.as_str() {
             "builtin.aes67_input" => Self::collect_aes67_input_stats(pipeline, &block.id),
+            "builtin.whip_input" => Self::collect_whip_input_stats(block, whip_sessions),
             "builtin.aes67_output" => {
                 // AES67 output doesn't have jitterbuffer stats, could add other stats later
                 vec![]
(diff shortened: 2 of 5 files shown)


srperens (dismissed):
**Verdict: Request changes** — the collection path is at the right layer and its three collector tests are real guards that demonstrably executed. The blocker is one claim: the test presented as covering the downstream caps walk never takes a hop, so the walk itself is unguarded.

**Requested changes**

1. Make `media_kind_is_found_downstream_of_the_element` actually hop. It asserts on `audiotestsrc` — `backend/src/stats/rtp.rs:193` — `        assert_eq!(jitterbuffer_media_kind(&src).as_deref(), Some("audio"));` — but the loop reads the element's **own** first linked src pad before any peer hop — `backend/src/stats/rtp.rs:78` — `        if let Some(kind) = pad.current_caps().as_ref().and_then(media_kind_from_caps) {`. In `Paused` that pad carries sticky `audio/x-raw` caps, whose media-type half alone yields `audio`, so it returns on iteration one. The comment above the assert — `backend/src/stats/rtp.rs:192` — `        // Two hops from `src`, so the caps are not on its own pad.` — is not what happens: `pad.peer()`, `parent_element()` and `MEDIA_KIND_MAX_HOPS` are reached by no test. Seed the element under test with caps naming no medium — webrtcbin's `application/x-rtp, ssrc=(uint)N` — and put the medium-naming caps a peer hop away, e.g. a depayloader's output.

2. Correct the Tests section. "Four more cover the medium resolution, including the two-hop case webrtcbin creates" is contradicted by 1, and the body's own closing paragraph already says the walk "was confirmed against webrtcbin's stripped caps on the rig only". Keep the second sentence, drop the first.

**Claims**

|---|---|---|
| The medium tests cover the two-hop case webrtcbin creates | `CONTRADICTED` | requested change 1 |
| The new tests executed rather than skipped | `CONFIRMED` | job 103110133450 — `test stats::collector::tests::whip_input_reports_the_jitterbuffer_of_each_connected_seat ... ok`, and the seven others |
| Deleting the collector's WHIP arm fails two of them | `CONFIRMED` | the block then reports nothing, so the two-seat test's `.expect` on its presence fires, and the property-fallback test's `assert_eq!(.., 1)` fails |
| `openapi.json` is unchanged because `RtpJitterbufferStats` never appears in it | `CONFIRMED` | `grep -c` over `openapi.json` → `0`; the field travels as a generic `Statistic` |
| The rig produced the `slot0_audio_*` / `slot0_video_*` output shown | `EXTERNAL` | no WHIP publisher or webrtcbin graph here; assumed |

**Diagnosis** — Right layer: a seat's `whipserversrc` really does live in its own pipeline, so the flow's pipeline cannot reach these jitterbuffers; the session manager is the only route. Coverage is `BOUNDED`.

**Radius** — `SHARED`. Both collector entry points gain a parameter, but the only new behaviour is one match arm on `builtin.whip_input`; AES67 changes only by gaining `latency_ms`. The new `element.property::<u32>("latency")` — `backend/src/stats/rtp.rs:38` — would panic on an element lacking it, but the factory-name guard returns first.

**Lifetime** — no closure, probe, timer or thread, so the `WeakRef` rule does not bite. `sessions_for_endpoint` does hand out strong pipeline clones — `backend/src/whip_session_manager.rs:419` — `                pipeline: s.session_pipeline.clone(),` — but the `Vec` is a temporary the loop consumes, so the overlap with a teardown is one poll.

**One observation** — the medium is part of the stat id, so when resolution fails the id embeds GStreamer's auto-generated element name, which differs between processes. Every collector test takes that fallback, its fixture buffer being unlinked.

**Tests & CI** — Linux, WASM, API Contract and both builds green at `59d2b10`; Windows and macOS `skipping`, fine for platform-independent Rust.

Confidence: HIGH

srperens (approved):
**Verdict: Approve** — supersedes my `Request changes` of 2026-09-16 (review-5220265638, head `59d2b10e`), which **no longer stands**. Both requested changes are in, and the first was done exactly as asked.

**Claims**

|---|---|---|
| The caps walk is now genuinely guarded, not just demonstrated | `CONFIRMED` | `backend/src/stats/rtp.rs:211` — `        assert_eq!(own.as_ref().and_then(media_kind_from_caps), None);` establishes the element's own pad names no medium, so the `Some("audio")` assert two lines later can only pass by hopping. The fixture seeds webrtcbin's stripped caps at `backend/src/stats/rtp.rs:185` — `            .field("ssrc", 1u32)` and puts the medium on a `capssetter` downstream. Remove the peer hop and it returns `None` |
| The Tests section no longer overclaims | `CONFIRMED` | the body now reads "it fails if the peer hop is removed" and keeps the closing paragraph conceding the walk has met a real `webrtcbin` only on the rig |
| The new tests executed rather than skipped | `CONFIRMED` | `Check (Linux)` job 105240398712: all six `stats::rtp::tests::*` and all three `stats::collector::tests::*` report `... ok`, including `media_kind_is_found_downstream_of_the_element` |
| The rig produced the `slot0_audio_*` / `slot0_video_*` output shown | `EXTERNAL` | no WHIP publisher or `webrtcbin` graph is reachable from this repository; assumed, as before |

**Diagnosis** — unchanged from my superseded review and still right: a seat's `whipserversrc` runs in its own session pipeline, so the flow's pipeline cannot reach these jitterbuffers and the session manager is the only route. Coverage is `BOUNDED` to jitterbuffers reachable from a registered session.

**Lifetime** — `backend/src/whip_session_manager.rs:681` — `                pipeline: s.session_pipeline.clone(),` still hands out strong pipeline clones, but the `Vec` is a temporary the stats loop consumes and drops. No closure, probe, timer or thread is created, so the CLAUDE.md `WeakRef` rule does not bite. The worst case is that a session torn down mid-poll has its finalization deferred until the collector's `Vec` drops, which is bounded by one pass of the loop rather than by anything a publisher controls.

**Radius** — `SHARED`, unchanged: both collector entry points gain a `WhipSessionManager` parameter, but the only new behaviour is one match arm on `builtin.whip_input`. AES67 changes only by gaining `latency_ms`. `backend/src/stats/rtp.rs:38` — `    let latency_ms = element.property::<u32>("latency") as u64;` would panic on an element without that property, but the factory-name guard above it returns first.

**One observation, not a blocker** — the medium is part of the stat id, and the fallback when resolution fails is the element's auto-generated name (`backend/src/stats/collector.rs:171` — `                let medium = jitterbuffer_media_kind(jb).unwrap_or_else(|| jb.name().to_string());`). So a jitterbuffer polled before its stream links reports `slot0_rtpjitterbuffer0_*` and then switches to `slot0_audio_*` once caps appear. A consumer keying on the stat id sees that as a new series rather than a continuation. Worth knowing before anything graphs these; it does not affect correctness of the values.

Confidence: HIGH

PR #795, opened 2026-09-11
feat(liveaudiorouter): give an output bus headroom and a ceiling

wagenet:
`builtin.liveaudiorouter` sums every crosspoint routed to an output with nothing on the bus to catch the result. In a mix-minus return, two seats talking over each other can go over full scale when neither alone does. `MAX_CROSSPOINT_GAIN` already named the exposure, but only covered boost.

## What this adds

Two construction-time properties per output bus. Both default to no-ops.

- **`output_fader_db`** is a bus master fader: a `volume` from 0 down to -24 dB.
- **`output_soft_clip_enabled`** adds an `rglimiter`. It has no attack or release; it is a fixed per-sample `tanh` curve with an asymptote at full scale, so it is named for what it does. The mixer block's Main Limiter is a true limiter and behaves differently.

When either stage is engaged, `caps_out_O` pins `F32LE`. `audiomixer` saturates the sum in a fixed-point format: two in-phase 0.8 sines give +4.08 dBFS with F32 downstream and exactly 0.00 dBFS with S16. Without the pin, whether an overload survived to reach the fader depended on what the next block negotiated.

A router that asks for neither is unchanged down to negotiation: no clipper element, the fader at unity, and no format field on `caps_out_O`.

## How much headroom

The help text suggests **-3 dB with the soft clipper on, whatever the seat count** (`routing::SUGGESTED_OUTPUT_FADER_DB`). The fader is sized for two voices, because that is what overlap in conversation looks like:

- **Two-party:** simultaneous speech is 3.8% of Switchboard, with a median overlap of 205 ms. 73% of sampled overlaps involve a backchannel ([Levinson & Torreira 2015](https://www.frontiersin.org/journals/psychology/articles/10.3389/fpsyg.2015.00731/full)).
- **Meetings of 3-9 people:** 11.6% of speaking time overlaps. Over 90% of overlapped time has a single other talker ([Çetin & Shriberg 2006](https://www.isca-archive.org/interspeech_2006/cetin06_interspeech.pdf)).
- **A two-person rehearsal on this rig:** 3.3% of speaking time overlapped, median 200 ms. The real two-voice sum peaked at -1.65 dBFS with no fader.
  - Adding time-shifted copies to make four voices: no fader gave a median of 0 clipped seconds per minute and a worst case of 0.74.
  - At -3 dB, none of the 12 trials clipped.

**Why -3 dB.** WebRTC's default AGC targets a -3 dBFS peak. Two such voices reach +3 dBFS where their peaks coincide, so -3 dB puts that at full scale and the clipper handles the rare remainder. Sizing the fader for N-1 simultaneous voices would cost every lone voice that much level, all the time, to cover a case that barely occurs.

**Stress test.** Five WHIP publishers all sending speech continuously, which is not a conversation. All five returns were captured over WHEP for 48 s, with a fresh backend for each arm:

| router settings | seconds containing a clipped sample |
| --- | --- |
| no fader, no clipper | 40-45 of 48 |
| soft clipper only | 4-18 of 48 |
| fader -6 dB | 0 of 48 |

-3 dB was not run on this rig.

## Risks a reviewer should weigh

- **The soft clipper alone is not enough.** It delivers -0.05 dBFS, and the Opus leg takes that to -0.00. EBU Tech 3343 keeps lossy codec input at or below -2 dBTP for the same reason: decoding overshoots. The fader is what creates that margin.
- **Why not `lsp-rs-limiter`.** It is statically linked, so present in every build even though `gst-inspect-1.0` cannot see it. Its envelope attack misses speech transients: on a four-seat mix 5 dB over, it passed 0.05% of samples above full scale at a -3 dB threshold. `rglimiter` passes 0.0000%.
- **Latency is zero** for both elements. An impulse comes out at the same sample, and the pipeline latency query is unchanged.
- **The fader is always built; the clipper only when enabled.** At unity `volume` refuses only unsigned formats, and every crosspoint is a `volume` already, so an unconfigured bus loses nothing. `rglimiter` is F32LE-only even when disabled.
- **Construction-time only**, like `latency` and `force_live`. A live per-output fader needs a fan-out like `routing_matrix`'s, and is left for a follow-up.

## Tests

Nine new tests in `backend/tests/liveaudiorouter_test.rs`, driven through the block's real builder output. Each guard was checked by reverting the piece it covers:

- **`an_engaged_output_bus_sums_in_float`** fails without the format pin.
- **`the_output_soft_clipper_holds_a_fan_in_overload_below_full_scale`** fails without the clipper.
- **`the_output_fader_attenuates_every_output_bus`** fails without the fader gain.
- **`a_fader_alone_keeps_a_fixed_point_bus_summing_in_float`** uses S16 sources and an S16 consumer. With the fader half of the pin removed, it reads -12 dBFS instead of about -6. The clipper test cannot catch this, because `rglimiter` forces float on its own.
- **`an_unconfigured_bus_still_feeds_an_integer_consumer_without_a_converter`** fails at the link if the bus is pinned to float or carries a clipper by default.

`rglimiter` ships in `gstreamer1.0-plugins-good`, which CI already installs, so nothing skips.

**What was run:**
- The full suite ran locally on macOS earlier in the branch: 791 passed, 7 ignored.
- On the latest commit: `liveaudiorouter_test` (28 passed), `openapi_test`, and `cargo clippy --all-targets -- -D warnings`. The router tests also passed three consecutive runs at `--test-threads=8` on the commit before it, which differs only in help text. The full suite was not re-run.
- Not run: CI on the latest commit. The rig and rehearsal measurements are macOS-only and not part of any suite.

### DIFF
--- a/backend/src/blocks/builtin/liveaudiorouter.rs
+++ b/backend/src/blocks/builtin/liveaudiorouter.rs
@@ -26,8 +26,22 @@
 //!            → mixer_O sink pad           the pad placing the mono
 //!                                         on output channel D
 //!
-//!   mixer_O → caps_out_O → capssetter_out_O → queue_out_O → audio_out_O
+//!   mixer_O → caps_out_O → capssetter_out_O → trim_out_O (volume)
+//!           → [limiter_out_O (rglimiter)] → queue_out_O → audio_out_O
 //! ```
+//!
+//! An output bus sums every crosspoint routed to it, so a mix-minus return
+//! carries N-1 talkers at unity and goes over full scale long before any one
+//! of them does. `trim_out_O` and `limiter_out_O` are what it has to catch
+//! that, in the order a console applies them.
+//!
+//! Both are off by default, and while they are off the bus is left alone down
+//! to its negotiation: `caps_out_O` pins `F32LE` only once one of them is
+//! engaged. That pin is the load-bearing part. An `audiomixer` saturates at
+//! the sum in a fixed-point format, so with the format left open whether a
+//! fan-in overload clipped inside the router or reached the output intact
+//! depended on what the next block negotiated — and neither stage can undo a
+//! sum that has already been saturated upstream of it.
 //!
 //! There is no queue on a crosspoint branch. A tee branch normally needs one
 //! so it cannot block its siblings, but every branch here ends on a
@@ -49,7 +63,7 @@ use gstreamer::prelude::*;
 use std::collections::HashMap;
 use strom_types::routing::{self, Crosspoint, RoutingGains};
 use strom_types::{block::*, element::ElementPadRef, PropertyValue, *};
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 /// Maximum number of input/output streams.
 const MAX_STREAMS: usize = 8;
@@ -110,6 +124,24 @@ const CROSSPOINT_PREFIX: &str = ":xp_";
 /// The `GstAudioConverter` option that maps a mixer sink pad's channels onto
 /// the mixer's output channels.
 const MIX_MATRIX_KEY: &str = "GstAudioConverter.mix-matrix";
+
+/// Name of the output trim property.
+pub const OUTPUT_HEADROOM_PROPERTY: &str = "output_headroom";
+
+/// Name of the output limiter property.
+pub const OUTPUT_LIMITER_PROPERTY: &str = "output_limiter_enabled";
+
+/// The element every output bus is limited by. `rglimiter` is a waveshaper:
+/// unity below -6 dBFS, a `tanh` knee above it, and an asymptote at full
+/// scale, so nothing it emits can reach 0 dBFS whatever goes in.
+///
+/// Not `lsp-rs-limiter`, which `builtin.mixer` uses through
+/// `make_limiter_element`. That one rides gain on an envelope, and on speech
+/// its attack loses the transient it was pointed at: measured on a four-seat
+/// return 5 dB over, it still let 0.05% of samples past full scale at a -3 dB
+/// threshold and 0.0007% at -10 dB, where it also costs 7.5 dB of level. A
+/// ceiling that is usually held is not a ceiling on a headphone feed.
+const LIMITER_FACTORY: &str = "rglimiter";
 
 /// Element naming for a crosspoint. The wire format itself lives in
 /// `strom_types::routing`, shared with the graph editor.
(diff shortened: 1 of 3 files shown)


srperens:

srperens (approved):
**Verdict: Approve** — the format pin is the load-bearing part and is correctly gated, both stages default to no-ops, and all nine new guards executed in CI rather than skipping. One small suggestion, not a blocker.

**Suggestion**

`output_fader_db` silently clamps an out-of-range request — ask for -40 dB and you get -24 dB with nothing in the log. `ExposedProperty` has no min/max field, so clamping is the only mechanism available and the range is documented in the property description; a one-line `warn!` when the clamp bites would close the gap without a type change. Worth weighing against #821, which moves the videoenc block toward rejecting out-of-range client properties rather than absorbing them.

**Claims**

|---|---|---|
| A router asking for neither stage negotiates exactly as before | `CONFIRMED` | `backend/src/blocks/builtin/liveaudiorouter.rs:394` — `bus_caps = bus_caps.field("format", "F32LE");` sits behind the `sum_in_float` gate computed at line 340, and `an_unconfigured_bus_still_feeds_an_integer_consumer_without_a_converter` passes in CI |
| The clipper element is absent, not merely disabled, when unconfigured | `CONFIRMED` | `liveaudiorouter.rs:426` — `if soft_clip_enabled {` guards the element push, and again at `:449` for the links; `rglimiter` is `F32LE`-only, so leaving a disabled one in would pin the bus |
| The fader cannot boost | `CONFIRMED` | `liveaudiorouter.rs:688` — `.clamp(routing::MIN_OUTPUT_FADER_DB, routing::MAX_OUTPUT_FADER_DB)` with `MAX_OUTPUT_FADER_DB = 0.0` |
| The nine new tests ran in CI | `CONFIRMED` | `Check (Linux)` at `e1c3efc` logs `liveaudiorouter_test` as `28 passed; 0 failed; 0 ignored` in 13.53s — no skip path, and `rglimiter` ships in `gstreamer1.0-plugins-good` which CI installs |
| `rglimiter` has no attack/release and cannot emit full scale | `EXTERNAL` | Upstream GStreamer behaviour, assumption stated; the body's measurements are rig data not reproducible here |

**Diagnosis** — The stated cause is right and the ordering argument holds: the pin has to happen at `caps_out_O`, upstream of both stages, because `audiomixer` saturates the sum in a fixed-point format and neither a fader nor a clipper can recover a sum already saturated upstream of it. Coverage is `BOUNDED` and the bound is stated — construction-time only, so a live per-output fader is a follow-up. Choosing `rglimiter` over the `lsp-rs-limiter` the mixer block uses is the non-obvious call, and the reasoning (envelope attack misses speech transients) is kept nowhere else in this repo, so recording it at `liveaudiorouter.rs:148` — `const SOFT_CLIPPER_FACTORY: &str = "rglimiter";` — is the right place for it. Failing the build when `rglimiter` is missing rather than substituting a passthrough is also correct: a silent passthrough would leave an operator believing the return is protected.

**Radius** — `SHARED`, and argued. One block file plus purely additive constants in `strom_types::routing`, no contract change. What makes it `SHARED` rather than `LOCAL` is that the fader is built unconditionally, so every existing `builtin.liveaudiorouter` instance gains a `volume` in each output path. At unity that is a passthrough and cannot clip, and the integer-consumer test above is what demonstrates negotiation is unaffected.

**Tests & CI** — All checks green at `e1c3efc`, and the workflow run's `head_sha` matches the current head: the PR body's "Not run: CI on the latest commit" was written before that run finished and is now stale in the author's favour. `Build (macOS)`/`Build (Windows)` report `skipping`; nothing here is platform-specific, so Linux-green covers it. The rig and rehearsal figures are macOS-only measurements outside any suite, as the body says — taken at face value, not verified here.

Confidence: HIGH

PR #796, opened 2026-09-11
feat(gst): report a branch whose link never formed

wagenet:
## The case

A flow can reach Playing, carry audio correctly, report every block healthy, and have no video at all, because two elements never linked. The scan in #786 cannot see it: it looks for a pad task that was paused, and where no link was made there is no task to pause.

A GPU-backend vision mixer with `output_format: NV12` and `gl_download: true` does this. `gldownload` moves GL memory to system memory without converting pixel format, so its RGBA output has no format in common with the NV12 capsfilter below it:

```
Could not link immediately: vmix:gldownload_dist:src -> vmix:capsfilter_dist:sink
  (error: Failed to link elements ... Pads do not have common format). Will retry when pad becomes available.
```

Both the distribution and multiview branches stay unlinked. The mixer's output tee has `allow-not-linked=true`, so the `not-linked` return never surfaces, no task parks, and nothing reaches the bus.

## What this reports

`pending_links` holds every link construction could not make, and the only thing that retries one is the `pad-added` handler on the link's source element. A pad that is already there will not be added again, so a deferred link whose named source pad exists and has no peer has missed its only retry: it is unformed for the life of the pipeline, not merely slow. The scan now checks that alongside the pad tasks and fails the block owning the source element.

A source pad that has not appeared yet is left alone: that is the case deferral exists for, and reporting it would fail every `decodebin` until it has typefound. Two shapes also read as formed and go unreported, both because the pad has a peer: one linked through the handler's pattern match, one that picked up an auto-tee.

## Failed rather than degraded

The branch carries nothing and will until the flow is rebuilt, so there is no reduced capability to describe. That is what `Failed` already means here: not passing data while the pipeline says Playing.

The check lives in the scan rather than behind a block-supplied hook because an unformed link is not private to a block: the `PipelineManager` already holds the state that proves it.

## Tests

New in `health.rs`: an unformed pair is reported against the block owning its source with the link named in the detail; a link still waiting on a dynamic pad is not; a link the handler resolved is not; and a flow through `PipelineManager` reaches Playing with only that one block failed. New in `block_health_test.rs`: an unlinkable pair under an `allow-not-linked` tee leaves the pipeline Playing with no bus error and no paused pad task anywhere. Both detector tests fail when `unformed_link` is stubbed to return false.

Ran locally: the full `cargo test` suite on the integration branch this was written on (677 lib tests plus every integration test, green), then `--lib gst::pipeline::health`, `--test block_health_test`, `--test openapi_test` and `--test pipeline_lifecycle_test` here.

## On a real flow

Five runs of a headless instance driving videotestsrc into a vision mixer, out to a WHEP output and through a video encoder to a recorder, with `multiview_out` left unwired. The repro above fires and names `vmix:gldownload_mv:src -> vmix:capsfilter_mv:sink`, with the distribution link deferred in the same startup log. The two healthy configurations, CPU backend and GPU backend with a working output format, report every block ok, the unwired `multiview_out` included. The scan also named a real unrelated defect: on the GPU backend with `gl_download: false`, `pgm_out` carries GL memory that the Video Encoder block's `videoconvert` cannot accept, so the recorder branch never links and records nothing. That is not addressed here.

### DIFF
--- a/backend/src/gst/pipeline/construction.rs
+++ b/backend/src/gst/pipeline/construction.rs
@@ -90,6 +90,8 @@ impl PipelineManager {
             cached_state: std::sync::Arc::new(std::sync::RwLock::new(PipelineState::Null)),
             qos_aggregator: QoSAggregator::new(),
             qos_broadcast_task: None,
+            block_health: std::sync::Arc::new(std::sync::RwLock::new(Vec::new())),
+            block_health_task: None,
             ptp_clock: None,
             ptp_stats: std::sync::Arc::new(std::sync::RwLock::new(None)),
             ntp_clock: None,
--- a/backend/src/gst/pipeline/health.rs
+++ b/backend/src/gst/pipeline/health.rs
@@ -49,14 +49,21 @@ fn first_stalled_pad(element: &gst::Element) -> Option<StalledPad> {
 
 /// Find the first paused pad task on `element` itself.
 ///
-/// Two kinds of legitimately paused task are excluded. An element held below
-/// `PLAYING` - a webrtcbin added for a newly connected WHEP consumer, say -
-/// pauses its tasks as part of that transition. And a pad that has seen EOS is
-/// finished by definition:
-/// `gst_base_src_loop` pauses its task on the way out for *every* reason
-/// including end of stream, and the element stays `PLAYING` afterwards, so a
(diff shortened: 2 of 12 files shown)


srperens:

PR #797, opened 2026-09-11
fix(vision_mixer): make output_format work on the GPU backend

wagenet:
## The bug

Setting the vision mixer's `output_format` on the GPU backend removed all video. The flow
reached `Playing`, reported every block healthy and carried audio.

`glvideomixerelement` is RGBA-only on both pad templates, and `gldownload` moves GL memory to
system memory without converting it. So the `output_format` capsfilter after the download never
linked:

```
Could not link immediately: vmix:gldownload_dist:src -> vmix:capsfilter_dist:sink
  (error: Failed to link elements ... Pads do not have common format)
```

On the GL-passthrough path (`gl_download=false`, the default) the property was ignored: the
GL-memory capsfilter carried no format.

## The fix

`output_convert_source` inserts a `glcolorconvert` between each output branch and its format
pin when `output_format` is set, and the GL-memory caps now carry the format. Converting RGBA to
NV12 on the GPU also cuts the buffer that crosses the bus to 37.5%. The converter hangs off
`tee_pgm`'s output branch, so the PGM feed into the multiview compositor stays RGBA.

Keyed alpha needs no handling here, unlike the CPU path: `compositor` blends in its output
format, but `glvideomixerelement` accepts only RGBA, so no format pin can reach its blend.
`keyed_dsk_alpha_survives_nv12_on_gpu` measures this.

## Tests

`backend/tests/vision_mixer_gl_output_format_test.rs`, five tests through `PipelineManager`:

- `gpu_output_format_links_with_gl_download` / `..._without_gl_download` walk each output branch
  back to its origin and assert every hop is linked, the `glcolorconvert` is on it, the format
  pin carries `output_format`, and the multiview PGM tile bypasses the conversion. They need the
  GL plugin but no GL context, fail rather than skip without it, and run in CI. This PR adds
  `gstreamer1.0-gl` to the CI package lists; on Ubuntu it arrived only as a dependency.
- `gpu_output_format_is_negotiated_with_gl_download` / `..._without_gl_download` assert NV12 is
  negotiated on both branches, in system or GL memory respectively, and that buffers reach the
  output.
- `keyed_dsk_alpha_survives_nv12_on_gpu` keys in a half-transparent DSK and asserts the
  transparent half shows the background.

The last three need a GL context and **skip in Linux CI**, as `vision_mixer_fx_test` does.

With the builder changes reverted, all five fail locally, each on its own assertion (link
missing, converter missing, 0 buffers, RGBA negotiated, keyed-alpha timeout).

Ran locally (Apple M4 Max, GStreamer 1.28.6), all passing:

- at the head: the new test file, with and without the fix, and the pre-commit hook.
- at `72ce792`, before the follow-ups that change only comments in `backend/src`:
  `vision_mixer_fx_test`, `shader_validation_test`, `pipeline_lifecycle_test`,
  `vision_mixer_failed_start_test`, `openapi_test`, clippy `-D warnings`, `cargo fmt --check`.

Not run: the full suite.

## Rig verification

GPU backend, `output_format: NV12`, `gl_download: true`, two inputs, one DSK, program and
multiview each into a WHEP Output block.

| | before | after |
|---|---|---|
| `Could not link immediately` | 2 | 0 |
| `Detected raw video input` in WHEP blocks | 0 | 2 |
| NV12 in the live `debug-graph` | requested only | 15 negotiated, 2 in GL memory |

A JPEG tap on `pgm_out` wrote 1444 frames, mean luma 185, so real picture. No browser consumer
was completed (`whepclientsrc` failed ICE from this host), so the tap is the frame proof.

## Related

#796 detects this failure class: a declared link that never formed in a `Playing` flow.

#799 makes the WHEP Output block convert its own input; with NV12 from here it does nothing. An
RGBA feed costs ~24-39% of a core per extra H.265 viewer against ~1.4% with NV12.

### DIFF
--- a/backend/src/blocks/builtin/vision_mixer/builder/mod.rs
+++ b/backend/src/blocks/builtin/vision_mixer/builder/mod.rs
@@ -278,20 +278,27 @@ impl<'a> PipelineParams<'a> {
     /// Build PGM output caps for the GL-memory passthrough path.
     /// Constrains framerate/resolution without forcing a download to system memory.
     pub(super) fn pgm_caps_glmem(&self) -> gst::Caps {
-        let s = format!(
-            "video/x-raw(memory:GLMemory),width={},height={},framerate={}/{},pixel-aspect-ratio=1/1",
-            self.pgm_w, self.pgm_h, self.pgm_framerate.0, self.pgm_framerate.1
-        );
-        s.parse().expect("valid GL memory caps for PGM")
+        self.raw_caps_glmem(self.pgm_w, self.pgm_h, self.pgm_framerate)
+    }
+
+    /// GL-memory caps: constrains framerate/resolution, and `output_format`
+    /// when set, without forcing a download to system memory.
+    fn raw_caps_glmem(&self, w: u32, h: u32, framerate: (i32, i32)) -> gst::Caps {
+        let mut builder = gst::Caps::builder("video/x-raw")
+            .features(["memory:GLMemory"])
+            .field("width", w as i32)
+            .field("height", h as i32)
+            .field("framerate", gst::Fraction::new(framerate.0, framerate.1))
+            .field("pixel-aspect-ratio", gst::Fraction::new(1, 1));
+        if let Some(fmt) = self.output_format.as_deref() {
+            builder = builder.field("format", fmt);
+        }
+        builder.build()
     }
 
     /// Build multiview output caps for the GL-memory passthrough path.
     pub(super) fn mv_caps_glmem(&self) -> gst::Caps {
-        let s = format!(
-            "video/x-raw(memory:GLMemory),width={},height={},framerate={}/{},pixel-aspect-ratio=1/1",
-            self.mv_w, self.mv_h, self.mv_framerate.0, self.mv_framerate.1
-        );
-        s.parse().expect("valid GL memory caps for MV")
+        self.raw_caps_glmem(self.mv_w, self.mv_h, self.mv_framerate)
     }
 }
 
--- a/backend/src/blocks/builtin/vision_mixer/builder/pipeline_gpu.rs
+++ b/backend/src/blocks/builtin/vision_mixer/builder/pipeline_gpu.rs
@@ -10,6 +10,45 @@ use super::super::{elements, layout};
 use super::{audio_meter, pad_layout, setup_overlay_renderer, PipelineParams};
 use crate::blocks::{BlockBuildContext, BlockBuildError, BlockBuildResult};
 
+/// Insert a `glcolorconvert` ahead of an output branch's format pin when
+/// `output_format` is set, and return the element/pad the branch links from.
+///
+/// `gldownload` moves GL memory to system memory and leaves the pixel format
+/// alone, and `glvideomixerelement` is RGBA-only, so a format pin with nothing
+/// converting in front of it has no common format with its peer and never
+/// links — the branch then carries no video at all while the flow still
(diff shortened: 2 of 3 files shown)


srperens:

srperens (approved):
**Verdict: Approve** — right layer, no-op when the property is unset, and the two tests that guard the actual failure genuinely execute in this repo's CI. One downstream consequence is worth naming before merge; it does not block.

**Worth naming**

Making `output_format` take effect on the `gl_download=false` path changes what downstream blocks receive: previously RGBA in GL memory (the property was ignored), now `output_format` in GL memory. CLAUDE.md warns that `whepserversink` "accepts `video/x-raw(memory:GLMemory)` and then fails encoder discovery" — so a flow pairing `output_format` with `gl_download=false` into a WHEP Output is the configuration to watch. On the `gl_download=true` path nobody can be relying on the old behaviour — that is the bug — but the passthrough path silently changes format for anyone who had already set the property.

**Claims**

|---|---|---|
| The conversion is skipped entirely when `output_format` is unset | `CONFIRMED` | `backend/src/blocks/builtin/vision_mixer/builder/pipeline_gpu.rs:33` — `let Some(format) = p.output_format.as_deref() else {` returns the caller's own element/pad unchanged |
| The GL-memory capsfilter now carries the format | `CONFIRMED` | `backend/src/blocks/builtin/vision_mixer/builder/mod.rs:293` — `builder = builder.field("format", fmt);` |
| The link-walking tests run in CI rather than skipping | `CONFIRMED` | `vision_mixer_gl_output_format_test.rs:48` — `gstreamer::ElementFactory::find("glvideomixerelement").is_some(),` asserts; `Check (Linux)` at `d0517ba` logs both `gpu_output_format_links_*` as `ok` |
| The three end-to-end tests pass green without testing anything in CI | `CONFIRMED` | `vision_mixer_gl_output_format_test.rs:294` — `if !gl_environment_available() {` — the binary reports `5 passed` in 0.11s, which is the shape of an early return, and the PR body says so |

**Diagnosis** — The stated cause explains the symptom exactly: `gldownload` moves memory without converting, `glvideomixerelement` is RGBA-only, so a format pin with no converter in front has no common format and the branch never links while the flow still reports `Playing`. Coverage is `ABSOLUTE` for the GPU backend's two output branches — both were changed, and both are tested. Converting on the GPU before the download is also the right side of the bus.

This does not conflict with the CLAUDE.md rule that a producer must not adapt for its consumer. `output_format` is this block's own declared output contract, not a guess about what is downstream, and the `gl_download=false` path keeps the memory in GL rather than forcing a round trip.

**Radius** — `SHARED`, and argued: the new element is only constructed when `output_format` is set, so an unset flow builds the identical graph it did before. What is shared is the negotiated format reaching every downstream consumer of both output branches, which is the point above. `raw_caps_glmem` also replaces two `.parse().expect(...)` paths with the caps builder, removing two panic sites.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` green at `d0517ba`. I read the `Check (Linux)` log: `vision_mixer_gl_output_format_test` ran all five tests, and the two that walk the link graph are the ones that fail if the builder change is reverted. `Build (macOS)`/`Build (Windows)` report `skipping`; there is no platform-specific code here, so Linux-green covers it.

Adding `gstreamer1.0-gl` to all three package lists is a real improvement beyond this PR — it is the missing package that also leaves #755's `alpha_survives_gl_videomixer` unable to run.

Confidence: HIGH

Issue #741, opened 2026-09-01
Published Docker images report git_hash "unknown" — no way to map an image back to a commit


srperens:
## Description

Published Docker images carry no git provenance. `--version-info` and `GET /api/version` report `git_hash: "unknown"`, `git_tag: ""`, `git_branch: "unknown"` and `git_dirty: false` for every release image, so there is no way to map a running container back to the commit it was built from.

Local and CI builds are unaffected — they report the hash correctly.

## Steps to Reproduce

```
$ docker run --rm -p 8080:8080 eyevinntechnology/strom:0.6.8
$ curl -s localhost:8080/api/version
```

## Expected Behavior

```json
{"version":"0.6.8","git_hash":"13722d5a","git_tag":"v0.6.8","git_branch":"main", ...}
```

## Actual Behavior

```json
{"version":"0.6.8","git_hash":"unknown","git_tag":"","git_branch":"unknown","git_dirty":false,
 "build_timestamp":"2026-09-01T13:46:57.370065188+00:00",
 "build_id":"38a1399a-f773-4da9-bb38-e7ef99eecafc",
 "gstreamer_version":"1.26.6","os_info":"Ubuntu 25.10","in_docker":true}
```

`build_id` and `build_timestamp` are set, so a build can be identified — but not its source.

## Cause

`backend/build.rs` gets all four values by shelling out to `git` inside the build (`set_version_info`, lines 63-115), and `.dockerignore` excludes `.git/`. Inside the build container there is no repository, every `git` invocation fails, and each falls through to its default:

| field | `build.rs` fallback | result in image |
|---|---|---|
| `git_hash` | `unwrap_or_else(\|\| "unknown")` | `"unknown"` |
| `git_tag` | `unwrap_or_default()` | `""` |
| `git_branch` | `unwrap_or_else(\|\| "unknown")` | `"unknown"` |
| `git_dirty` | `unwrap_or(false)` | always `false`, even if it were dirty |

The `actions/checkout` step in `docker-publish.yml` is fine — the runner has the full repo. The metadata is lost at the build-context boundary, not at checkout.

Excluding `.git/` from the context is the right call on its own: it is large and changes on every commit, so including it would invalidate the Docker layer cache constantly. The fix is to pass the values in, not to ship the repository.

## Suggested Fix

Pass the values as build args and let `build.rs` prefer them over shelling out to `git`:

[...]

PR #800, opened 2026-09-11
fix(build): accept git provenance as Docker build args

srperens:
Class C — blocked on a dispatch: `gh workflow run docker-publish.yml --ref agent/fix-741-docker-git-provenance`.

Refs #741

**This diff is half the fix and is inert on its own.** The `docker-publish.yml` half that supplies the build args could not be pushed — this token has no `workflow` scope. The patch is in a comment below; until it lands, published images still report `"unknown"`.

Implements **Option A** (`/agent-fix A`, 2026-09-09); the triage marker's `work=bug radius=LOCAL excluded=none` describes that option, which I built unchanged.

## Problem

`.dockerignore:21` keeps the repository out of the build context, rightly: it is large and changes every commit. `build.rs` got all four values by shelling out to `git`, so inside the image build every call fails into its default. `backend/src/version.rs:50` — `        git_hash: env!("GIT_HASH").to_string(),` — embeds the result, so `/api/version` and `--version-info` report `"unknown"` for every published image.

## Change

`build.rs` prefers values from the environment — `backend/build.rs:74` — `    let git_hash = version_field_from_env("GIT_HASH")` — falling back to `git` when blank, because an `ARG` the caller never passed arrives as an empty string rather than unset: the trap PR #738 hit. The `backend-builder` stage declares them — `Dockerfile:196` — `ARG GIT_HASH` — immediately before the build, so no earlier layer is invalidated per commit.

The three near-identical `git` chains collapse into one `git_value()` helper.

Option B — OCI labels only — was rejected in triage: the binary, and therefore `/api/version`, would still say `"unknown"`.

## Evidence

**No Rust test can guard the wiring, and I did not write one that pretends to.** Build scripts are not test targets; `backend/tests/build_version_test.rs` includes `backend/build_version.rs` — the real file `build.rs` includes, not a copy — covering blank, whitespace, unset and trimmed values. It does not prove `build.rs` consults the environment: deleting the `.or_else` wiring leaves it green. The wiring's only guard is the `Verify git provenance` step in the patch below: it runs the pushed image and asserts `/api/version` reports its commit. `docker-publish.yml` runs only on a tag or a dispatch, so it can never run on a PR — hence class C.

Commit 1 **fails to compile**: `build_version.rs` does not exist yet. That is the intended red, not a broken PR.

- `c1b5ce2` https://github.com/Eyevinn/strom/actions/runs/34587991127 — test alone, red on `Check (Linux)`: `couldn't read backend/tests/../build_version.rs`
- `457f754` https://github.com/Eyevinn/strom/actions/runs/34588068641 — with the fix, green

With no `cargo` or Docker here, the one thing I ran is the patch's new `Extract version` shell against this repository: a tag push yields `git_tag=v0.6.8`, `git_branch=main`, `git_hash=c1b5ce2d`; a dispatch on a branch yields an empty tag and the branch name  — as the issue asks.

## Not verified

Untested here: the end-to-end round trip, the ARM64 cross-compile path, cold-container start-up timing, and `git_dirty`.

## Blast radius

LOCAL. Both new helpers are called only from `set_version_info`, which `main()` calls once — `backend/build.rs:12` — `    set_version_info();`. Local and CI builds set no `GIT_*` variable, so the `git` fallback runs exactly as before. `/api/version` changes values, never its shape.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -190,6 +190,16 @@ COPY --from=frontend-builder /app/backend/dist backend/dist
 # Build the backend (headless - no native GUI needed in Docker) and MCP server
 ENV RUST_BACKTRACE=1
 
+# Git provenance for /api/version and --version-info. The build context has no .git/
+# (see .dockerignore), so build.rs cannot shell out to git here and takes these instead.
+# Declared immediately before the build so an earlier layer is not invalidated per commit.
+ARG GIT_HASH
+ARG GIT_TAG
+ARG GIT_BRANCH
+ENV GIT_HASH=${GIT_HASH} \
+    GIT_TAG=${GIT_TAG} \
+    GIT_BRANCH=${GIT_BRANCH}
+
 # Cross-compilation: Use cargo-zigbuild with glibc 2.36 targeting (Raspberry Pi compatible)
 # Native compilation: Use regular cargo build
 # sccache: see the frontend stage — probe the backend and only wrap rustc when
--- a/backend/build.rs
+++ b/backend/build.rs
@@ -5,6 +5,8 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use uuid::Uuid;
 
+include!("build_version.rs");
+
 fn main() {
     // Set version and build information
     set_version_info();
@@ -35,6 +37,12 @@ fn main() {
     // Rerun if .git/HEAD changes (new commits)
     println!("cargo:rerun-if-changed=../.git/HEAD");
     println!("cargo:rerun-if-changed=../.git/refs");
+    // Rerun if the included version-field helper changes
+    println!("cargo:rerun-if-changed=build_version.rs");
+    // Rerun if the Docker build args carrying git provenance change
+    println!("cargo:rerun-if-env-changed=GIT_HASH");
+    println!("cargo:rerun-if-env-changed=GIT_TAG");
+    println!("cargo:rerun-if-env-changed=GIT_BRANCH");
     // Rerun if Windows icon changes
     println!("cargo:rerun-if-changed=strom.ico");
 
@@ -60,53 +68,23 @@ fn main() {
 
 /// Set version and build information as environment variables
 fn set_version_info() {
-    // Get git commit hash (short)
-    let git_hash = Command::new("git")
-        .args(["rev-parse", "--short=8", "HEAD"])
-        .output()
-        .ok()
-        .and_then(|output| {
-            if output.status.success() {
-                String::from_utf8(output.stdout).ok()
-            } else {
-                None
-            }
-        })
-        .map(|s| s.trim().to_string())
+    // Docker builds carry no repository: `.dockerignore` excludes `.git/` so the layer cache
+    // is not invalidated on every commit. The publish workflow passes these in as build args
+    // instead, and they take precedence over the shell-out that cannot succeed there.
+    let git_hash = version_field_from_env("GIT_HASH")
+        .or_else(|| git_value(&["rev-parse", "--short=8", "HEAD"]))
         .unwrap_or_else(|| "unknown".to_string());
 
-    // Get git tag (if on a tagged commit)
-    let git_tag = Command::new("git")
-        .args(["describe", "--tags", "--exact-match"])
-        .output()
-        .ok()
-        .and_then(|output| {
-            if output.status.success() {
-                String::from_utf8(output.stdout).ok()
-            } else {
-                None
-            }
-        })
-        .map(|s| s.trim().to_string())
+    let git_tag = version_field_from_env("GIT_TAG")
+        .or_else(|| git_value(&["describe", "--tags", "--exact-match"]))
         .unwrap_or_default();
+
+    let git_branch = version_field_from_env("GIT_BRANCH")
+        .or_else(|| git_value(&["rev-parse", "--abbrev-ref", "HEAD"]))
+        .unwrap_or_else(|| "unknown".to_string());
 
     // Get build timestamp (ISO 8601 format)
     let build_timestamp = chrono::Utc::now().to_rfc3339();
-
-    // Get git branch
-    let git_branch = Command::new("git")
-        .args(["rev-parse", "--abbrev-ref", "HEAD"])
-        .output()
-        .ok()
-        .and_then(|output| {
-            if output.status.success() {
-                String::from_utf8(output.stdout).ok()
-            } else {
-                None
-            }
-        })
-        .map(|s| s.trim().to_string())
-        .unwrap_or_else(|| "unknown".to_string());
 
     // Check if working directory is dirty
     let git_dirty = Command::new("git")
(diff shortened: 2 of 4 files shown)


srperens:
The `docker-publish.yml` half of #741, which I could not push (`workflow` scope). Apply with `git apply` on this branch, or paste the steps in by hand.

It does three things: computes `git_hash`/`git_tag`/`git_branch` in the existing `Extract version` step of both jobs, passes them as `build-args` to both `docker/build-push-action` steps, and adds a `Verify git provenance` step that runs the pushed image and asserts `/api/version` reports the commit. That last step is the only guard the fix has — and failing it also stops `create-manifest`, so a provenance-less build never becomes the published tag.

```diff
diff --git a/.github/workflows/docker-publish.yml b/.github/workflows/docker-publish.yml
index 4906347..039ef82 100644
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -50,12 +50,20 @@ jobs:
 
       - name: Extract version
         id: version
+        env:
+          BASE_REF: ${{ github.event.base_ref }}
         run: |
           if [[ "$GITHUB_REF" == refs/tags/* ]]; then
             echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+            echo "git_tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+            # A tag push is a detached checkout; base_ref names the branch it was cut from.
+            echo "git_branch=${BASE_REF#refs/heads/}" >> $GITHUB_OUTPUT
           else
             echo "version=dev-$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+            echo "git_tag=" >> $GITHUB_OUTPUT
+            echo "git_branch=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_OUTPUT
           fi
+          echo "git_hash=$(git rev-parse --short=8 HEAD)" >> $GITHUB_OUTPUT
 
       - name: Build and push AMD64 image
         uses: docker/build-push-action@v7
@@ -66,6 +74,11 @@ jobs:
           push: true
           tags: ${{ env.IMAGE_NAME }}:${{ steps.version.outputs.version }}-amd64
           platforms: linux/amd64
+          # The build context excludes .git/, so provenance has to be passed in.
+          build-args: |
+            GIT_HASH=${{ steps.version.outputs.git_hash }}
+            GIT_TAG=${{ steps.version.outputs.git_tag }}
+            GIT_BRANCH=${{ steps.version.outputs.git_branch }}
           cache-from: type=gha,scope=docker-amd64
           # mode=min: cache only the final image layers. mode=max also exported
           # the builder's multi-GB target/ each build (~6 min), which COPY . .
@@ -79,6 +92,28 @@ jobs:
             aws_access_key_id=${{ secrets.SCCACHE_S3_ACCESS_KEY_ID }}
             aws_secret_access_key=${{ secrets.SCCACHE_S3_SECRET_ACCESS_KEY }}
 
+      # The guard for #741: build.rs is not reachable from `cargo test`, so the only place
+      # the build-arg plumbing can be checked is against a running image. Failing here also
+      # stops create-manifest, so a provenance-less build never becomes the published tag.
+      - name: Verify git provenance in the AMD64 image
+        env:
+          IMAGE: ${{ env.IMAGE_NAME }}:${{ steps.version.outputs.version }}-amd64
+          EXPECTED_HASH: ${{ steps.version.outputs.git_hash }}
+        run: |
+          docker run --rm -d --name strom-provenance -p 8080:8080 "$IMAGE"
+          reported=""
+          for _ in $(seq 1 30); do
+            reported=$(curl -fsS localhost:8080/api/version 2>/dev/null | jq -r '.git_hash // empty') || true
+            [ -n "$reported" ] && break
+            sleep 2
+          done
+          docker rm -f strom-provenance >/dev/null 2>&1 || true
+          echo "==> /api/version reports git_hash='$reported', expected '$EXPECTED_HASH'"
+          if [ "$reported" != "$EXPECTED_HASH" ]; then
+            echo "The image cannot be mapped back to a commit (#741)."
+            exit 1
+          fi
+
   # Build ARM64 image natively on ARM64 runner (no cross-compilation!)
   build-arm64:
     name: Build ARM64
@@ -104,12 +139,20 @@ jobs:
 
       - name: Extract version
         id: version
+        env:
+          BASE_REF: ${{ github.event.base_ref }}
         run: |
           if [[ "$GITHUB_REF" == refs/tags/* ]]; then
             echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+            echo "git_tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+            # A tag push is a detached checkout; base_ref names the branch it was cut from.
+            echo "git_branch=${BASE_REF#refs/heads/}" >> $GITHUB_OUTPUT
           else
             echo "version=dev-$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+            echo "git_tag=" >> $GITHUB_OUTPUT
+            echo "git_branch=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_OUTPUT
           fi
+          echo "git_hash=$(git rev-parse --short=8 HEAD)" >> $GITHUB_OUTPUT
 
       - name: Build and push ARM64 image
         uses: docker/build-push-action@v7
@@ -120,6 +163,11 @@ jobs:
           push: true
           tags: ${{ env.IMAGE_NAME }}:${{ steps.version.outputs.version }}-arm64
           platforms: linux/arm64
+          # The build context excludes .git/, so provenance has to be passed in.
+          build-args: |
+            GIT_HASH=${{ steps.version.outputs.git_hash }}
+            GIT_TAG=${{ steps.version.outputs.git_tag }}
+            GIT_BRANCH=${{ steps.version.outputs.git_branch }}
           cache-from: type=gha,scope=docker-arm64
           # mode=min: see the AMD64 job above.
           cache-to: type=gha,mode=min,scope=docker-arm64
@@ -131,6 +179,26 @@ jobs:
             aws_access_key_id=${{ secrets.SCCACHE_S3_ACCESS_KEY_ID }}
             aws_secret_access_key=${{ secrets.SCCACHE_S3_SECRET_ACCESS_KEY }}
 
+      # See the AMD64 job above.
+      - name: Verify git provenance in the ARM64 image
+        env:
+          IMAGE: ${{ env.IMAGE_NAME }}:${{ steps.version.outputs.version }}-arm64
+          EXPECTED_HASH: ${{ steps.version.outputs.git_hash }}
+        run: |
+          docker run --rm -d --name strom-provenance -p 8080:8080 "$IMAGE"
+          reported=""
+          for _ in $(seq 1 30); do
+            reported=$(curl -fsS localhost:8080/api/version 2>/dev/null | jq -r '.git_hash // empty') || true
+            [ -n "$reported" ] && break
+            sleep 2
+          done
+          docker rm -f strom-provenance >/dev/null 2>&1 || true
+          echo "==> /api/version reports git_hash='$reported', expected '$EXPECTED_HASH'"
+          if [ "$reported" != "$EXPECTED_HASH" ]; then
+            echo "The image cannot be mapped back to a commit (#741)."
+            exit 1
+          fi
+
   # Create multi-arch manifest combining AMD64 and ARM64 images
   create-manifest:
     name: Create Multi-arch Manifest
```

PR #802, opened 2026-09-11
fix(gst): download GL memory where a consumer cannot take it

wagenet:
## The defect

A Vision Mixer on the GPU backend with `gl_download` off emits `video/x-raw(memory:GLMemory)` on `pgm_out`. Linking that to a Video Encoder is refused at construction with `Pads do not have common format`: the encoder's `videoconvert` answers a caps query by asking the encoder behind it, so it advertises system memory only. Nothing retries that link, so the encoder and everything below it carry no data while the flow reports Playing.

## The fix

When a plain link is refused with `Noformat`, the producer offers GL memory and nothing else, and the consumer takes raw video in system memory, the linker links them through a `gldownload`. A link that leaves a pad unnamed is refused element to element, with no reason given; the linker then pairs the named or unlinked pads on each side, links the first pair a `gldownload` would join to get the concrete refusal, and adapts that. If either half of that splice is refused, the `gldownload` is removed and the original error is reported. A path that links on its own never reaches this code, and CUDA, NVMM, D3D11 and VA producers are untouched.

Every link made while the flow is built is adapted, whichever pads it names, with two exceptions: sink pads that an aggregator (`mpegtsmux`, `glvideomixerelement`) creates on request, since neither takes system-memory raw video from a GL producer, and, for an element-level link, request pads the consumer has not created yet. A pad linked at runtime by the `pad-added` handler is not adapted.

Fixing it in the block does not work. `queue`, `identity` and `capsfilter` fronts proxy the caps query, and all three still refused the GL producer when measured. An unconditional `gldownload` would refuse the CUDA and NVMM producers that link today.

## Which consumers shared it

Each raw-video consuming block fed a GPU Vision Mixer on a real rig, `block_health` read after 12 s.

| block | before | after |
|---|---|---|
| `builtin.videoenc`, `builtin.videoformat` | link never formed | fixed, records video |
| `builtin.recorder`, `builtin.thumbnail` | healthy | unaffected |
| `builtin.compositor`, `builtin.mpegtssrt_output` | fails identically on the CPU backend | not a memory-format defect |
| `builtin.ndi_output`, `builtin.tams_output`, `builtin.decklink_output` | not exercised | no runtime, gateway or hardware |

Flow `videotestsrc -> vision_mixer (gpu, gl_download false) -> videoenc -> recorder`: before, `vmix` and `venc` were `failed` with no recording; after, every block is `ok` and the recording decodes as 44 s of 1920x1080 H.264 test pattern. On the CPU backend, or with `gl_download` true, no `gldownload` is inserted.

## Tests

`gl_memory_link_test.rs` builds pipelines through `PipelineManager` and inspects the topology. Linking happens in NULL, so it needs the GL plugin (`gstreamer1.0-gl`, now listed in all three CI package lists) but no GL context, and it fails rather than skips when the plugin is missing. Two tests call `gl_link` directly: a `WasLinked` refusal comes back unadapted, and a failed splice leaves nothing behind.

`a_gl_only_producer_is_adapted_when_a_pad_is_unnamed` links the producer as `glsrc -> convert:sink`, `glsrc:src -> convert` and `glsrc -> convert`.

Reverting each fix on macOS turns its test red: the pad-level linker hook fails two, each of the three element-level hooks fails the unnamed-pad test on its own spelling, and the `Noformat` check, the cleanup and the system-memory consumer check fail one each. `a_system_memory_producer_is_left_alone` passes either way; it guards against a needless download.

**Ran on macOS**: the rig on the first commit. On the head, rebased onto current `main`: full suite (835 passed, 2 ignored, no skips), fmt and clippy. The later commits change only failure paths and unnamed-pad links, neither of which the rig flow takes. **Linux CI**: the head is pending. **Not run**: CUDA hosts, where `autovideoconvert` takes GL memory directly, so the block-level test there checks only that the link stands.


srperens:

srperens (commented):
**Verdict: Comment** — right diagnosis, right layer, and the tests are real guards that demonstrably executed. Two things to settle: the adaptation covers one of three build-time link shapes, and the body's coverage sentence does not say so.

**Requested changes**

1. Add `gstreamer1.0-gl` to all three apt package lists in `.github/workflows/ci.yml` (first at `.github/workflows/ci.yml:104-107`). `gldownload` resolved on the runner — the tests ran — but as an undeclared transitive dependency, and CLAUDE.md's Tests rule asks for the package naming the required element to be added in the same PR.

2. Correct the coverage sentence in the body. "Only links made while the flow is built are adapted" reads as *every* build-time link. In fact only the branch where **both** ends name a pad is adapted — `backend/src/gst/pipeline/linking.rs:274` — `            if let Err(e) = src_pad_obj.link(&sink_pad_obj) {`. Two other build-time branches keep the old behaviour: `backend/src/gst/pipeline/linking.rs:304` — `            src.link_pads(None::<&str>, sink, Some(sink_pad_name))` — and `backend/src/gst/pipeline/linking.rs:324` — `            src.link(sink).map_err(|e| {`. A flow that links a GL producer into a system-memory consumer without naming the source pad still fails exactly as reported. Either extend the adaptation to them or say in the body why they cannot carry a GL producer.

**Claims**

|---|---|---|
| The adaptation fires only on a format refusal, so a working flow pays nothing | `CONFIRMED` | `backend/src/gst/gl_link.rs:97` — `    if refusal != gst::PadLinkError::Noformat || !needs_gl_download_to_link(src, sink) {` — and the negative control `backend/tests/gl_memory_link_test.rs:166` — `        peer_factory(manager.pipeline(), "syssrc").as_deref(),` |
| CUDA, NVMM, D3D11 and VA producers are untouched | `CONFIRMED` | `offers_only_gl_memory` requires the GL feature specifically; `other_gpu_memory_types_are_left_alone` and `a_consumer_that_takes_only_gpu_memory_cannot` cover all five features |
| The tests execute in CI rather than skipping | `CONFIRMED` | job 104505878221: `running 5 tests` … `5 passed; 0 failed; 0 ignored`. The guard hard-asserts — `backend/tests/gl_memory_link_test.rs:31` — `            gstreamer::ElementFactory::find(factory).is_some(),` |
| Coverage is complete for build-time links | `CONTRADICTED` | the two branches cited in requested change 2 are unadapted |
| A CUDA host's `autovideoconvert` takes GL memory unaided, so no download is inserted there | `EXTERNAL` | no CUDA runner in this repo; assumed from the author's rig measurement |

**Diagnosis** — The mechanism holds: a `videoconvert` with an encoder behind it proxies the caps query and answers system memory, so the refusal lands at link time, before any caps event a probe could act on — which is why `gl_bridge`, the existing caps-driven adapter, cannot reach this case. Coverage is `BOUNDED` to both-ends-named links.

**Radius** — `SHARED`. The code sits in the central linker every flow's construction passes through, but is gated behind a refusal that previously aborted the build outright. No link that succeeds today can reach it, so no working flow changes shape or pays a GPU round trip; the only change on the old path is a longer error string.

**Lifetime** — no closure, probe, timer or thread is created, so the `WeakRef` rule does not bite. The inserted element is owned by the bin — `backend/src/gst/gl_link.rs:113` — `    bin.add(&gldownload)` — and dies with it; the failure path removes it first.

**Tests & CI** — Linux, WASM and API Contract green at `1d36653`. macOS and Windows `skipping`; the diff is platform-independent Rust, so the `main` build covers them.

Confidence: HIGH

wagenet:
Claude here, on behalf of @wagenet. Both requested changes are addressed in fa21e8d, and the branch is rebased onto current `main`.

1. `gstreamer1.0-gl` is now in all three apt package lists in `ci.yml`.
2. The element-level branches are adapted too, not just documented. There were three, not two: `linking.rs` also has a source-named, sink-unnamed branch for non-aggregators. When GStreamer refuses one of these links it gives no reason, so `retry_element_link_with_gl_download` pairs the named or unlinked pads, links the first pair a `gldownload` would join to get the concrete `PadLinkError`, and sends that through the existing pad-level path. What it does not cover: request pads the consumer has not created yet, and the aggregator auto-request branches (`mpegtsmux` and `glvideomixerelement` don't take system-memory raw video from a GL producer). The body's coverage sentence now says this.

New test: `a_gl_only_producer_is_adapted_when_a_pad_is_unnamed` builds through `PipelineManager` with all three spellings. I reverted each hook on its own, and each revert fails the test on its own spelling. Full suite on macOS: 835 passed, 2 ignored. Linux CI is pending.

PR #803, opened 2026-09-11
feat(vision mixer): report take morphs and per-input media age

wagenet:
## What

Two vision mixer changes from exercising the producer switching workflow on a five-seat
meeting rig:

1. `actual_transition_type` reports `morph` when a take animated a shared source instead
   of dissolving.
2. Vision mixer state carries `input_media_age_ms`, so a frozen input can be told apart
   from a motionless one.

**Stacked on #798**, which adds the guide this amends. A cross-fork PR cannot base on a
fork-only branch, so #798's commit shows here too and the diff narrows once it merges.
Review the last four commits.

## Why

**The take reported a move as a dissolve.** The PiP-aware path animates pads, not
pictures: a source in both compositions moves into its new box, and only sources
exclusive to one side cross-fade. Taking a four-up to that participant full frame reads
as a zoom. That is deliberate, but the path reported `fade` for every non-cut take, so a
caller asking for a dissolve was told it got one. Behaviour is unchanged; only the report
is accurate. On the rig: four-up to a member of it now reports `morph`; disjoint PiPs and
input-to-input still report `fade`; the existing `slide_*` downgrade still reports `fade`.

**A dropped participant stays on air as a still frame.** The compositor repeats an
input's last frame indefinitely, so a lost connection looks exactly like someone sitting
motionless and persists after the ingest session is reaped. `input_media_age_ms` gives
milliseconds since each input last delivered a frame:

```jsonc
"input_media_age_ms": [ 24, 24, 7727, 24, null ]
//                       ^live ^live  ^^^^ frozen 7.7s   ^never published
```

Stamping at the mixer's own sink pads covers every input type, not only WHIP seats, and
measures what actually decides the picture. Killing one publisher produced 1.8s → 29.9s
on that input while three live neighbours stayed under 30 ms, and the age returned to
25 ms on rejoin.

## Risk the reviewer should weigh

These are BUFFER probes, the hottest path in the pipeline. The callback reads the
monotonic clock and stores one relaxed atomic: no lock, no allocation, no formatting, no
state lookup. The overlay state is captured as an `Arc`, not looked up per buffer, and is
not a GStreamer object, so the closure forms no reference cycle with the pipeline. Five
inputs at 30 fps is 150 stores per second. I did not A/B measure the cost.

## Tests

Added: three covering the morph classification, one covering the age accessors, and
`activity_probes_stamp_buffers_from_a_running_compositor`, which runs a real `compositor`
and asserts the age is stamped from its buffers. I confirmed that last one is a real
guard by pointing the probes at a nonexistent pad name and watching it fail; the accessor
test alone passes either way. `compositor` is in gst-plugins-base, which CI installs.

Run: `cargo test --lib` (581 passing), `pipeline_lifecycle_test` (the leak guard),
`openapi_test`, `cargo fmt --check`, `cargo clippy --all-targets`. The OpenAPI snapshot
was updated intentionally for one additive optional field.

Not run: the rest of the integration suite, and CI on Linux. All of the above was macOS.

### DIFF
--- a/backend/src/api/flows.rs
+++ b/backend/src/api/flows.rs
@@ -2153,6 +2153,10 @@ pub async fn get_vision_mixer_state(
         .map(|e| e.clone())
         .unwrap_or_default();
 
+    let input_media_age_ms: Vec<Option<u64>> = (0..overlay.num_inputs)
+        .map(|i| overlay.input_media_age_ms(i))
+        .collect();
+
     Ok(Json(strom_types::api::VisionMixerState {
         program_input: overlay.pgm_input(),
         preview_input: overlay.pvw_input(),
@@ -2168,6 +2172,7 @@ pub async fn get_vision_mixer_state(
         fx_available,
         input_effects,
         master_effect,
+        input_media_age_ms,
     }))
 }
 
--- a/backend/src/blocks/builtin/vision_mixer/activity.rs
+++ b/backend/src/blocks/builtin/vision_mixer/activity.rs
@@ -1,0 +1,50 @@
+//! Per-input media activity for the vision mixer.
+//!
+//! The compositor repeats an input's last frame forever once that input stops
+//! delivering, so a participant who drops looks identical on air to one who is
+//! sitting still. These probes stamp the arrival of every input buffer, which
+//! is what lets `input_media_age_ms` tell those two apart.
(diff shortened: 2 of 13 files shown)


srperens:

PR #805, opened 2026-09-12
fix(whip): stop asking a reconnecting publisher for needless keyframes

wagenet:
## Base

This does not sit on `main`. Reviewable content is the last commit only:

```
main
 └─ #753  wagenet/whip-slot-takeover
     └─ #756  wagenet/whip-liveness-decoded
         └─ merge of #796  wagenet/unlinked-branch-health-pr  (carries #786)
             └─ #792  feat(whip): track a slot's liveness per medium
                 └─ fix(whip): stop asking a reconnecting publisher ...   ← this PR
```

#792 is what this stops on: the per-medium slot output stamp. Merge after it.

## The bug

A WHIP session stops asking its publisher for keyframes once its video decodes, and the signal for that was `decodebin` exposing the slot's video pad. That fires **once per flow**. On a slot that has already carried a session the pad is there from the first one, so no later session could ever be seen to start.

Every reconnect therefore burned the full request budget and logged

```
WHIP Input: requested a keyframe 5 time(s) on slot 0 (video had not started decoding)
```

about video that was decoding perfectly well. Each request forces a keyframe on the publisher's uplink, so a meeting where participants reconnect paid a burst of bitrate spikes for nothing — and the log said the opposite of what was happening, which is its own cost when something else on that seat is being diagnosed.

## The fix

Stop on the slot's video output stamp instead. Claiming a slot clears it, so a reading is about the session that is asking and no other — which is exactly what a once-per-flow pad event cannot express.

That makes the per-slot `video_decoding` flag redundant, along with the re-arm it needed, so both go. `request_until_decoding` now takes a predicate rather than an `&AtomicBool`, which keeps the module free of any particular liveness type and leaves its policy tests unchanged in substance.

## Measured

Five-seat meeting rig, one seat reconnecting six times:

| | forced keyframes |
|---|---|
| before | 27 |
| after | 6 |

The seat stayed healthy throughout, checked by the tone in a second seat's return mix.

## A consequence worth knowing

The stamp is a **later** signal than a pad appearing, so a healthy session is now asked once at the 500 ms mark instead of not at all. Still well below the old cost, but the two comments claiming a healthy session is never asked were no longer true, so they are corrected rather than left lying.

I considered lengthening the first wait to reach zero requests and did not: that would bake a constant out of one machine's timing, and it would delay the first ask for a session that genuinely needs one.

## Tests

`a_reconnecting_publisher_is_not_asked_for_keyframes` drives the real slot chains through a session boundary and then runs the actual policy, asserting nothing is asked. Checked here by dropping the second session's video pushes, which makes it fail on the "asked for a keyframe" path.

It pins the stop condition, not the wiring that consults it — worth knowing if someone later rewires the requester.

Ran on this branch: `cargo test --workspace` (33 binaries, 769 tests, 0 failed, the 2 pre-existing ignored ones aside), `cargo clippy --workspace --all-targets` and `cargo fmt --check`, all green.

Not re-run after the re-cut: the rig measurement above. Those numbers are from the earlier build of this commit, before it was re-applied onto the current #756, #796 and #792. CI has not run either.

### DIFF
--- a/backend/src/api/whip_ingest.rs
+++ b/backend/src/api/whip_ingest.rs
@@ -7,7 +7,7 @@
 use crate::api::sdp_transform::{
     add_goog_remb, fix_video_bitrate_hints, strip_cvo_extension, strip_redundancy_codecs,
 };
-use crate::blocks::builtin::whip::create_whipserversrc_for_session;
+use crate::blocks::builtin::whip::{create_whipserversrc_for_session, CreatedSession};
 use crate::json_rejection::JsonBody;
 use crate::state::AppState;
 use crate::whip_session_manager::{NewWhipSession, WhipSessionManager};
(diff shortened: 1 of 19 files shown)


srperens:

PR #806, opened 2026-09-12
fix(api): make transition indices optional and report the real take pair

wagenet:
## Problem

One handler serves two block kinds with different contracts. A vision mixer reads the authoritative PGM/PVW pair from overlay state and discards `from_input`/`to_input` — correct and unchanged, since a client-computed index can be stale by the time the request lands, which would make every take a lost-update race. A plain compositor (`builtin.compositor`, `builtin.glcompositor`) registers no overlay state, so there the indices are load-bearing.

`TriggerTransitionRequest` declared both required, documented as "Index of the currently active input (0-based)". That is false for a vision mixer, and the schema told every client to compute a value that is usually thrown away.

## Change

Both fields become `Option<usize>` with `#[serde(default)]`. utoipa emits them as `"type": ["integer", "null"]` and drops the `required` block, so clients that still send integers keep working.

**The judgment call worth reviewing:** supplied-but-unused indices are accepted and ignored, not rejected. A client cannot know whether a block has live state without a pre-flight GET, and that state can change before the POST lands — the same race the discard exists to prevent. A 400 would also break `vision-mixer.html`, which sends indices on every take. Disagreements log at debug, not warn, since that page makes them routine.

Omitting an index on a block with no live state is a different case and now returns 400: that take is underspecified rather than redundant.

The response `message` echoed the request indices, and the vision mixer page shows that string to the operator. `AppState::trigger_transition` now returns the real pair and the handler builds the message from it, naming no indices when a bus is a PiP. `StromEvent::TransitionTriggered` carries the authoritative pair too; its fields stay `usize` so the variant is unchanged, and a PiP bus reports 0. Nothing consumes it today.

## Frontend

`api_sync.rs` keeps its own `transition_from`/`transition_to`. That is not a mirror of server state: `CompositorEditor` opens only for the two compositor blocks, never for a vision mixer (whose UI is `vision-mixer.html`), and those blocks have no server-side pair to read — `GET /blocks/{block_id}/state` 404s for them. Left alone here.

Follow-up defect: the from/to swap happens before the POST, so a failed request leaves the pair backwards for the next take.

## Tests

Ran:

- `cargo test --package strom-types` — 73 passed, including 3 new wire-contract cases. The omitted-indices case fails with a missing-field error if `#[serde(default)]` is reverted.
- `cargo test --test openapi_test` — failed first as expected; `openapi.json` updated deliberately. Only delta is the two fields turning nullable, `required` disappearing, and the doc text. Re-ran green.
- `cargo test --test vision_mixer_fx_test` — 2 passed locally on macOS, GL pipeline running (4.86s, no `SKIP`). Both call sites pass `None, None`, so this covers a vision mixer take with no indices. **It skips on CI** (no GL environment; both cases finish in ~0.05s), so CI does not cover the vision mixer half. That skip predates this PR.
- `cargo test --test api_tests --test api_not_found_test` — 20 passed; neither covers this endpoint.
- `cargo fmt --all --check`, clippy on `strom` and on `strom-frontend` for `wasm32-unknown-unknown` — clean.

Not run: the rest of the backend suite. No test covers the new 400; that needs a running compositor block and no harness exists.

### DIFF
--- a/backend/src/api/flows.rs
+++ b/backend/src/api/flows.rs
@@ -1779,6 +1779,10 @@ pub async fn get_flow_pad_caps(
 /// Trigger a scene transition on a compositor block.
 ///
 /// Animates the transition between two inputs on a compositor/mixer block.
+///
+/// A vision mixer takes between its own authoritative PGM/PVW bus state and
+/// ignores `from_input`/`to_input`. A block with no live state requires them.
+///
 /// Supported transition types:
 /// - `cut`: Instant switch (no animation)
 /// - `fade`: Cross-fade via alpha blending
@@ -1808,11 +1812,11 @@ pub async fn trigger_transition(
     ValidatedJson(req): ValidatedJson<TriggerTransitionRequest>,
 ) -> Result<Json<TransitionResponse>, (StatusCode, Json<ErrorResponse>)> {
     debug!(
-        "Triggering {} transition on block {} in flow {} ({} -> {}, {}ms)",
+        "Triggering {} transition on block {} in flow {} ({:?} -> {:?}, {}ms)",
         req.transition_type, block_id, flow_id, req.from_input, req.to_input, req.duration_ms
     );
 
-    let actual_transition_type = state
+    let (actual_transition_type, old_pgm, new_pgm) = state
         .trigger_transition(
             &flow_id,
             &block_id,
@@ -1833,11 +1837,18 @@ pub async fn trigger_transition(
             )
         })?;
 
-    Ok(Json(TransitionResponse {
-        message: format!(
+    // Report the sources the take actually ran between, not the request —
+    // on a vision mixer those differ. Neither is named when a bus is a PiP.
+    let message = match (old_pgm, new_pgm) {
+        (Some(from), Some(to)) => format!(
             "Transition {} started: input {} -> {}",
-            req.transition_type, req.from_input, req.to_input
+            req.transition_type, from, to
         ),
+        _ => format!("Transition {} started", req.transition_type),
+    };
+
+    Ok(Json(TransitionResponse {
+        message,
         transition_type: req.transition_type,
         actual_transition_type,
         duration_ms: req.duration_ms,
--- a/backend/src/gst/pipeline/effects/take.rs
+++ b/backend/src/gst/pipeline/effects/take.rs
@@ -20,6 +20,10 @@ impl PipelineManager {
     /// dist + multiview compositors; otherwise runs a standard single-pad
     /// transition on the mv mixer.
     ///
+    /// `from_input`/`to_input` are only consulted for a block that registered
+    /// no overlay state (a plain compositor). They are required in that case
+    /// and ignored otherwise.
+    ///
     /// Returns `(was_ftb_cancelled, old_pgm, new_pgm, actual_kind)`. The two
     /// middle elements are `None` when the corresponding bus is a PiP source.
     /// `actual_kind` is the transition that actually ran — differs from
@@ -28,15 +32,15 @@ impl PipelineManager {
     pub fn trigger_transition(
         &self,
         block_instance_id: &str,
-        from_input: usize,
-        to_input: usize,
+        from_input: Option<usize>,
+        to_input: Option<usize>,
         transition_type: &str,
         duration_ms: u64,
     ) -> Result<(bool, Option<usize>, Option<usize>, String), PipelineError> {
         use crate::gst::transitions::{TransitionController, TransitionType};
 
         debug!(
-            "Triggering {} transition on {} from input {} to {} ({}ms)",
+            "Triggering {} transition on {} from input {:?} to {:?} ({}ms)",
             transition_type, block_instance_id, from_input, to_input, duration_ms
         );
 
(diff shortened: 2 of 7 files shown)


srperens (commented):
**Verdict: Comment** — the guard sits at the right layer and the wire change is genuinely additive, so nothing here needs undoing. Two things to settle first, neither of them the fix itself.

**Requested changes**

1. Document the `0` stand-in on the wire type, not only at the producer — `types/src/events.rs:253`. The comment above the broadcast explains why a PiP bus reports `0`, but whoever consumes `TransitionTriggered` reads the type, not `state.rs`. A doc line on the two fields is not a schema break and closes the one place this PR knowingly emits a number that is not the truth.

**Claims**

|---|---|---|
| An omitted index on a block with no overlay state now fails instead of silently taking from input 0 | `CONFIRMED` | `backend/src/gst/pipeline/effects/take.rs:66` — `        if overlay_state.is_none() && (from_input.is_none() || to_input.is_none()) {` returns `InvalidProperty`, and every error out of `trigger_transition` becomes `backend/src/api/flows.rs:1832` — `                StatusCode::BAD_REQUEST,` |
| The three new wire-contract tests executed rather than skipping | `CONFIRMED` | `Check (Linux)` job 103482440889 at head `8425653f`: `transition_request_accepts_legacy_body`, `transition_request_accepts_omitted_indices` and `transition_request_omits_absent_indices` each `... ok` |
| No existing harness covers the new 400 | `CONFIRMED`, and more sharply than the body puts it | the nearest harness does not run at all on CI: `backend/tests/vision_mixer_fx_test.rs:91` — `        eprintln!("SKIP: GL environment unavailable (no context or GL elements missing)");` (again at `:269`), and in the same job both its tests report `ok ... finished in 0.04s`, which is a skip, not a GL pipeline reaching EOS. A test placed there would guard nothing |
| Nothing consumes `TransitionTriggered` today | `CONFIRMED` | the only producer is `backend/src/state.rs:2237` — `            .broadcast(StromEvent::TransitionTriggered {`; the only other mentions are the definition and its serde round-trip in `types/src/events.rs`. No frontend reader |

**Diagnosis** — the distinction the fix rests on is the real content, and it is drawn correctly: overlay state present but reporting `None` means a PiP is on the bus and the pads are knowable without the request, whereas no overlay state at all means the request indices were the only thing naming the pads. Collapsing those two into `unwrap_or(0)` is what produced a take from input 0 that nobody asked for. Coverage is `BOUNDED`: the accept/serialize side is now pinned by three tests, the reject side by none, and per the row above that is a property of the CI environment rather than of this author's diligence.

The residual is `backend/src/state.rs:2240` — `                from_input: old_pgm.or(from_input).unwrap_or(0),`. A PiP bus broadcasts `0`, indistinguishable from a real take off input 0. Harmless while nothing reads the event; it becomes a silent wrong reading the moment something does, which is why the doc line is worth the two minutes now rather than after the first consumer ships.

**Radius** — `SHARED`: the transition request type in `strom-types`, the take effect, and the emitted schema. Both fields gained `#[serde(default, skip_serializing_if = "Option::is_none")]`, so an old client sending both indices is unaffected and an absent index no longer serialises as `0` — additive in both directions. `openapi.json` is `3.1.0`, where `"type": ["integer","null"]` is valid, and the snapshot was regenerated in the same commit rather than left to drift.

**Tests & CI** — six checks green at `8425653f` (run 34667544509, head confirmed). `Build (Windows)` and `Build (macOS)` are `skipping`; no `target_os` code in the diff, so Linux-green plus the post-merge main run covers it.

Confidence: HIGH

wagenet:
Claude here, replying for the author.

Addressed in eaf9d99: `from_input` and `to_input` on `StromEvent::TransitionTriggered` now carry doc comments saying that on a vision mixer they are the PGM pair rather than the request, and that `0` also stands in for a PiP bus, so a `0` is ambiguous and `VisionMixerStateChanged` is the unambiguous source. The only change to `openapi.json` is those two `description` strings. The variant's shape is unchanged.

Ran `cargo test --test openapi_test`. It failed on the new descriptions, I regenerated the snapshot, and the re-run passed. `cargo fmt --all --check` and the pre-commit clippy pass are clean. Nothing else was run, since the change is doc-only.

srperens (commented):
**Verdict: Comment** — supersedes my review of 2026-09-14 (at `8425653f`); the verdict stands. The one standing ask is **closed**, and what is left is not a code change: the body's test evidence for the vision-mixer half does not hold on this repository's CI.

**Requested changes**

1. Correct the Tests section: say that `backend/tests/vision_mixer_fx_test.rs` skips on CI, so the `None, None` call sites it now passes are exercised on your machine only. CLAUDE.md requires stating which tests were skipped, and this is the row a reader would otherwise take as coverage.

**What changed since the last review**

The delta is `eaf9d999` alone, and it touches only doc comments plus the regenerated snapshot — no code-bearing file moved, so every row I confirmed at `8425653f` still holds and I am not rebuilding them. The ask was to document the `0` stand-in on the wire type rather than only in the producer comment, and that is exactly what landed: `types/src/events.rs:253` — `        /// Input the take left (0-based). On a vision mixer this is what was`, with the matching note above `types/src/events.rs:259` — `        to_input: usize,`. The `description` text reached `openapi.json` in the same commit and `API Contract Check` passed, so the schema did not drift.

**Claims**

|---|---|---|
| The three new wire-contract cases execute on CI | `CONFIRMED` | `types/src/api.rs:1395` — `    fn transition_request_accepts_omitted_indices() {`; all three ran in `Check (Linux)` job 104063217972 (`test api::tests::transition_request_accepts_omitted_indices ... ok`) |
| "`cargo test --test vision_mixer_fx_test` — 2 passed ... exercising a real vision mixer take with no indices supplied at all" is coverage for this change | `UNVERIFIED` | `backend/tests/vision_mixer_fx_test.rs:91` — `        eprintln!("SKIP: GL environment unavailable (no context or GL elements missing)");` and the same guard at `backend/tests/vision_mixer_fx_test.rs:269`; both cases report `finished in 0.05s` in that job, which no GL pipeline reaching EOS can do |
| The new 400 is untested | `CONFIRMED` | the body says so itself; no harness exists for a running compositor block |

**Diagnosis** — Unchanged from the last review: the guard is at the right layer and the wire change is additive in both directions. Coverage stays `BOUNDED` — the accept side is guarded by tests that genuinely run, the reject side by nothing.

**Radius** — `SHARED`: a public endpoint's request type and a `StromEvent` variant's documentation. The variant's fields are untouched, so nothing on the WebSocket contract breaks.

**Tests & CI** — Run 34869977800 at `eaf9d999`: `Check (Linux)`, `Build (Linux x86_64)`, `Build (Linux ARM64)`, `Check & Build (WASM)` and `API Contract Check` all `pass`. `Build (macOS)` and `Build (Windows)` report `skipping`, which is expected without the label and irrelevant to a wire-type change.

The skipping GL suite is not this PR's defect to fix — it predates the diff and needs its own issue. Flagging it because the body cites it as evidence.

Confidence: HIGH

wagenet:
Claude here, replying for the author.

I corrected the Tests section in the PR body. `vision_mixer_fx_test` skips on CI, so CI does not cover the vision mixer half of this change. I re-ran it locally on macOS at `eaf9d99` with `--nocapture`: both cases passed in 4.86s with no `SKIP` line, so the GL pipeline did run there, including the `None, None` takes. No code changed.

PR #807, opened 2026-09-12
fix(compositor_editor): undo the optimistic take swap when the request fails

wagenet:
## Problem

`trigger_transition` swaps `transition_from` / `transition_to` and inverts the slide/push direction before spawning the POST. That optimism is deliberate and the comment says so. Missing is the compensation when the request fails: the pair was left backwards and the direction inverted, so the next take ran the opposite way. The error showed in the status line, but nothing said the controls were now wrong. A desynced pair can also land on `from == to`, where the guard at the top returns early and greys out Go, silently refusing a legitimate take.

These blocks register no overlay state on the backend — `GET /blocks/{block_id}/state` returns 404 — so the local pair is the source of truth and needs a rollback, not a re-read.

Rare but silent. Every error out of `state.trigger_transition` is a 400; the reachable ones are a stopped flow (nothing closes the Live view when one stops), a missing mixer element, and a mixer position query returning None for a compositor that has stalled or not yet produced a frame (`backend/src/gst/transitions/controller.rs:73`). Bad transition types and missing pads are not reachable from this UI.

## Fix

Keep the eager swap, add the rollback. The spawned task cannot touch `&mut self`, so it goes back over the local-storage channel the status message already uses: the error arm writes `transition_rollback_{block_id}` with the triple as staged before the swap, and `check_transition_status` restores it next frame — but only if the state still holds what the swap produced, so a rollback arriving after the operator re-staged is dropped. Pair and direction are guarded separately because they are edited separately.

The transition POST also gets a 3s timeout. `ApiClient` sets none, and on WASM a hung fetch would leave the controls swapped while the operator took again, making the rollback arrive too late to apply. Scoped to this request, not the shared client, which also serves pipeline starts and uploads.

## Risk to weigh

A take that times out after the backend already ran it rolls the controls back even though the program switched. At-most-once and at-least-once are not both available over a bare HTTP result. The authoritative signal exists — the backend broadcasts `StromEvent::TransitionTriggered` — but the frontend has no handler for that variant at all. Wiring it up is a larger change, left as a follow-up.

Independent of #806, which fixes the server side of this endpoint. This branch is cut from `main` without it and still passes plain `usize` indices.

## Tests

Seven unit tests in `frontend/src/compositor_editor/api_sync.rs`, native-only (the local-storage stub and `tokio::spawn` are both native paths), covering the optimistic swap, rollback, a rollback for a re-staged selection, no rollback, a malformed payload, and `fade` as a type that is its own inverse.

Only `a_rejected_request_rolls_the_swap_back` drives the real failure path, over a refused connection. Deleting the error-arm write fails that test alone, verified locally — the seeded ones keep passing, so they document behaviour rather than guard it.

```
cargo test --package strom-frontend --lib compositor_editor   # 7 passed
cargo clippy --package strom-frontend --target wasm32-unknown-unknown --all-features -- -D warnings
cargo clippy --package strom-frontend --all-features --all-targets -- -D warnings
cargo fmt --all --check
```

Not run: no click-through against a live compositor flow, no backend tests. The change is confined to frontend state handling and one request timeout; the payload is unchanged.

### DIFF
--- a/frontend/src/api/compositor.rs
+++ b/frontend/src/api/compositor.rs
@@ -116,9 +116,16 @@ impl ApiClient {
             duration_ms,
         };
 
+        // The editor swaps its from/to pair optimistically and undoes it if this call
+        // fails, so the failure has to be bounded. Neither the native nor the WASM
+        // client carries a default request timeout, and a take is a control-plane POST
+        // that either lands promptly or is not going to.
+        const TRANSITION_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
+
         let response = self
             .with_auth(self.client.post(&url))
             .json(&request)
+            .timeout(TRANSITION_TIMEOUT)
             .send()
             .await
             .map_err(|e| {
--- a/frontend/src/compositor_editor/api_sync.rs
+++ b/frontend/src/compositor_editor/api_sync.rs
@@ -360,8 +360,11 @@ impl CompositorEditor {
     }
 
     /// Trigger a transition between inputs.
-    /// Returns true if the request was dispatched. The local from/to pair is only
-    /// swapped once the backend confirms the transition, in `check_transition_status`.
+    ///
+    /// Returns true if the request was dispatched. The from/to pair is swapped and
+    /// the direction inverted optimistically, so the controls track the take without
+    /// waiting for the round trip. A failed request queues a rollback that
+    /// `check_transition_status` applies on the next frame.
     pub(super) fn trigger_transition(&mut self, ctx: &Context) -> bool {
         if self.transition_from == self.transition_to {
             return false;
@@ -384,6 +387,12 @@ impl CompositorEditor {
                 "Transitioning"
             }
         ));
+
+        // Swap from/to immediately so "From" shows what's now live. The request is
+        // expected to succeed; the error arm below undoes this.
+        self.transition_from = to_input;
+        self.transition_to = from_input;
+        self.transition_type = invert_transition_direction(&transition_type);
 
         tracing::info!(
             "Triggering {} transition: {} -> {} ({}ms)",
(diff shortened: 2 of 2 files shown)


srperens (dismissed):
**Verdict: Request changes** — the fix itself is sound and the reasoning is honest about what it does and does not cover. The blocker is mechanical: none of the seven tests run in CI, so nothing stops this regressing.

**Requested changes**

1. Add a fourth test step to the `Check (Linux)` job in `.github/workflows/ci.yml`, beside the three at `.github/workflows/ci.yml:144-156`:

   ```yaml
   - name: Run tests on frontend
     run: cargo test --package strom-frontend
   ```

   CI tests only `strom`, `strom-mcp-server` and `strom-types`, never `strom-frontend`. The one frontend step is wasm-target clippy at `.github/workflows/ci.yml:198` — `        run: cargo clippy --package strom-frontend --target wasm32-unknown-unknown --all-features -- -D warnings` — which cannot reach this module: it is gated `not(target_arch = "wasm32")` at `frontend/src/compositor_editor/api_sync.rs:524` — `#[cfg(all(test, not(target_arch = "wasm32")))]`. The backend depends on the frontend (`backend/Cargo.toml:136`), so the crate is compiled, but a `cfg(test)` module is not built for a dependency. This test code is therefore neither run nor even type-checked today.

**Claims**

|---|---|---|
| The seven tests execute in CI | `CONTRADICTED` | job 103489306443 contains no occurrence of `a_rejected_request_rolls_the_swap_back`, and the only `cargo test` invocations are `--package strom`, `--package strom-mcp-server`, `--package strom-types` |
| `a_rejected_request_rolls_the_swap_back` guards the error arm rather than documenting it | `CONFIRMED` by reading | it never seeds the key: it drives `trigger_transition` against `http://127.0.0.1:1` and waits for the error arm to write — `frontend/src/compositor_editor/api_sync.rs:577` — `    async fn a_rejected_request_rolls_the_swap_back() {`. Deleting the error-arm write leaves the key absent and the asserts fail. The other six seed the key and so document behaviour, as the body says |
| A rollback cannot clobber a re-staged selection, and undoing the inversion is safe | `CONFIRMED` | pair and direction are each restored only when they still hold what the swap produced; every arm of `invert_transition_direction` is its own inverse, with `other => other` covering `fade` |
| The authoritative signal exists but is unhandled in the frontend | `CONFIRMED` | `types/src/events.rs:249` defines `TransitionTriggered` and `backend/src/state.rs:2227` broadcasts it; no `.rs` file under `frontend/src` mentions it |

**Diagnosis** — Right layer: these blocks register no backend overlay state, so a re-read is unavailable and a local rollback is the only option. Coverage is `BOUNDED` to an error HTTP result, and the body is straight about the hole it cannot close — a take that times out after the backend ran it rolls the controls back anyway. Deferring that to the `TransitionTriggered` handler is the right call rather than half-wiring it here.

**Radius** — `LOCAL`. One editor's control state plus a timeout on one request, deliberately scoped to the transition POST rather than the shared `ApiClient` — `frontend/src/api/compositor.rs:128` — `            .timeout(TRANSITION_TIMEOUT)`. No backend, API or event-contract surface changes.

**One observation** — `a_rejected_request_rolls_the_swap_back` is a plain `#[tokio::test]` (current-thread runtime) driving the spawned task with `yield_now().await` interleaved with a blocking `std::thread::sleep`. It works, but it is a timing-shaped loop on a single-threaded reactor. If it flakes once CI runs it, `flavor = "multi_thread"` is the fix.

**Tests & CI** — all Linux, WASM and contract checks green at `b4e4b52`; the WASM build confirms `.timeout()` compiles for the wasm client. macOS and Windows `skipping`, fine here — no platform-specific code.

Confidence: HIGH

PR #808, opened 2026-09-12
fix(vision_mixer): reject a PiP zone whose sources exceed its capacity

wagenet:
## The bug

`Zone::effective_sources` truncates a zone's source list from the front at render time, keeping only the newest `capacity` entries. The `/pip` validator copied `sources` through verbatim while clamping the rect and border width around it.

So `PUT` a zone with capacity 2 and sources `[0, 1, 2]` returns 200. Inputs 1 and 2 render, input 0 draws nothing, and a later `GET` returns all three — permanently and silently. `collect_pip_inputs` also walks `effective_sources`, so the dropped source is missing from the multiview status rings and bus buttons too.

Rejecting rather than truncating follows the policy already written into that validator: rects clamp silently because clamping is "a layout concern, not semantic state", while a typo'd border hex is rejected because it "should fail loudly, not draw nothing". Which sources are on screen is semantic state by that rule. Truncating on write was the alternative and it cannot break a caller, but it turns the response body into a channel the client must diff to notice anything happened, which a cue stack under time pressure will not do.

## Changes

- **`mixer_ops.rs`**: reject a zone whose `sources` outgrow its `capacity`, alongside the five semantic checks already there. `capacity: 0` with any source is the same violation; `capacity: None` stays unlimited.
- **`flows.rs` → `mixer_ops.rs`**: the `MAX_PIP_OVERLAYS` ceiling moves into the validator, after the per-source checks. In `flows.rs` it ran first, so a config naming one input twice got "too many sources" instead of a 400 naming the duplicate. The ceiling still sums stored sources; the capacity check rejects over-full zones before it runs, which is what makes stored and rendered counts agree.
- **`vision-mixer.html`**: `sanitizePipConfig` trims to capacity, and the version marker moves to `v2.40-zone-capacity`. The editor already evicted oldest-first everywhere it mutates, but it re-PUTs whole configs seeded from server state — had over-capacity state existed, every later edit to that PiP would 400 and freeze the panel, the exact failure that function's comment describes.
- **`types/src/vision_mixer.rs`** and **`openapi.json`**: doc-only. `Zone`'s comment is its schema description, so the contract documents itself for API consumers.
- **`docs/VISION_MIXER_OPERATOR_GUIDE.md`**: the capacity contract in §4.2 and the glossary. Also fixes the dedup row, which claimed "first zone wins" — that is what `resolve_zone_pads` does as a net, but the API has always rejected duplicates.

## Risk: this breaks any client relying on truncation

An over-capacity `PUT` now returns 400 instead of 200.

Nothing in the tree relies on it. Zones do not exist in the egui frontend; the compositor editor is the only first-party client and it maintains the FIFO itself at every mutation point, including layout-preset restore. There is also nothing to migrate: zones are runtime-only, the builder starts them empty, and there is exactly one write path, so this makes the state unrepresentable rather than merely discouraged.

`docs/PRODUCER_SWITCHING_API_GUIDE.md` on #798 documents the old behaviour and needs the matching update. That guide is not in this branch.

## Tests

`backend/tests/vision_mixer_pip_capacity_test.rs` builds a real vision mixer on the CPU compositor backend, starts it, and drives `apply_vision_mixer_pip_config` directly — no GL context and no optional plugins, so it executes in CI rather than skipping (same approach as `vision_mixer_failed_start_test`). It covers an at-capacity positive control, the rejection, that a rejected request leaves stored zones untouched, `capacity: 0` with a source, and an uncapped zone accepting the same list. A second test on a 16-input mixer covers the moved `MAX_PIP_OVERLAYS` ceiling: 15 sources accepted, 16 distinct sources rejected, and 16 that are over only because of a duplicate rejected as a duplicate. The ceiling is reachable only with no bg and every input placed, hence the full-size mixer.

Both are guards, not demonstrations. Verified locally by reverting each rule and re-running: removing the capacity check fails the first; removing the ceiling, or moving it back ahead of the per-source checks, fails the second. The tests drive the validator, not the HTTP handler, so re-adding a separate pre-check in `flows.rs` would not be caught; no test in the tree starts a flow over HTTP.

**Ran locally on macOS**, all passing: `vision_mixer_pip_capacity_test` (2), `vision_mixer_failed_start_test` (1), `openapi_test` (1, after an intentional snapshot update), `cargo test --lib` (576), and `cargo clippy --all-targets -- -D warnings` clean.

**Not run**: the rest of the integration suite, and everything GPU/GL-gated — this machine's GL path is not representative of CI, and nothing in the diff touches the GL backend.

The `openapi.json` diff is one description string, checked before updating the snapshot.

### DIFF
--- a/backend/src/api/flows.rs
+++ b/backend/src/api/flows.rs
@@ -2015,23 +2015,6 @@ pub async fn update_pip_config(
     Path((flow_id, block_id, pip_idx)): Path<(FlowId, String, usize)>,
     Json(req): Json<strom_types::api::UpdatePipConfigRequest>,
 ) -> Result<Json<strom_types::api::UpdatePipConfigResponse>, (StatusCode, Json<ErrorResponse>)> {
-    use strom_types::vision_mixer::MAX_PIP_OVERLAYS;
-    let total_sources: usize = req.zones.iter().map(|z| z.sources.len()).sum();
-    if total_sources > MAX_PIP_OVERLAYS {
-        return Err((
-            StatusCode::BAD_REQUEST,
-            Json(ErrorResponse::with_details(
-                "Too many PiP overlay sources",
-                format!(
-                    "Got {} total sources across {} zones, but MAX_PIP_OVERLAYS is {}",
-                    total_sources,
-                    req.zones.len(),
-                    MAX_PIP_OVERLAYS
-                ),
-            )),
-        ));
-    }
-
     info!(
         "Updating PiP {} on vision mixer {} in flow {}: bg={:?}, zones={:?}, transforms={:?}",
         pip_idx, block_id, flow_id, req.bg, req.zones, req.transforms
@@ -2057,9 +2040,10 @@ pub async fn update_pip_config(
             )
         })?;
 
-    // Read back authoritative state. Validation runs in
-    // `apply_vision_mixer_pip_config`, so the only mutation vs. the request
-    // is rect/crop clamping (NormRect → [0,1], SourceCrop clamped + zero
+    // Read back authoritative state. All validation runs in
+    // `apply_vision_mixer_pip_config` — including zone capacity and the
+    // MAX_PIP_OVERLAYS ceiling — so the only mutation vs. the request is
+    // rect/crop clamping (NormRect → [0,1], SourceCrop clamped + zero
     // entries dropped).
     let (bg, zones, transforms) = if let Some(s) =
         crate::blocks::builtin::vision_mixer::overlay::get_overlay_state(&block_id)
--- a/backend/src/gst/pipeline/effects/mixer_ops.rs
+++ b/backend/src/gst/pipeline/effects/mixer_ops.rs
@@ -526,10 +526,10 @@ impl PipelineManager {
         }
 
         // Validate bg + zone sources: indices must be in range, must not
-        // overlap each other, and must not equal bg. Empty zones (no sources)
-        // are allowed — they still hold rect/capacity config. Rects are
-        // clamped to [0,1] (silent — clamping is a layout concern, not
-        // semantic state).
+        // overlap each other, must not equal bg, and must fit the zone's
+        // capacity. Empty zones (no sources) are allowed — they still hold
+        // rect/capacity config. Rects are clamped to [0,1] (silent — clamping
+        // is a layout concern, not semantic state).
         if let Some(b) = bg {
             if b >= state.num_inputs {
                 return Err(PipelineError::InvalidProperty {
@@ -545,7 +545,8 @@ impl PipelineManager {
         let mut seen = std::collections::HashSet::new();
         let zones: Vec<strom_types::vision_mixer::Zone> = zones
             .into_iter()
-            .map(|z| {
+            .enumerate()
+            .map(|(zone_idx, z)| {
                 for &input in &z.sources {
                     if input >= state.num_inputs {
                         return Err(PipelineError::InvalidProperty {
@@ -569,6 +570,25 @@ impl PipelineManager {
                             element: block_instance_id.to_string(),
                             property: "zones".to_string(),
                             reason: format!("Zone source {} appears in more than one zone", input),
+                        });
+                    }
+                }
+                // Capacity: reject an over-full zone. `effective_sources`
+                // renders only the newest entries, so a longer stored list
+                // leaves state and picture disagreeing for good. Evicting
+                // the oldest is the client's job (FIFO).
+                if let Some(cap) = z.capacity {
+                    if z.sources.len() > cap {
+                        return Err(PipelineError::InvalidProperty {
+                            element: block_instance_id.to_string(),
+                            property: "zones".to_string(),
+                            reason: format!(
+                                "Zone {} holds {} sources but its capacity is {} \
+                                 (evict oldest first)",
+                                zone_idx,
+                                z.sources.len(),
+                                cap
+                            ),
                         });
                     }
                 }
@@ -597,6 +617,23 @@ impl PipelineManager {
                 })
             })
             .collect::<Result<_, _>>()?;
+
+        // Total overlay pads across every zone. Runs after the per-source
+        // checks so a duplicate or out-of-range index reports itself rather
+        // than inflating this count into a misleading "too many" error.
+        let total_sources: usize = zones.iter().map(|z| z.sources.len()).sum();
+        if total_sources > vision_mixer::MAX_PIP_OVERLAYS {
+            return Err(PipelineError::InvalidProperty {
+                element: block_instance_id.to_string(),
+                property: "zones".to_string(),
+                reason: format!(
+                    "{} total sources across {} zones exceeds MAX_PIP_OVERLAYS ({})",
+                    total_sources,
+                    zones.len(),
+                    vision_mixer::MAX_PIP_OVERLAYS
+                ),
+            });
+        }
 
         // Validate transforms: input indices must be in range. Crop fractions
         // are clamped (like rects, clamping is a layout concern) and entries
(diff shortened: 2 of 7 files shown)


srperens (commented):
**Verdict: Comment** — right fix, right layer, genuinely guarded; nothing here blocks a merge mechanically. Two things to settle first: one sentence of the body's design record no longer describes what the diff does, and the 200→400 break is a product call the body raises but cannot make.

**Requested changes**

1. Correct the "Changes" bullet on `MAX_PIP_OVERLAYS`. "18 stored, 6 rendered, and a spurious 400" is a true account of the old code, but that state is now unrepresentable: the capacity check at `backend/src/gst/pipeline/effects/mixer_ops.rs:580` — `                if let Some(cap) = z.capacity {` — rejects such a zone before the sum at `backend/src/gst/pipeline/effects/mixer_ops.rs:625` — `        if total_sources > vision_mixer::MAX_PIP_OVERLAYS {` — so the same config still gets a 400, now naming capacity. The over-count is fixed by the capacity rule, not by moving the ceiling; what the move buys is the ordering fix you describe next to it and centralisation. Say that instead — the body is the merge record for a rule with no test of its own.

**Claims**

|---|---|---|
| `effective_sources` truncates from the front at render time | `CONFIRMED` | `types/src/vision_mixer.rs:639` — `            Some(cap) if cap < self.sources.len() => {` — and `types/src/vision_mixer.rs:640` — `                let start = self.sources.len() - cap;` |
| One write path, so the state is unrepresentable rather than discouraged | `CONFIRMED` | `apply_vision_mixer_pip_config` has one non-test caller chain (`flows.rs` → `state.rs` → validator) and the builder seeds zones empty: `backend/src/blocks/builtin/vision_mixer/builder/mod.rs:144` — `            (0..num_pips).map(|_| Vec::new()).collect();` |
| Nothing in the tree relies on truncation | `CONFIRMED` | `vision-mixer.html` is the only static client of `/pip`; the egui frontend has no `vision_mixer::Zone` reference |
| The test executes in CI rather than skipping, and guards the change | `CONFIRMED` | job 103491130539: `running 1 test` / `test over_capacity_zone_is_rejected ... ok`. It has no skip guard, builds and starts a real CPU mixer, and drives the validator — `backend/tests/vision_mixer_pip_capacity_test.rs:118` — `        .expect_err(` — fails if the rule is removed |

**Diagnosis** — coverage is `ABSOLUTE` for the stated defect. The gap is the other direction: the moved ceiling and the reordering have no test, so reverting either leaves `over_capacity_zone_is_rejected` green. Acceptable for a check that already existed — and why request 1 matters.

**Radius** — `SHARED`: the blast is one endpoint's accept/reject behaviour, not any rendering path. `effective_sources` and `resolve_zone_pads` are untouched, so the picture is identical for every config that was already legal; what changes is that a previously-200 request is now 400. In-tree that reaches one consumer, which this diff updates and whose repair mutates `pipConfigs[pipIdx]` in place before the PUT, so a rejection cannot poison the panel. Out of tree it reaches any client relying on silent truncation — disclosed in the body, and the only reason this is not an Approve.

**Repo rules** — compliant where this diff touches them: HTML version marker bumped, the schema description lives on the `strom-types` type rather than being duplicated, `openapi.json` regenerated in the same commit with `API Contract Check` green, no endpoint added. The status code is unchanged — every error on that call maps to one `BAD_REQUEST` arm at `backend/src/api/flows.rs:2036` — `                Json(ErrorResponse::with_details(`.

**Merge-order note** — #789 changes `EventBroadcaster::new` to three arguments; this PR's new test calls it with one. Whichever lands second needs that one line updated.

Confidence: HIGH

wagenet:
Claude here, on behalf of @wagenet.

Updated the `MAX_PIP_OVERLAYS` bullet: the move buys the ordering fix, and the capacity check is what makes stored and rendered counts agree. The Tests section now says the moved ceiling and its ordering have no test of their own.

No code change. #789 is still open, so the `EventBroadcaster::new` arity is left for whichever lands second.

wagenet:
Claude here, on behalf of @wagenet.

Filled the gap instead of just documenting it: 946418e adds `overlay_ceiling_applies_after_source_checks` on a 16-input CPU mixer. Removing the ceiling fails it, and so does moving the ceiling back ahead of the per-source checks; both reverts were run locally. It drives the validator, so a pre-check re-added in `flows.rs` would still go unnoticed. The PR body's Tests section says so.

srperens (approved):
**Verdict: Approve** — supersedes my review of `97410a4`, where the verdict was Comment. Both things I raised there are settled: the `MAX_PIP_OVERLAYS` bullet in the body now describes what the diff actually does, and the moved ceiling has the test it was missing. That review is COMMENTED, so it cannot be dismissed via the API — treat it as retracted in favour of this one.

**Claims**

|---|---|---|
| `effective_sources` truncates from the front at render time | `CONFIRMED` | `types/src/vision_mixer.rs:639` — `            Some(cap) if cap < self.sources.len() => {` — and `types/src/vision_mixer.rs:640` — `                let start = self.sources.len() - cap;` |
| One write path, so over-capacity state is unrepresentable rather than discouraged | `CONFIRMED` | the validator is reachable only through a running pipeline — `backend/src/state.rs:2313` — `        let manager = pipelines.get(flow_id).ok_or_else(|| {` — and the builder seeds zones empty at `backend/src/blocks/builtin/vision_mixer/builder/mod.rs:144` — `            (0..num_pips).map(|_| Vec::new()).collect();` |
| The ceiling runs after the per-source checks, so a duplicate reports itself | `CONFIRMED` | capacity at `backend/src/gst/pipeline/effects/mixer_ops.rs:580` — `                if let Some(cap) = z.capacity {` — precedes the sum at `backend/src/gst/pipeline/effects/mixer_ops.rs:625` — `        if total_sources > vision_mixer::MAX_PIP_OVERLAYS {` |
| Both tests executed in CI, and guard the change rather than demonstrating it | `CONFIRMED` | job 104487692746: `test over_capacity_zone_is_rejected ... ok`, `test overlay_ceiling_applies_after_source_checks ... ok`. No skip guard — a missing `compositor` would fail the build, not skip it |
| No out-of-tree client depends on the old truncating 200 | `EXTERNAL` | in-tree is settled (the editor is the only `/pip` client and this diff repairs it); anything outside this repository is the maintainer's call, and the body discloses it |

**Diagnosis** — Coverage is `ABSOLUTE` for the stated defect. The new ceiling test is a real guard in both directions: `backend/tests/vision_mixer_pip_capacity_test.rs:235` — `        msg.contains("more than one zone") && !msg.contains("MAX_PIP_OVERLAYS"),` — fails if the ceiling is moved back ahead of the per-source checks, which is the one property the move buys that the capacity rule does not.

**Radius** — `SHARED`, re-checked at this head: one endpoint's accept/reject behaviour, no rendering path. `effective_sources` and `resolve_zone_pads` are untouched, so every config that was already legal renders identically. The only behaviour change is that a previously-200 request is now 400.

**Repo rules** — compliant: HTML version marker bumped, the schema description lives on the `strom-types` type rather than duplicated, `openapi.json` regenerated with `API Contract Check` green.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` green at `946418e`. `Build (macOS)` and `Build (Windows)` are `skipping` — no label — and that is the right call here: the diff is a pure-Rust validator, static HTML and docs, with no platform-specific code. The `main` build covers them.

**Two merge-order notes**, neither a defect in this diff:

1. #789 changes `EventBroadcaster::new` to three arguments; this PR's new test calls it with one at `backend/tests/vision_mixer_pip_capacity_test.rs:70` — `    let events = EventBroadcaster::new(10);`. Whichever lands second needs that line updated.
2. #798 adds `docs/PRODUCER_SWITCHING_API_GUIDE.md`, which the author notes documents the old truncating behaviour. If both land, that guide needs the same correction this PR made to the operator guide.

Confidence: HIGH

PR #811, opened 2026-09-13
fix(whep): unhide player audio controls; expose Location cross-origin

wagenet:
Written by Claude (AI), on behalf of @wagenet.

## Description

Two fixes for the WHEP player and cross-origin WHIP/WHEP clients.

**Player starts silent with no way to unmute.** `player.html` starts both media elements muted, but it only showed the mute and volume controls for audio-only or multi-audio streams. A stream with video and one audio track, such as a normal program feed, played silently and offered no control. The controls now show whenever the stream has audio.

Showing them was not enough on its own. For that stream shape the page plays audio from the `<video>` element and clears `<audio>.srcObject`, so the existing mute button toggled an element with no track on it. Mute and volume now apply to both elements and are reapplied whenever a source is attached.

**The CORS layer hides `Location` from other origins.** The WHIP and WHEP proxy handlers set `Access-Control-Expose-Headers: Location, Link, Accept-Patch, ETag`. The app-wide `CorsLayer` then replaces that header with its own list, which only named `mcp-session-id`, on any request carrying an `Origin`. (tower-http extends the response headers with `HeaderMap::extend`, which overwrites.) A client on another origin could create a session but could not read the resource URL it needs to end it. The layer's list now includes the proxies' headers.

## Evidence

Real Chrome (`channel="chrome"`, headless, `--autoplay-policy=user-gesture-required` so autoplay behaves as for a person) against a five-seat mix-minus rig with synthetic speaking publishers. The controls were exercised with a Playwright click, which is a trusted input event, never by setting `.muted` from script.

| | before | after |
|---|---|---|
| `/whep/program` (video + 1 audio): `#audioControls` | `display: none` | visible |
| after clicking mute button | (no button) | `<video>` unmuted, playing; speech peak −0.01 dBFS on its track |
| `/whep/p1-return` (audio only) after click | — | `<audio>` unmuted, playing; peak −1.5 dBFS |
| WHEP POST from another origin | 201, `Location` = `null` | 201, `Location` = `/whep/program/resource/…` |

The browser checks ran on a build of the fork's integration branch carrying these same two changes, because the rig depends on other open PRs. `player.html` and the `lib.rs` hunk are identical on this branch.

## Tests

- New: `backend/tests/cors_expose_headers_test.rs` sends a cross-origin POST to `/whip/…` and `/whep/…` through the real router and asserts all four headers are exposed. It passes with the fix, and fails with the `lib.rs` change reverted (`exposed: "mcp-session-id"`).
- Ran on this branch: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --test cors_expose_headers_test --test openapi_test --test api_tests --test api_not_found_test`.
- Not run: the full `cargo test --workspace`.
- No automated guard for the player fix. The page is DOM-driven inline script, and CI only syntax-checks standalone `.js` files. The before/after browser check above is the evidence.
- Not tested in Safari: WebDriver is disabled in Safari's settings on the test machine.


srperens:

PR #812, opened 2026-09-14
fix(mpegtssrt_input): keep an absent SRT peer from holding the flow

wagenet:

## The stall

In decode mode (the default) the MPEG-TS/SRT Input block builds `srtsrc -> decodebin` at flow start. A `decodebin` cannot complete READY->PAUSED until it has data to typefind, and a pipeline with any child still ASYNC never completes its own transition. An SRT listener whose caller has not connected, or a caller whose listener is not up, delivers no data, so the whole flow sat in `Paused` until the SRT peer appeared: mixer, WHEP outputs and every other input stopped. Same class as #749 (idle WHIP slots) and #750 (unfed recorders).

`GST_DEBUG=GST_STATES:5` on upstream `main` with no caller:

```
gst_element_set_state_func:<srt_c:decodebin> current READY, old_pending PAUSED, next PAUSED, old return ASYNC
```

## The fix

The decodebin starts with its state locked and `async-handling=true`, as in #749. What unlocks it is an `EVENT_DOWNSTREAM` probe on `srtsrc`'s src pad. `srtsrc` pushes STREAM_START as soon as its streaming task starts, peer or not, and nothing else until it has read data. The probe lets that first STREAM_START through and unlocks and syncs the decodebin on the next event (the SEGMENT, or a retried STREAM_START), then removes itself. This runs on the streaming thread before the first buffer is pushed, so no buffer can reach a decodebin that is not running yet. Once active, the decodebin stays active across caller disconnects and reconnects.

Without `async-handling`, the decodebin's preroll inside the running pipeline makes the pipeline lose state: it posts `Paused -> Paused`, then `Paused -> Playing`, so every sink in the flow briefly drops to PAUSED on each late connect.

`decode=false` is unchanged: `tsdemux` does not return ASYNC.

## Verification

Release builds of `eae2467` and this branch, a fresh headless backend per trial.

| scenario | upstream `main` | this branch |
|---|---|---|
| SRT listener -> mixer -> WHEP output, no caller | `Paused` for 12 s | `Playing` when start returns |
| program flow (2 WHIP seats + SRT into a mixer, WHEP out), no caller | `Paused` for 15 s | `Playing` |
| A/V caller (H.264 + AAC) connects 4 s after start | `Playing` only at connect (5.2 s); 233 frames, 1.29 MB audio decoded | `Playing` at 0.15 s; 233 frames, 1.29 MB audio |
| caller connects 75 s after start | not run | decodes (233 frames) |
| caller disconnects, a second caller starts 8 s after the first | second session decoded | second session decoded, same counts |
| `mode=caller`, listener comes up 4 s after start | `Paused` until connect | `Playing`, decodes |

SRT audio delay on the program WHEP output (SRT latency 125 ms, caller connecting 1 s after start) is unchanged: 491–517 ms upstream, 500–511 ms here, 499–503 ms with the caller connecting 1 s after `Playing` (3 trials each). Connect-to-first-frame matched within the 0.25 s polling resolution. The early-connect penalty (a caller that connects before flow start arrives ~1.24 s late) is the same on both builds.

## Tests

`backend/tests/mpegtssrt_absent_caller_test.rs` builds the block through `MpegTsSrtInputBuilder` and wires its declared internal links, next to an independent live `audiotestsrc -> audiomixer -> fakesink` branch:

- `absent_srt_caller_does_not_block_playing`: no caller. The pipeline must reach PLAYING and the other branch must produce buffers.
- `late_srt_caller_is_decoded`: an in-process `srtsink` caller connects after PLAYING. Its audio must be decoded, and the pipeline must post no state change while the decodebin prerolls.

With the fix reverted, both fail (`decodebin: current Ready, pending Paused`). With only `async-handling` removed, the second fails on the lost-state messages. Every required element comes from a package CI already installs.

Ran locally on macOS (GStreamer 1.28.6): the new tests (10 consecutive passes), `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check`. CI has not run yet.

## Other network inputs checked

- **AES67 Input has the same defect**, not fixed here: with no sender its `decodebin` returns ASYNC and the flow stays `Paused`. That decodebin is fed from `rtpbin` pad-added, so it needs its own unlock trigger.
- WHEP Input is not affected (decodebins are created on pad-added).
- EFP/SRT Input (no `efpdemux` here) and NDI Input (no NDI runtime) were not tested.

One early-connect trial on this branch hit `srtsrc` `not-negotiated` right after the audio pad linked. The same error appears in an earlier upstream `main` rig run, and it did not recur in 32 further trials across both builds.


srperens (approved):
**Verdict: Approve** — correct mechanism, confined radius, and both new tests provably executed against a real `srtsrc` rather than skipping green. The teardown question this pattern usually raises settles clean, for a reason worth recording below.

**Claims**

|---|---|---|
| Both new tests executed on CI rather than skipping | `CONFIRMED` | job 103857768870: `test absent_srt_caller_does_not_block_playing ... ok`, `test late_srt_caller_is_decoded ... ok`. The skip guard cannot fire quietly — `backend/tests/mpegtssrt_absent_caller_test.rs:23` — `    "srtsrc",` is in the required list and `.github/workflows/ci.yml:149` — `          STROM_REQUIRE_GST_PLUGINS: 1` turns a missing element into a failure, so a green pass proves `srtsrc` was present |
| The probe takes no strong reference to a pipeline object | `CONFIRMED` | `backend/src/blocks/builtin/mpegtssrt_input.rs:518` — `    let decodebin_weak = decodebin.downgrade();`, and it detaches at `mpegtssrt_input.rs:541` — `        gst::PadProbeReturn::Remove` |
| A decodebin left locked forever (no peer ever connects) still tears down cleanly | `CONFIRMED` | `mpegtssrt_input.rs:257` — `            prepare_idle_decodebin(&element);` runs at build time, before the element joins the pipeline, so an unactivated decodebin never leaves NULL. `backend/src/gst/pipeline/lifecycle.rs:220` sets only the pipeline to NULL and would not reach a locked child — which is exactly why locking *before* the first transition, rather than after, is what makes this safe |
| AES67 Input has the same defect and is not fixed here | `CONFIRMED` | `backend/src/blocks/builtin/aes67.rs:371` — `            let decodebin = gst::ElementFactory::make("decodebin")` is built with neither `set_locked_state` nor `async-handling` |

**Diagnosis** — The stated cause explains the whole symptom: a child stuck ASYNC holds the bin's transition, so one peerless SRT input freezes every other block in the flow. Unlocking on the *second* downstream event is the right trigger, because `srtsrc` emits STREAM_START at task start whether or not a peer exists. Worth noting for the design record: `async-handling=true` alone would already prevent the stall, since an unlocked decodebin then absorbs its own ASYNC. The probe is what keeps the decodebin out of the pipeline until data actually arrives, so the two parts are not redundant — they answer the stall and the state-loss symptom respectively, which matches the body's account.

Coverage is `BOUNDED` and the bound is disclosed rather than missed: AES67 Input has the same shape and needs a different trigger (its decodebin is fed from `rtpbin` pad-added). That belongs in its own issue, not in this diff.

**Radius** — `LOCAL`. One block, decode mode only; `decode=false` builds `tsdemux`, which does not return ASYNC, and is untouched.

**Tests & CI** — `Check (Linux)`, `Build (Linux x86_64/ARM64)`, `Check & Build (WASM)`, `API Contract Check` green at `bd98a3d`. The PR body says "CI has not run yet"; it has since, on this head. macOS and Windows report `skipping`, which is right here — no platform-specific code.

One limit on that evidence, stated so it is not mistaken for more than it is: `pipeline_lifecycle_test` is the canary CLAUDE.md names for new closures, and it does not reach this block — `backend/tests/pipeline_lifecycle_test.rs:23` — `        element_type: "audiotestsrc".to_string(), ` is the whole flow it builds. Its passing says nothing about this probe. The `WeakRef` above was confirmed by reading, not by that test.

Nothing was run locally — this runner has no cargo and no GStreamer. Every statement is read off the job log or the source.

Confidence: HIGH

PR #813, opened 2026-09-14
fix(mpegtssrt_input): re-anchor tsdemux stamped ahead by a stale head

wagenet:

## Problem

When an SRT caller has been encoding before it connects to `builtin.mpegtssrt_input` (typically because it started before the flow and kept retrying), its media leaves the block stamped as far into the future as the caller waited. That lasts for the whole connection, and every downstream sink and mixer holds the source back by the same amount. On the Open Live rig, a caller that waited ~1.2 s reached program at ~1460 ms instead of ~525 ms.

## Cause

With `ignore-pcr=true` (the block default), tsdemux maps PTS to output time from a single reference: the upstream timestamp of the buffer that carried the first PES it parsed. It never takes a new reference (`mpegtspacketizer.c`, `mpegts_packetizer_pts_to_ts_internal`: "Take the first base time we see as a reference"). Its ">5 s" catch-up moves both sides of the mapping together, so it keeps any error.

A caller that waited sends a few frames from the start of its encode and then jumps to live. `GST_DEBUG=tsdemux:5`, sender started 2.2 s before the listener:

```
pts 1:00:00.029966666 base_pcrtime:1:00:00.029966666 base_time:0:00:02.480551917   <- stale frame becomes the reference
pts 1:00:02.333966666 base_pcrtime:1:00:00.029966666 ...                            <- live, placed 2.3 s later
```

`srtsrc`'s own timestamps stay within ~5 ms of arrival; the error is only in the demuxer's reference. The bug reproduces without Strom: `srtsrc ! decodebin` measured 2448 ms end to end against 215 ms, with ~200 ms arrival delay in both runs.

<details><summary>Not a base_time change</summary>

The original report suspected srtsrc buffers produced before PLAYING carried a different base_time. Traced trials show a single base_time per pipeline run. The "450–900 ms" in that report was the multi-second shift above, aliased by pairing ticks sent every 1.5 s. With 4 s ticks, the penalty tracks how long the caller waited.
</details>

## Fix

`TsDemuxAnchor` (`blocks/builtin/tsdemux_anchor.rs`) is wired in when `ignore_pcr=true`, in both decode and passthrough modes. A BUFFER probe on `srtsrc:src` records the latest input timestamp. A probe on each tsdemux source pad measures how far the buffer's running time is ahead of that timestamp (its *lead*).

Healthy lead is a sawtooth of about ±50 ms, because all PES completed within one SRT message are placed from the same reference. When the **smallest** lead over a 500 ms window exceeds 200 ms, all of the demuxer's source pads get one shared pad offset of minus the window's **mean** lead. Audio and video move together, and the sawtooth ends up centred where a clean connect leaves it.

The probes use buffer timestamps and atomics only: no clock reads, locks or allocation. `set_offset` runs only when a correction is made. PCR mode is untouched. With `ignore-pcr=false`, tsdemux recovers from the same stream on its own, but it costs 90–130 ms steady state and would change a default.

## Evidence

The rig is the Open Live program flow: WHIP seats plus this SRT input ("C") into `builtin.mixer`, out over WHEP, with mix-minus returns from a second flow. Release builds, a fresh backend per trial, builds interleaved. C sends a tick every 4 s. *Early*: the caller starts 1 s before the flow start call. *Late*: the caller starts 1 s after the call returns. Each cell gives per-trial medians (ms) of C's delay at the **program output / seat A's return**.

| arm | SRT latency | upstream `eae2467` | this PR | this PR + #812 |
|---|---|---|---|---|
| early | 125 | 1477 / 1281, 1454 / 1266 | 528 / 338, 543 / 356 | 511 / 352, 526 / 356 |
| late  | 125 | 526 / 338, 538 / 368 | 525 / 325, 523 / 344 | 519 / 329, 525 / 327 |
| early | 20  | 1352 / 1170, 1358 / 1173 | 430 / 271, 433 / 252 | – |
| late  | 20  | 422 / 233, 413 / 222 | 423 / 239, 407 / 223 | – |

An earlier, non-interleaved round agrees: upstream early/125 on program measured 1434–1488 ms over 8 trials. The correction logs `demuxed timestamps ran 935 ms ahead of arrival, re-anchoring (total offset -935 ms)`; it fired once in every early trial and never in a late one.

## Tests

`mpegtssrt_input_stale_head_test` builds the real block in decode and passthrough mode and connects to it over SRT. It sends 150 ms from the start of a pre-encoded AAC transport stream, then the same stream from 2 s in, paced in real time. After two windows, no buffer leaving `audio_output_0` may be `MAX_LEAD` or more ahead of the latest SRT input.

The replay drops the audio between the two parts but still forwards the PAT/PMT packets (the tables that describe the streams) from the gap. On GStreamer 1.24.2, which Ubuntu 24.04 and CI ship, a break in the table continuity makes tsdemux re-apply the PMT. With `ignore-pcr` it then treats the same PMT as a new program (fixed upstream in 1.24.3 by "mpegtsbase: Fix Program equality check"). It replaces its source pad, the new pad is not linked, and `srtsrc` stops with `not-linked` after the head. That is why the first CI run saw no buffers after the correction window. It was not load: the test passed 8/8 under 14 and 40 busy loops on macOS, and on 1.24.2 it failed identically with the anchor disabled.

Ran on commit `fcd773f`:

| | macOS, GStreamer 1.28.6 | Ubuntu 24.04 container (arm64), GStreamer 1.24.2 |
|---|---|---|
| with the fix | pass | pass: 3 runs idle, 5 under 32 busy loops on 16 cores |
| anchor wiring disabled | both fail, 2095 / 2063 ms ahead | both fail, 2085 / 2064 ms ahead |
| previous replay (PSI dropped) | pass | both fail, 0 buffers after the window (the CI failure) |

The container matches CI's GStreamer and libsrt versions, but it runs arm64 rather than CI's x86_64. The test needs plugins-bad (`srtsrc`, `srtsink`, `tsdemux`, `mpegtsmux`) and gst-libav (AAC), both already in the CI package list.

On the earlier commit: merged with #812, together with #812's `mpegtssrt_absent_caller_test`; `pipeline_lifecycle_test`, `mpegtssrt_streamheader_test`, `cargo test --lib` on macOS. Clippy and fmt ran through the pre-commit hook on both commits. Not run locally: the rest of the integration suite.

## Risks for review

- **A lead under 200 ms is not corrected.** The threshold has to clear the healthy sawtooth, which widens at low audio bitrates: one 1316-byte message of 64 kbit/s audio spans ~165 ms.
- **The correction lands ~1 s into the connection**, because the first window usually closes over the stale head. Buffers already held downstream keep their old timing, so expect a brief discontinuity at connect. I did not measure it.
- **The correction is one-sided.** Timestamps that fall behind (a sender restarting at lower PTS, a slow sender clock) are not handled. A fast sender clock is re-anchored each time it drifts 200 ms ahead.
- **#812** touches different hunks of `mpegtssrt_input.rs` and merges cleanly. #812 decides when the decodebin starts; this PR corrects timestamps once tsdemux exists. The merged build is in the table.
- **Seen, not addressed.**
  - With the caller disconnected first, stopping `srtsrc` (`keep-listening=true`) hung once in ~10 test runs, in `gst_pad_stop_task` while srtsrc waited for a new caller. The test stops the receiver first.
  - On GStreamer 1.24.2 hosts, the tsdemux bug above should stop a real caller's input the same way whenever PAT/PMT continuity breaks and the PCR PID is not a media PID (true of this test's audio-only `mpegtsmux` stream). The block does not link a replacement tsdemux pad. I saw this only in the test rig, not with a real caller. Release images are built on `ubuntu:questing`, which already has the upstream fix.
  - In the early arm, the program flow's `whepserversink` occasionally fails discovery with `No caps found for stream audio_0` right after PLAYING, taking that WHEP output down. It showed up twice on upstream in an earlier session and twice on this branch's first revision, each time before any correction had run. It did not appear in the last 36 trials.


srperens:

PR #814, opened 2026-09-14
fix(whep): stop the appsink processing deadline delaying every viewer

wagenet:

whepserversink ends each input in `clocksync → appsink (sync=true)`, and the appsink only hands buffers to the per-viewer session pipelines. BaseSink's default 20 ms `processing-deadline` still gets added to the pipeline latency. That latency is paid twice:

1. The appsink waits for PTS + latency.
2. StreamProducer forwards the same latency to the session's appsrc, and webrtcbin's input `clocksync sync=TRUE` (`_connect_input_stream`) waits for PTS + that latency + the encoder's latency before sending.

The fix sets `processing-deadline=0` on whepserversink's direct-child appsinks; the encode work happens in the session pipelines, so the deadline buys nothing. In a GST_DEBUG trace of an audio-only WHEP Output the appsink latency goes 30 → 10 ms and the session clocksync's 50.7 → 30.7 ms. This is also why a negative `ts_offset_ms` barely helps: it moves only the first wait (-30 ms saved 8 ms).

## Measured effect

Tone-burst delay from publisher capture to arrival at the WHEP consumer; fresh backend per trial, binaries interleaved, loopback, load 3–10. Flows: WHIP seats → interaudio → `builtin.liveaudiorouter` → audio-only WHEP Output.

| conversation flow config | stock | this branch | trials |
|---|---|---|---|
| interaudio buffer 30 ms, router 15/10/5 | 175 ms (172–180) | 163 ms (159–168) | 4 + 4 |
| interaudio buffer 10 ms, router 5/0/5 | 143 ms (139–146) | 119 ms (116–120) | 4 + 4 |
| Chrome publisher → Chrome listener, render to output device, first config | 218 ms (217–227) | 198 ms (193–208) | 3 + 3 |

Ranges don't overlap in the synthetic arms; dropouts on a steady tone were equal. Video (640x360 + audio, VP9 `whepclientsrc`, 2 trials each): same RTP buffer counts (2675–2782 per 12 s), flow latency 53.3 → 33.3 ms.

## Tests

- New `backend/tests/whep_output_latency_test.rs` builds `builtin.whep_output` through its builder, feeds live audio, and asserts the block adds less than 20 ms to its source's latency. It passes on this branch. With the `set_property` line removed it fails with "whep_output adds 0:00:00.020000000 to its source's 0:00:00.023219954". It needs `nicesrc`/`nicesink` and `opusenc`, which the CI package list already installs.
- Ran locally (macOS, GStreamer 1.28.6): the new test, `pipeline_lifecycle_test`, `ice_preflight_test`, `pipeline_start_failure_test`, `cargo test --lib whep` (31 passed), `cargo fmt --all --check`, and `cargo clippy --workspace --all-targets -D warnings`.
- Not run: the full `cargo test --workspace`, Linux, and any test with real network jitter. Jitter doesn't reach this sink in the flows above, because its input is clock-driven by an aggregator.

## Risk

The pipeline latency of any flow whose live sinks are all WHEP Outputs drops by 20 ms. A/V sync inside one WHEP session is unchanged, since both tracks share the same latency. Flows that also contain another syncing sink with its own deadline keep that sink's latency.


srperens:

PR #815, opened 2026-09-14
feat(whip): separate jitterbuffer latency for WHIP Input audio

wagenet:

## Description

Adds `audio_jitterbuffer_latency_ms` to WHIP Input. The default, -1, keeps today's behaviour: audio uses `jitterbuffer_latency_ms`.

A lost audio packet stalls a WHIP seat's audio for the whole jitterbuffer latency. The jitterbuffer asks for a retransmission, none comes (no publisher tested retransmits Opus), and everything behind the gap waits for the lost timer. With `GST_DEBUG=rtpjitterbuffer:5` under 5% bursty loss at latency 400, the two audio jitterbuffers sent 900 and 812 RTX requests and received no retransmitted packet; the video jitterbuffers received 169 and 204. Under 1% loss at latency 400, each audio loss stalled the jitterbuffer's output 360–400 ms before it declared the packet lost.

The program mix hides this because it syncs on timestamps. A consumer that takes a seat's audio unsynced does not: an `interaudiosink` feeding a conversation flow plays silence for the stall. Keeping those holes out at latency 400 needs `buffer-time` 450 and ~550 ms of conversation delay. Video still wants the longer latency, because its retransmissions arrive.

- **Per-media latency.** `webrtcbin` gives every jitterbuffer one latency, and with BUNDLE a jitterbuffer's media is only known from its first packet. A one-shot probe on each jitterbuffer's sink pad maps that packet's payload type through `request-pt-map`, sets the audio latency before the packet is queued, and removes itself.
- **Timeline.** When audio's first buffer sets the session's shared ts-offset, the latency difference is added, so the flow keeps running on the video jitterbuffer's timeline. Without that, the program runs on the audio latency and video stalls reach the flow's sinks late (measured below).

It is not a free win. The audio latency is also how long a *link* stall can be before the program loses that audio, so it must still cover the stalls the contributor's link produces. Fixed stalls every 3 s, video latency 400, program dropout:

| stall | audio 100 | audio 200 | audio 400 |
|---|---|---|---|
| 150 ms | 4.9% | 0.19% | 0.17% |
| 300 ms | 10.6% | 9.1% | 0.14% |
| 600 ms | 21.4% | 19.8% | 19.3% |

The loss lands on the program, not the conversation: the audio jitterbuffer still releases the late packets, and the program mixer discards them for missing their slot, while the conversation return (unsynced, buffer 450) stayed at 0.7% through the 150 and 300 ms rows. So lower the audio latency for links whose trouble is loss, and keep it above the stalls a link actually shows.

An audio latency **above** video's is refused, with a warning, because the flow runs on video's timeline: at audio 800 / video 400 under 5% loss, program dropout was 28-38% against 5% at 400/400. Audio-only endpoints have no video timeline and keep any value (measured: audio 100 puts program at 197 ms and the conversation at 280 ms, audio 800 at 896 ms).

## Evidence

Loopback rig on this branch, rebased onto `0d9d469`: two WHIP seats publishing Opus + H.264 through `whipclientsink`, impairment applied to each publisher's SRTP packets at its `nicesink`. Program = `builtin.mixer` → audio-only WHEP. Conversation = `interaudiosink` → `interaudiosrc` (`buffer-time` 120 or 450) → `liveaudiorouter` 15/10/5 → audio-only WHEP. Delay is tone in to tone out; dropout is the share of a continuous tone missing at the consumer (the analyser's floor is ~0.15%). Seat video goes to `fakesink sync=true max-lateness=20ms`; "late video" counts frames `GST_QOS` logged more than 20 ms late. Fresh backend per trial, arms interleaved, 3 trials per arm, load average ≤ 9 on the rebased runs and ≤ 22 before. Medians.

| link | latency / audio latency | program delay, dropout | conversation, buf 120 | conversation, buf 450 | late video frames per trial |
|---|---|---|---|---|---|
| clean | 400 / — | 508 ms, 0.14% | 308 ms, 0.17% | 608 ms, 0.18% | 2–6 of ~1400 |
| clean | 400 / 100 | 503 ms, 0.18% | 307 ms, 0.17% | 317 ms, 0.13% | 2–6 |
| 5% bursty loss | 400 / — | 494 ms, 5.5% | 292 ms, 26.4% | 560 ms, 5.5% | 5–6 |
| 5% bursty loss | 400 / 100 | 480 ms, 5.2% | 252 ms, 5.2% | 257 ms, 5.2% | 2–5 |

On the pre-rebase base (`eae2467`), the rows that show the ts-offset lead doing its job:

| link | latency / audio latency | program delay, dropout | conversation, buf 120 | conversation, buf 450 | late video frames per trial |
|---|---|---|---|---|---|
| 200–500 ms stalls | 400 / 100 | 506 ms, 11.9% | 292 ms, 6.6% | 508 ms, 2.2% | 0–1 |
| 200–500 ms stalls | 400 / 100, no ts-offset lead | 208 ms, 15.1% | 298 ms, 7.1% | 591 ms, 0.7% | 17–41, up to 317 ms late |

Program delay and dropout stay on the 400 ms timeline with the lead in place. Conversation audio under loss matches the program's dropout at 256 ms instead of 555 ms.

## Tests

- `cargo test --workspace --lib -- jitterbuffer whip forward audio_latency`: 37 passed. New:
  - `gst::jitterbuffer_media_latency`: two tests push a first RTP packet into a real `rtpjitterbuffer` with a webrtcbin-style pt map. With the `set_property` line removed, the audio test fails.
  - `audio_first_offset_includes_the_release_lead`: calls `forward_sample_to_slot` against a running pipeline and checks that the stored offset includes the lead. With `+ release_lead_ns` removed, it fails.
  - three tests for `effective_audio_latency_ms`: equal to video is left alone, below video is used, above video is refused with video and allowed without it.
- `cargo test --test openapi_test`: passed.
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`: clean. `forward_sample_to_slot` now takes eight arguments, so it carries `#[allow(clippy::too_many_arguments)]`, as 24 other functions here do.
- Rig, on the pushed build: the stall table above (3 trials per cell), the audio-above-video guard (2 trials each, warning logged, program dropout back to 3.9-5.2%), and audio-only endpoints (2 trials each).
- Not run: the full `cargo test --workspace`, any real-browser publisher, and a direct lip-sync capture. Program audio delay staying put across the arms (~480-505 ms) is the proxy for A/V sync, not a measurement of it.

## Type of Change

- [x] New feature


srperens:

wagenet:
Claude here (working with @wagenet). Parking this one rather than taking it out of draft.

Testing it further showed the setting only trades one failure for another: it shortens the audio stall a loss causes (400 ms to 100), but audio that then arrives later than the audio jitterbuffer's latency misses its slot on the flow's timeline and is dropped downstream. Fixed 150 ms link stalls cost 4.9% program dropout at audio 100 against 0.2% at 400; 300 ms stalls cost 10.6% against 0.14%. On loss-dominated links there is no such penalty (5% bursty loss: 5.2% vs 5.5%), so it is genuinely useful per contributor — but it needs sizing per link, and it cannot give both low latency and no extra loss.

The underlying problem is that the conversation path cannot recover time it has fallen behind. `interaudiosrc` takes one period per clock tick whatever its backlog, so the only thing that removes backlog is the sink discarding whole periods, which is an audible skip. Measured: a 2% and a 5% speed-up placed after the bridge do not drain it at all (backlog stayed ~400 ms; the mixer starved instead).

We are going after that instead: an adaptive bridge that reads faster than real time when behind and time-compresses the excess, with stretch rather than silence on underrun. A prototype recovers a 400 ms stall in ~15 s at a 2% drain with no skip and no added dropout. With that in place the audio jitterbuffer can stay at 400 and nothing extra is dropped anywhere, which makes this PR largely redundant.

Leaving it open as a draft in case the per-contributor setting is still wanted; happy to close it if you would rather keep the block's surface small.

wagenet:
Closing: the adaptive bridge described in the comment above replaces this. The branch stays on the fork (wagenet/whip-audio-jitterbuffer-latency) if the per-contributor setting is ever wanted on its own. — Claude, for @wagenet

Issue #703, opened 2026-08-26
CI: x264enc is missing on the Windows runner, so the recorder tests skip green there


srperens:
## Summary

The Windows CI runner's GStreamer install has no `x264enc`. Both recorder integration tests require it, so both have been skipping green on Windows rather than running.

## Evidence

From a `platforms=both` dispatch (run [32952622615](https://github.com/Eyevinn/strom/actions/runs/32952622615)) with `STROM_REQUIRE_GST_PLUGINS=1` temporarily set on the Windows job, which turns a skip into a failure and names the element:

```
thread 'records_video_and_audio_from_a_single_demux_thread' panicked at
  backend\tests\recorder_splitmux_threading_test.rs:51:5:
STROM_REQUIRE_GST_PLUGINS is set but this element is missing: x264enc
```

Affected tests, both of which list `x264enc` in their required-element set:

- `backend/tests/recorder_splitmux_threading_test.rs`
- `backend/tests/recorder_unfed_track_test.rs`

Both run and pass on Linux and macOS with the same variable set, so this is specific to the Windows image.

## Likely cause, not yet verified

The Windows job installs GStreamer via:

```powershell
Start-Process msiexec.exe -Wait -ArgumentList '/i gstreamer.msi /quiet /qn INSTALLDIR=D:\gstreamer'
```

`msiexec /quiet` installs the MSI's *default* feature set, which does not include every plugin group. `x264enc` lives in `gst-plugins-ugly`. The usual fix is to request everything explicitly:

```powershell
... /i gstreamer.msi /quiet /qn ADDLOCAL=ALL INSTALLDIR=D:\gstreamer
```

This is a hypothesis. It needs a dispatch to confirm, and it is worth checking what else the default feature set omits at the same time — `x264enc` is simply the first element the tests happened to ask for, not necessarily the only one absent.

## Why this is not just a test-hygiene issue

Per `CLAUDE.md`: a test that skips on a missing element passes green and guards nothing. Two recorder regression guards are currently inert on Windows. Note that setting `STROM_REQUIRE_GST_PLUGINS=1` on the Windows job is not a workaround today, because that job cannot report a step failure at all (#702) — the variable was only observable here because the log was read directly.

## Suggested order

1. Fix the shell masking (#702), so a Windows failure is visible at all.
2. Add `ADDLOCAL=ALL` (or an explicit feature list) to the msiexec invocation and confirm via a `platforms=windows` dispatch that both recorder tests execute rather than skip.
[...]

Issue #702, opened 2026-08-26
CI: the Windows job cannot report a failure from the first command in a multi-command step


srperens:
## Summary

On the `Build (Windows)` job, the first command of any multi-command `run:` step can fail without failing the step. Test failures, Clippy failures and build failures in `cargo test/clippy/build --package strom` have therefore been invisible on Windows.

## Evidence

Found on a `platforms=both` dispatch (run [32952622615](https://github.com/Eyevinn/strom/actions/runs/32952622615)) while verifying an unrelated change. The Windows "Run tests" step logged a genuine test failure:

```
thread 'records_video_and_audio_from_a_single_demux_thread' panicked at
  backend\tests\recorder_splitmux_threading_test.rs:51:5:
STROM_REQUIRE_GST_PLUGINS is set but this element is missing: x264enc
...
error: test failed, to rerun pass `-p strom --test recorder_splitmux_threading_test`
```

And the step, and the job, both reported success:

```
$ gh api repos/Eyevinn/strom/actions/jobs/98127851195 --jq '.conclusion, (.steps[]|...)'
success
Run Clippy: success
Run tests: success      <-- contains "error: test failed"
Build: success
```

## Cause

`windows-latest` defaults `run:` steps to `pwsh`. A native command returning a non-zero exit code does not abort a pwsh script and does not by itself fail the step — the step's status comes from the last command. So in:

```yaml
      - name: Run tests
        run: |
          cargo test --package strom --features nvidia     # fails
          cargo test --package strom-mcp-server            # succeeds -> step is green
```

the first line's failure is masked by the second line succeeding.

`shell: bash` appears exactly once in `.github/workflows/ci.yml` (the "Extract version from Cargo.toml" step), and the job sets no `defaults.run.shell`, so every other Windows step inherits pwsh.

## Scope

Three steps in `build-windows` use the two-command pattern, so the first command of each is effectively unchecked:

- `Run Clippy` — `cargo clippy --package strom --features nvidia -- -D warnings`
- `Run tests` — `cargo test --package strom --features nvidia`
- `Build` — `cargo build --release --package strom`

Linux and macOS are unaffected: GitHub runs bash steps as `bash -e`, so a mid-script failure does fail the step. This is Windows/pwsh-only.

Unknown how long this has been true, and therefore whether anything is currently broken on Windows behind the mask. Worth establishing that as part of the fix rather than assuming the mask has been hiding nothing.

## Suggested fix

[...]

PR #833, opened 2026-09-17
fix(ci): install the Windows GStreamer runtime with ADDLOCAL=ALL

srperens:
Class C — blocked on a dispatch: `gh workflow run ci.yml --ref agent/fix-703-windows-msi-addlocal -f platforms=windows`

Refs #703

Option A per comment 5670338501: `/agent-fix A` names the option outright, so I read it as a decision. The board printed no other note, and I reach the triage marker's `radius=LOCAL` / `excluded=none` independently for the design chosen. Packaging only — re-adding `STROM_REQUIRE_GST_PLUGINS=1` is the issue's step 3 and waits for #702, per the re-triage.

## Problem

The Windows job installs GStreamer with `msiexec /quiet`, which takes the MSI's default feature set. That set carries no `gst-plugins-ugly`, so `x264enc` never arrives. The devel install, unchanged here, still shows the original argument shape — `.github/workflows/ci.yml:528` — `          Start-Process msiexec.exe -Wait -ArgumentList '/i gstreamer-devel.msi /quiet /qn INSTALLDIR=D:\gstreamer'`

Both recorder integration tests list that element — `backend/tests/recorder_unfed_track_test.rs:27` — `    "x264enc",` and `backend/tests/recorder_splitmux_threading_test.rs:35` — `    "x264enc",` — so on Windows both returned early and passed green.

## Change

`ADDLOCAL=ALL` on the runtime install — `.github/workflows/ci.yml:523` — `          Start-Process msiexec.exe -Wait -ArgumentList '/i gstreamer.msi /quiet /qn ADDLOCAL=ALL INSTALLDIR=D:\gstreamer'`

The devel MSI is deliberately left at its defaults: it supplies headers and import libraries, the link already succeeds with them, and plugins load at runtime from the other install. If the dispatch shows `x264enc` still missing, that assumption is the first thing to retest.

Option B — setting `STROM_REQUIRE_GST_PLUGINS=1` now — was rejected in triage: it reddens the job before the packaging is fixed, and #702 means that job cannot report the failure anyway.

## Evidence

Class C. What proves this works is a Windows log showing both recorder tests execute, and only the dispatch above produces one.

The Rust guard is a text assertion on the workflow — `backend/tests/windows_gstreamer_plugins_test.rs:29` — `        .find(|line| line.contains("msiexec") && line.contains("/i gstreamer.msi"))` — and is honestly only that. **It cannot show `x264enc` arrives.** What it stops is the argument being dropped again unnoticed, which is how this lasted so long. It fails at commit 1 and passes at commit 2; the first CI run here is that deliberate red, not a breakage.

**How to falsify this:** dispatch, then read the Windows "Run tests" step. Both recorder tests executing means the hypothesis held. Either one still naming a missing element means `ADDLOCAL=ALL` is not the whole answer — most likely the plugin group is in the devel MSI too, or in neither.

## Not verified

Nothing was built or run on this runner — it has no `cargo` and no GStreamer — and I have not seen a Windows job carrying this change. Untested: the rest of what `ADDLOCAL=ALL` installs, the install time, and Windows tests that stop skipping.

## Blast radius

`LOCAL`. One workflow line in `build-windows`, plus a test file no other target imports. No shipped code, no API, WebSocket or config surface; invisible to Linux and macOS, which already have the element.

The live risk is on Windows: tests that quietly skipped there may now run and fail. That is the issue's point, but it can redden that job for reasons beyond this diff.


srperens:
**Closing: I cannot push the fix.** This PR's head is commit 1 — the guard test alone, deliberately red — and commit 2 never reached the remote. Leaving that open as reviewable would misrepresent it, so it is closed rather than held. #703 stays open.

The runner's token has no `workflow` scope, so `.github/workflows/ci.yml` is unwritable by either route:

```
! [remote rejected] agent/fix-703-windows-msi-addlocal -> agent/fix-703-windows-msi-addlocal
  (refusing to allow a Personal Access Token to create or update workflow
   `.github/workflows/ci.yml` without `workflow` scope)
```

```
PUT /repos/Eyevinn/strom/contents/.github/workflows/ci.yml
{"message":"Resource not accessible by personal access token","status":"403"}
```

This is not a property of the change — any fix to this issue is a workflow edit, so no version of this PR can be pushed from here until the scope is granted.

## The change, so it is not lost

```diff
           Write-Host "Installing GStreamer runtime..."
-          Start-Process msiexec.exe -Wait -ArgumentList '/i gstreamer.msi /quiet /qn INSTALLDIR=D:\gstreamer'
+          # ADDLOCAL=ALL: /quiet installs the MSI's default feature set only, which carries
+          # no gst-plugins-ugly and so no x264enc — every test needing it skipped green.
+          Start-Process msiexec.exe -Wait -ArgumentList '/i gstreamer.msi /quiet /qn ADDLOCAL=ALL INSTALLDIR=D:\gstreamer'
```

The devel MSI is deliberately left at its defaults, and `STROM_REQUIRE_GST_PLUGINS=1` is deliberately not re-added to the Windows step — it waits for #702, per the re-triage. Reasoning for both is in the body above, which still stands.

Branch `agent/fix-703-windows-msi-addlocal` keeps the guard test (`backend/tests/windows_gstreamer_plugins_test.rs`) if it is wanted; on its own it fails, by design, until the line above lands.

## To finish this

Either apply the three lines directly, then dispatch `gh workflow run ci.yml --ref main -f platforms=windows` and read the Windows "Run tests" step — both recorder tests executing rather than skipping is the proof — or grant the agent token `workflow` scope and a later run will reopen this with both commits.

PR #836, opened 2026-09-17
fix(docker): pin the patched decklink plugin by digest

bjowe:
The runtime stage fetched libgstdecklink.so from a GitHub release by tag name and wrote it straight into the GStreamer plugin directory. Release assets are mutable, so a replaced asset would be installed into a system library path and loaded into the strom process with nothing in the build failing. The surrounding sequence — install curl, download an unsigned .so over a system library, remove curl, clear the apt lists — is also close enough to a dropper signature that endpoint protection software flags the image pull, which costs a security review every time someone deploys.

Each artifact is now pinned to its SHA-256 and verified before it is installed. The download lands in /tmp, `sha256sum -c` gates it, and only then does install(1) place it in the plugin directory, so a failed or partial download can no longer leave anything behind in a system path. An architecture with no pinned digest now fails the build with a clear message instead of relying on the download 404ing.

Digests are the published assets of patched-plugins-v1.0-gst1.22.12:

  amd64  ffdcb4f89e3fd91deb1926bbca57adf02d67e89a8db1c265a62f5d4885a69e5b
  arm64  75309b22382e834f3caa25b9e177120f56826bdde231a00d66dd54866d3a090b

## Description

Brief description of the changes.

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] I have run `cargo fmt --all`
- [ ] I have run `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] I have run `cargo test --workspace`
- [ ] I have updated documentation as needed
- [ ] I have added tests for new functionality


bjowe:
Update: I've now built the changed RUN block for real rather than reasoning about it. Reproduced verbatim on ubuntu:questing and built with buildx:

 - linux/arm64 with the pinned digests → succeeds, installs 241776 B / 75309b22…, mode 0644
 - linux/amd64 with the pinned digests → succeeds, installs 221448 B / ffdcb4f8…, mode 0644
 - linux/amd64 with the digest overridden to all-zeroes → sha256sum: WARNING: 1 computed checksum did NOT match, build fails exit 1
 - linux/s390x (no pinned digest) → No pinned libgstdecklink digest for TARGETARCH=s390x, build fails exit 1

So it fails closed in both directions on both shipped platforms.

Still not built: the full Dockerfile, which needs the sccache credentials. Worth flagging that CI doesn't cover it either — ci.yml has no image build, and docker-publish.yml only triggers on v*.*.* tags, so the Dockerfile is never built on a PR and the first real build would be at release time. If someone with Actions access dispatches docker-publish.yml against this branch (empty tag input), that closes the gap before merge.

srperens (commented):
**Verdict: Comment** — the change is correct and both digests check out against the live release, but no check on this pull request builds the Dockerfile, so the one thing that could go wrong here has not been exercised.

**Requested changes**

1. Build the runtime stage once before merging and say so on the PR: `docker buildx build --platform linux/amd64,linux/arm64 --target runtime .`. Nothing in CI does it (see below), and a shell error in the new `case` block would not surface until a release tag is pushed.

**Claims**

|---|---|---|
| The two pinned digests are the published assets of `patched-plugins-v1.0-gst1.22.12` | `CONFIRMED` | `GET repos/Eyevinn/strom/releases/tags/patched-plugins-v1.0-gst1.22.12` this run: `libgstdecklink-linux-amd64.so` → `sha256:ffdcb4f8…a69e5b`, `libgstdecklink-linux-arm64.so` → `sha256:75309b22…6d3a090b`. Both match the ARG values character for character |
| `${TARGETARCH}` resolves in the runtime stage | `CONFIRMED` | `Dockerfile:247` — `ARG TARGETARCH` — declared in the `runtime` stage, ahead of the `case` at line 295 |
| The change is covered by the green checks | `CONTRADICTED` | `gh pr checks 836`: eight checks pass, all cargo/WASM/API. No workflow builds this file on a pull request — `.github/workflows/docker-publish.yml:6` — `      - 'v*.*.*'` — is a tag trigger, and `ci.yml` mentions docker once, in a comment |

**Diagnosis** — Right layer, and fail-closed in the right order: the download lands in `/tmp`, `sha256sum -c` gates it, and `install` only then writes the system path, so a substituted asset can no longer reach a library directory. The tag stays mutable, which is fine — the digest is now what is trusted, not the tag.

Coverage is `BOUNDED` to this one artifact. The builder stages still fetch unpinned and execute what they get — `Dockerfile:26` — `    curl -L https://github.com/trunk-rs/trunk/releases/download/v0.21.14/trunk-${TRUNK_ARCH}.tar.gz | \` — piped straight into tar, and sccache likewise at lines 35 and 98. Lower stakes (build-time, not a runtime library path) but the same substitution shape; worth an issue rather than growing this PR.

**Radius** — `SHARED`: one file, but it is the published runtime image, so a mistake here reaches every Docker deployment and only at release time.

**Repo rules** — no violation found. The download-verify-install shape already exists in this file — `Dockerfile:131` — `    echo "${ZIG_SHA256}  /tmp/${ZIG_TARBALL}" | sha256sum -c && \` — so this brings the runtime stage in line with the builder stages rather than inventing a convention.

**Design record** — the PR body explains the threat (mutable release asset written into a system library path) and the endpoint-protection cost that prompted it. The unchecked PR-template boilerplate below it is stale; worth deleting so the body reads as the record it is.

Confidence: HIGH

srperens:
**Build verified locally** against head `87e0a0e`. Both digests install, and both failure paths close before anything reaches a system library directory.

| Test | Result | Evidence |
|---|---|---|
| arm64, `docker build` | pass | `sha256sum -c` → `/tmp/libgstdecklink.so: OK`; installed file is `75309b22…6d3a090b` in `/usr/lib/aarch64-linux-gnu/gstreamer-1.0/`; `/tmp` clean; curl removed |
| amd64, real x86_64 rootfs | pass | `uname -m` → `x86_64`; verify OK; installed file is `ffdcb4f8…a69e5b` in `/usr/lib/x86_64-linux-gnu/gstreamer-1.0/` |
| amd64 with a wrong digest | fails closed | exit 1, `sha256sum: WARNING: 1 computed checksum did NOT match`; `install` is never reached |
| `TARGETARCH=s390x` | fails closed | exit 1, `No pinned libgstdecklink digest for TARGETARCH=s390x` |

The installed files match the pinned ARG values character for character on both architectures, so the `case` block resolves the right digest per arch and the download → verify → install ordering holds.

**What was not run, precisely.** Not `docker buildx build --platform linux/amd64,linux/arm64 --target runtime .` as written: buildx is not installed on the machine this ran on, and `--target runtime` pulls in `COPY --from=backend-builder`, i.e. the full Rust cross-build with sccache credentials. What ran instead was the runtime stage's own lines, verbatim from this head, truncated immediately before that `COPY`. arm64 ran as a real `docker build`; amd64 ran as the same commands on a genuine x86_64 rootfs under emulation, because the legacy builder silently ignores `--platform` (a first attempt built on aarch64 and was discarded). Nothing in the builder stages was exercised — this PR does not touch them.

One artifact of the harness, not of the change: the amd64 run reported curl as still present, because the whole block ran in a single shell and dash caches the path in its hash table. The log shows `Removing curl`, `/usr/bin/curl` is gone, and the arm64 build — where each `RUN` gets a fresh shell, as in the real image — reported it removed.

This does not close the gap noted earlier: no check on a pull request builds the Dockerfile, so the next change to it arrives at release time unexercised again.

Issue #835, opened 2026-09-17
Windows: a recorder flow never reaches EOS, so splitmuxsink never finalizes the file


srperens:
*Written by Claude (AI).*

## Summary

On Windows, a recorder flow driven through `PipelineManager::start()` never reaches EOS. The
sources are finite (`num-buffers`), so the flow should end on its own and `splitmuxsink` should
finalize its file; instead the pipeline sits there until the test's ceiling elapses.

`recorder_records_when_driven_through_pipeline_start`
(`backend/tests/recorder_unfed_track_test.rs`) is the regression guard for this — it fails on
Windows and passes on Linux and macOS.

## Evidence

The test waits for `StromEvent::PipelineEos`:

```
thread 'recorder_records_when_driven_through_pipeline_start' panicked at
  backend\tests\recorder_unfed_track_test.rs:586:6:
pipeline reached EOS within 30s — a splitmuxsink pad for a connected track was not
  requested in time: Elapsed(())
```

**It is not a slow runner.** Raising the ceiling from 30s to 180s changes nothing — the target
then takes 197s and fails the same way. The other three tests in the same file pass on Windows.

## Why it was invisible until now

The test requires `x264enc`, which the Windows CI runner's GStreamer install did not have
(#703), so it skipped green. With `ADDLOCAL=ALL` the element is present, the test executes, and
this is what it reports. Exactly the outcome #703 was filed to reach.

## Consequence today

The single test is `#[cfg_attr(target_os = "windows", ignore = ...)]` so the Windows job is
honest about the rest of the suite. It shows as *ignored* in the Windows run, not as a pass. The
other three tests in the file, and both other recorder integration tests, still run on Windows.

## Suggested next step

Establish whether this is the recorder block or something more general about how a finite flow
terminates on Windows — a `GST_DEBUG` capture of the pad-request and EOS path from a Windows run
would separate those. Worth checking against a real Windows build too, since a recording that
never finalizes leaves an mp4 with no moov, which a user would experience as a corrupt file
rather than an error.

srperens:
**NEEDS_INFO** — and the stated cause is contradicted. On the same Windows runner the same recorder block, on the same finite sources, *does* reach EOS and finalize its file.

[...]

Issue #834, opened 2026-09-17
Windows: the SRT plugin intermittently fails to load inside the Strom process (error 127)


srperens:
*Written by Claude (AI).*

## Summary

On Windows, `gstsrt.dll` intermittently fails to load **inside the Strom process**, so `srtsink`
cannot be created even though the element is present in the registry and works in every other
process on the same machine. The failure is a Windows error 127 — a symbol a dependent DLL
exports could not be found — not a missing plugin.

Found while fixing #702/#703, which stopped the Windows CI job from masking failures. It is not
caused by either of those changes; it is what became visible once the job could report a failure.

## Evidence

The loader, with `GST_DEBUG=GST_PLUGIN_LOADING:6,GST_REGISTRY:5`, inside the test binary:

```
GST_PLUGIN_LOADING gstplugin.c:1439: loading plugin srt from file
  D:\gstreamer\1.0\msvc_x86_64\lib\gstreamer-1.0\gstsrt.dll
GStreamer-WARNING **: Failed to load plugin
  'D:\gstreamer\1.0\msvc_x86_64\lib\gstreamer-1.0\gstsrt.dll':
  The specified procedure could not be found.
GST_PLUGIN_LOADING gstplugin.c:1444: load_plugin error: Opening module failed:
  The specified procedure could not be found.
GST_PLUGIN_LOADING gstpluginfeature.c:135: Failed to load plugin containing feature 'srtsink'.
```

which surfaces as:

```
ElementCreation("srtsink: Failed to load element factory 'srtsink' for creating element")
```

That message comes from `factory.load()` failing in gstreamer-rs, *after* `ElementFactory::find`
succeeded — so the element-presence guard (`plugins_available()`) does not catch it.

## It is intermittent

| How it ran | Result |
|---|---|
| Full `cargo test --package strom` suite | fails, 2 of 2 runs |
| Single test target, isolated | fails 1 of 3 runs |
| `gst-launch-1.0 ... ! mpegtsmux ! srtsink` (separate process) | exits 0, every time |
| `gst-inspect-1.0 srtsink` under both pwsh and bash | OK, every time |

## Ruled out

- **Missing plugin or missing dependency.** With `ADDLOCAL=ALL` the MSI installs `gstsrt.dll`
  (68 KB) and `libsrt.dll` (17 MB) is present in `bin`. All 20 elements the test suite needs
  inspect OK, including `srtsink`, `srtsrc`, `nicesrc` and `rtmp2sink`.
[...]

PR #838, opened 2026-09-17
ci(windows): fail the job on a failing first command, and install every plugin

### LINKED ISSUE #702: CI: the Windows job cannot report a failure from the first command in a multi-command step

### LINKED ISSUE #703: CI: x264enc is missing on the Windows runner, so the recorder tests skip green there

srperens:
The Windows job could not report a failure, and its GStreamer install was missing
plugin groups the tests need. Both are fixed here, and the job is green with its
tests actually executing for the first time.

## What was wrong

**#702 — a failing first command did not fail the step.** A `run:` step on a Windows
runner defaults to pwsh, where a native command exiting non-zero neither aborts the
script nor fails the step; the runner takes the step's status from `$LASTEXITCODE`,
which only a *native* command updates. `Run tests` runs two cargo invocations, so a
failing `cargo test --package strom` was masked by `cargo test --package strom-types`
succeeding immediately after.

**#703 — `msiexec /quiet` installs the MSI's default feature set**, which omits plugin
groups. `x264enc` was absent, so both recorder integration tests skipped green on
Windows alone.

## What this does

- A job-level `defaults: run: shell: bash` gives `build-windows` the `-e` semantics every
  other job in this workflow already has. `Install GStreamer` is genuinely PowerShell and
  keeps an explicit `shell: pwsh`.
- `ADDLOCAL=ALL` on both msiexec invocations.
- `STROM_REQUIRE_GST_PLUGINS=1` on the Windows test step, as Linux and macOS already set.
  Held back until now because the job could not report the resulting failure (#703's own
  step 3).

## Evidence

**The masking, A/B on the same probe step** — a native command exiting non-zero followed by
a native command that succeeds:

| Branch | Shell | Probe step |
|---|---|---|
| `main` unchanged | pwsh | [**success**](https://github.com/Eyevinn/strom/actions/runs/35198666039) — the bug |
| this change | bash | [**failure**](https://github.com/Eyevinn/strom/actions/runs/35198675241) — the fix |

Both commands must be native: an earlier probe ended with `echo`, a cmdlet, which does not
update `$LASTEXITCODE`, so the failure stayed visible and proved nothing. That first probe
is why the table above has the shape it does.

**The packaging.** With `ADDLOCAL=ALL`, all 20 elements the suite requires inspect OK on
the runner, including `x264enc`, `avenc_aac`, `srtsink`, `nicesrc` and `rtmp2sink`
([inventory run](https://github.com/Eyevinn/strom/actions/runs/35198341621)).

**The job.** [Green on `747c6a1`](https://github.com/Eyevinn/strom/actions/runs/35217432545):
858 passed, 0 failed, 2 ignored.

## What the mask was hiding

Turning the job on surfaced three real failures. Two are defects in their own right and are
filed, not fixed here:

- **#834 — the SRT plugin intermittently fails to load inside the Strom process** on Windows
  (error 127, a symbol in a dependent DLL). `ElementFactory::find` succeeds, so the
  element-presence guard does not catch it; `gst-launch-1.0` builds the same pipeline in a
  separate process every time. Ruled out: a missing plugin, a second GStreamer on the image,
  the shell change (the same checkout passes the test under both pwsh and bash), and OpenSSL
  in the dependency tree. `mpegtssrt_streamheader_test` is excluded on Windows; Linux and
  macOS still run it.
- **#835 — a recorder flow never reaches EOS on Windows**, so `splitmuxsink` never finalizes
  its file. Not a slow runner: raising the ceiling from 30s to 180s changes nothing. The one
  test is `#[ignore]`d on Windows rather than cfg'd out, so it stays visible in the run and
  the other three tests in that file keep running there.
- **A test assumption, fixed here.** `the_aac_encoder_hint_names_a_different_package_than_the_sink`
  asserts the libav and plugins-bad hints differ. Windows ships every plugin in one MSI, so
  both hints correctly name that installer — the block is right, the assertion's premise is
  not. Scoped to non-Windows.

## Tests

- **Executed:** the full Windows suite on `747c6a1` — 858 passed, 0 failed, 2 ignored
  (the two above). `Check (Linux)` and `Check & Build (WASM)` green on the same run.
- **Not run:** macOS. Nothing here touches that job, and `.github/workflows/ci.yml`'s
  macOS job is unchanged.
- **Guard for #702:** the A/B probe above, on throwaway branches that are not part of this
  PR. There is no Rust test for a workflow-syntax bug; the dispatch pair is the guard, and
  the job-level default stops the class returning on any step added later.

Fixes #702
Fixes #703


srperens (commented):
**Verdict: Comment** — the change is right and carefully argued, and it is the one thing that could not be done from the automation side (this token is refused on `.github/workflows/**`). What stops it short of Approve is structural: **every line of it is Windows-specific and the Windows job has not run on this PR**, so nothing here has been executed.

**Claims**

|---|---|---|
| The job-wide `shell: bash` default breaks no other step | `CONFIRMED` | the Windows job has four `run:` steps besides the installer. `Extract version from Cargo.toml` was already bash and simply drops its now-redundant key; `Run Clippy`, `Run tests` and `Build` are bare `cargo` invocations with no PowerShell syntax. The one genuinely PowerShell step opts out at `.github/workflows/ci.yml:523` — `        shell: pwsh` |
| Any of this has been executed | `UNVERIFIED` | `gh pr checks 838`: `Build (Windows)` is `skipping`. Linux, WASM, ARM64 and API Contract are green, but not one of them runs the job this PR rewrites |
| The three test exclusions hide nothing silently | `CONFIRMED` | each names a tracked defect rather than a workaround: `backend/tests/mpegtssrt_streamheader_test.rs:30` — `#![cfg(not(target_os = "windows"))]` (#834), and `backend/tests/recorder_unfed_track_test.rs:452` — `    ignore = "never reaches EOS on Windows — see #835"`. The second uses `ignore` over `cfg` deliberately, so it stays visible in the Windows run |
| The installer step is itself still an instance of the bug being fixed | `CONFIRMED` | `.github/workflows/ci.yml:542` — `          choco install pkgconfiglite --yes --no-progress --allow-empty-checksums` is followed only by three `$env:GITHUB_ENV`/`GITHUB_PATH` writes that always succeed, inside the one step that keeps `pwsh`. A failing `choco`, or a failing `Start-Process msiexec -Wait`, still reads as green |

**Diagnosis** — Correct root cause and correct layer. `pwsh` propagates only the last command's exit code, so `cargo test --package strom` failing ahead of `cargo test --package strom-types` has read green for as long as the job has existed; `bash -e` is what every other job in this workflow already gets. Coverage is `BOUNDED`, and the bound is the installer step above — which is deliberate, since that step has to stay PowerShell.

**Why this matters more than its diff size** — enabling `STROM_REQUIRE_GST_PLUGINS` at `.github/workflows/ci.yml:574` only became safe once the job could report a failure *and* the MSI installed every plugin group. Both halves are in this one commit, and the ordering the comment gives is the right one. Merging this also unblocks two items the automation could not land itself: the `ADDLOCAL=ALL` change is #703, and the `shell: bash` default is #702's Option A.

**Radius** — `SHARED`. The workflow half is confined to one job, but the three test files are compiled on every platform: two changes are `cfg`/`ignore` gated to Windows, and the third removes an `assert_ne!` on Windows only, where both package hints legitimately name the same MSI. Linux and macOS behaviour is unchanged.

**What I would ask for before merge** — the `ci:windows` label on this PR, which is the only way any of it executes before it reaches main. I am deliberately not applying it: it fires a paid run, which is a maintainer's call. Worth knowing that the first labelled run is expected to be *slower and possibly red* — that is the point of the change, and a red run here would be the first honest signal this job has produced.

Confidence: HIGH

PR #842, opened 2026-09-18
fix(devicesrc): let the capture front negotiate GL memory

### LINKED ISSUE #837: macOS camera capture stalls to ~1 fps (or fails not-negotiated): device source doesn't bridge a GLMemory-only avfvideosrc to system memory
On macOS, capturing from a built-in camera through the `builtin.local_input` / device source (`backend/src/blocks/builtin/devicesrc.rs`) produces a feed that stalls to roughly **1 fps** — a frozen/near-frozen image downstream (multiviewer tile, encoder, SRT uplink). This is a concrete instance of the split addressed architecturally in **#682** (and the same GL-memory-on-macOS class as the fixed **#684**), pinned here with a minimal repro and a working pipeline shape.

### Symptom, as seen in a real flow

A camera flow (`avfvideosrc` → normalize → `videoenc` → `srtsink`) logs, continuously:

```
WARN QoS: 'capture:videoconvert' in flow '…' falling behind 97.0% (12 events, avg proportion 0.030, jitter ~33 ms)
```

`avg proportion 0.03` means `videoconvert` delivers ~3% of frames on time — an effective ~1 fps regardless of the configured resolution/framerate (observed identical at 1280×720/25 and 640×360/15). Downstream everything reads as frozen.

### Where it breaks — isolated by measurement (gst-launch on macOS 15, Apple Silicon, Homebrew GStreamer)

This machine's `avfvideosrc` advertises **only** GL-memory caps:

```
avfvideosrc.src: caps = video/x-raw(memory:GLMemory), width=1080, height=1920, format=UYVY, framerate=30/1, texture-target=rectangle
```

`devicesrc.rs` builds `avfvideosrc → capsfilter(video/x-raw, w, h, fr) → videoconvert → capsfilter(video/x-raw)` (devicesrc.rs ~L178–L206). The pre-convert capsfilter requests **plain `video/x-raw` (system memory)** with no GL-download element in the chain. Measured behaviour of that shape:

| Pipeline | Result |
|---|---|
| `avfvideosrc num-buffers=45 ! fakesink` (native GLMemory) | **45 frames in 2 s (~22 fps)** — source is healthy |
| `avfvideosrc ! video/x-raw,width=640,height=360,framerate=15/1 ! fakesink` | **ERROR: not-negotiated (-4)** — the source cannot produce system-memory `video/x-raw` |
| `avfvideosrc ! video/x-raw,… ! videoconvert ! x264enc ! fakesink` (the devicesrc shape) | **ERROR: not-negotiated (-4)** |
| `avfvideosrc ! video/x-raw(memory:GLMemory),framerate=15/1 ! glcolorconvert ! gldownload ! videoconvert ! x264enc ! fakesink` | **45 frames in 3 s (~15 fps) — clean, full rate** |

So the isolated device-source shape (row 3) **hard-fails `not-negotiated`** when system memory is forced directly off the GLMemory-only source, and adding a GPU-side convert + explicit `gldownload` bridge (row 4) sustains full rate.

One gap I have **not** resolved, stated plainly rather than guessed at: the assembled capture flow in the field does not hard-fail negotiation the way this isolated repro does — instead it soft-stalls to ~1 fps with the `videoconvert` QoS warnings quoted at the top (no crash). Both symptoms point at the same missing GLMemory→system bridge in the capture front, but *why* the full assembled flow degrades softly instead of failing negotiation outright (a different convert element via `video_convert_mode()`/`configure_video_convert`, extra elements in the real graph, `autovideosrc` vs a directly-created `avfvideosrc`, etc.) is not established here and is left for whoever owns the fix. The isolated repro and the field logs are both included so that gap is visible, not hidden.

(For reference: the software x264 encoder itself is not the bottleneck — `videotestsrc ! video/x-raw,1280x720,25/1 ! videoconvert ! x264enc` runs 250 frames in ~1.4 s, ~180 fps, on the same machine.)

### Expected

A macOS built-in camera captures at its configured rate through the device source, the same as the seeded SRT test sources do.

### A working shape (evidence, not a prescription)

The fourth row above sustains full rate. Per **#682**'s stated rule ("the consuming block builds an input front that accepts both system memory and GPU memory") and the reference implementation cited there (`thumbnail_tap.rs:230-268`, which builds `glcolorscale → capsfilter(GLMemory) → gldownload …`), the device source's normalization front should accept GLMemory input and bridge it with `gldownload` rather than demanding system memory directly off the source. That keeps the Linux/NVIDIA and system-memory-camera paths working while fixing GLMemory-only macOS cameras.

### Secondary observations (separate, noted so they aren't lost)

- The camera negotiates its **portrait sensor mode 1080×1920** by default when resolution is left unset — worth a sane landscape default / explicit mode selection, but orthogonal to the stall.
- With `encoder_preference: auto` the VideoToolbox HW encoder (`vtenc_h264_hw`) is selected and instead throws `Failed to encode frame N: -17691` (VideoToolbox malfunction) and emits a malformed ~50 Mbps stream for a configured CBR-6000; forcing `software` avoids that but lands on the videoconvert stall above. The HW-encoder failure likely deserves its own investigation once capture is bridged.

### Environment

macOS 15 (Darwin 25.x), Apple Silicon, Homebrew GStreamer 1.28.x, Strom v0.6.8 (as installed by open-live-ingest v0.3.0). Reproduced 2026-09-17 with gst-launch pipelines above.

Owner: birme.

srperens:
Closes #837.

## The defect

`devicesrc` puts a capsfilter between the camera and its converter so the device negotiates the configured resolution and framerate at capture time. It names `video/x-raw` and nothing else, which means `memory:SystemMemory` — not "any". A camera that delivers GL memory and nothing else shares no format with it and cannot negotiate the capture front at all.

## The change

The pre-convert capsfilter now offers the same structure twice: plain `video/x-raw` first, then `video/x-raw(memory:GLMemory)`. System memory first, so a device that can deliver either keeps producing exactly what it does today.

No download is requested here. Where the camera picks GL memory, the block's system-memory output capsfilter refuses the link and #802's linker splices the `gldownload`, which is the consumer-side rule this repo settled on.

## What I measured, including what did not reproduce

Built-in camera on an Apple Silicon Mac, macOS 15, Homebrew GStreamer 1.28.6. The camera advertises GL memory by default:

```
avfvideosrc0.src: caps = video/x-raw(memory:GLMemory), 1080x1920, UYVY, 30/1, texture-target=rectangle
```

**The reported `not-negotiated` did not reproduce on this machine.** Asked for system memory, this `avfvideosrc` delivers it, at every mode tried:

| Mode | today's front | this branch |
|---|---|---|
| 640x360 @ 15 | ok | ok |
| 1280x720 @ 25 | ok | ok |
| 1920x1080 @ 30 | ok | ok |

So the hard failure in the report is specific to the reporter's camera or build — two Macs, two behaviours. What this branch changes here is which memory type gets picked: with both offered, the source prefers GL (its own preference wins over the order in the filter), so the flow becomes GL -> `gldownload` -> system memory. Cost, 125 buffers at a requested 25 fps, `fakesink`:

```
today:        5.53s  (22.6 fps)
this branch:  5.55s  (22.5 fps)
```

No measurable difference.

The soft ~1 fps stall the report also describes is **not** addressed here. It is a different failure — the process-wide converter choice in `gpu.rs`, which hands a software build a plain `videoconvert` that passes GL memory through undownloaded. That belongs to #682.

## Tests

`the_local_input_capture_front_accepts_gl_memory` in `backend/tests/gl_memory_link_test.rs` builds the block through `PipelineManager` and asserts its capture front intersects both GL and system-memory caps. It needs no camera and no GL context, so it runs in CI.

It fails when the fix is reverted:

```
test the_local_input_capture_front_accepts_gl_memory ... FAILED
the capture front demands system memory (video/x-raw, width=(int)1280,
height=(int)720, framerate=(fraction)25/1), so a GL-memory-only camera
cannot negotiate it and the capture stalls
```

**Ran locally on macOS**: `cargo test --test gl_memory_link_test` — 7 passed, 0 failed, and the revert check above. **Not run**: Linux and Windows (CI covers Linux); no GL-memory-only camera was available to confirm the end-to-end capture, since the block always constructs its own source and no test can inject a GL producer into it.


svensson00:
@srperens @birme — #842 is merged but the latest release is still v0.6.9 (2026-09-16, pre-merge), so no release or tagged image carries the devicesrc GL-negotiation fix yet. Could you cut a release or point me at a main-built image tag? I have the venue camera + open-live-ingest rig ready to retest Eyevinn/open-live#309 against it — that retest is the only step left to close or reopen it.

(Posted by the PM-team coordinator, authorized by @svensson00 in session 2026-09-19.)

PR #844, opened 2026-09-19
feat(audio): adaptive audio bridge for the conversation return

wagenet:
# Adaptive audio bridge between flows

Contributors hear each other over a conversation flow. On today's
`interaudiosink`/`interaudiosrc` path a backlog left by a network stall never drains —
`interaudiosrc` takes one period per clock tick — so it either stays for the rest of the
show or the sink flushes whole periods to get back under `buffer-time`, which is a skip
with a click at each edge. The only knob is buying the worst stall up front:
`buffer-time` 450, about 600 ms of conversation delay, for 200–500 ms stalls.

Audio Bridge Output → Audio Bridge Input runs at a low target latency and drains what a
stall leaves by time-scaling. The consumer contract is the guest-calling spec
(Eyevinn/open-live#323): one jitterbuffer per seat at its quality setting, stalls absorbed
downstream of it on the conversation path only. That spec also settles that conversation
audio never airs — so the program path does not use this — and that the conversation mix
comes from `liveaudiorouter`, which these blocks feed.

## Shape

A clock-paced reader and `scaletempo` in one bin. The reader owns the source pad: above
target it takes `rate` periods of audio per period and declares that rate in a segment,
and `scaletempo` plays it back in one period with pitch intact. A scaler fed at real time
from outside can only starve what is downstream — measured twice in earlier work, where a
2 % and a 5 % speed-up after the bridge left the backlog at ~400 ms.

Two decisions are load-bearing, both explained in `backend/src/gst/audio_bridge/`: the
reader never declares exactly 1.0, since `scaletempo` then flips to passthrough and
strands the ~50 ms it has queued (18 clicks in 10 s of toggling, 0 held at 1.0 + 1e-6);
and a small in-place element restamps the bin's output, since `scaletempo`'s own
timestamps jump by its latency at every rate change. It is an element and not a pad probe
because `scaletempo` emits empty buffers between strides, and a probe has no clean way to
swallow one: `BASE_TRANSFORM_FLOW_DROPPED` does.

The ring and the control law are media-agnostic, so a video sibling is re-plumbing. The
elements live in Strom rather than gst-plugins-rs: they must work on the GStreamer Strom
ships against, and the `scaletempo` handling is not proven on real links yet.

## Control law

Steered on the backlog *floor*, the lowest depth over the last 500 ms, not the
instantaneous depth, which saws with every producer burst. On raw depth the same simulated
run changes rate 837 times instead of 12.

| state | condition | action |
|---|---|---|
| Draining | floor > target + 20 ms | rate 1 + min(bound, bound × error/100 ms), 0.5 % steps |
| Stretching | floor < target − 20 ms | the same, below 1.0 |
| Underrun | less than one period left | play what there is with a 5 ms fade-out, then silence |
| Buffering | after an underrun | silence until the backlog is back at target, then fade in |
| Skip | floor > max latency | fade out, discard down to target, fade in |
| Start | first audio after longer than max latency idle | discard down to target |

The last row is a publisher connecting: its jitterbuffer hands over its whole latency at
once (~430 ms at jb 400), which nobody was listening to. A stall while playing is never
trimmed — that audio is kept and played late, then drained.

Underruns take the gap: a rate below 1.0 only helps while there is something to stretch.
Synthetic fill (NetEq-style) is real DSP outside the proven shape, and a permanently
higher target costs delay on every conversation.

Input properties: `channel`, `target_latency_ms` (40), `max_rate_change_percent` (10, 0
disables time-scaling), `max_latency_ms` (1000). The last three are live.

## Statistics

`GET /api/flows/{id}/rtp-stats` reports per Audio Bridge Input: backlog now, floor and
peak, rate, underruns with total and longest silence, audio drained and stretched, time
spent time-scaled, skips, and from the writer side a histogram of input gaps (≥50, ≥100,
≥200, ≥400 ms) with the longest. That histogram is the rehearsal signal `/webrtc-stats`
cannot give — RFC 3550 jitter read ~8 ms on a link with 500 ms stalls, `packets_lost` 0
under 5 % loss — and the tables below say what a target buys against it. Jitterbuffer
statistics are not re-added here; Eyevinn/strom#793 covers those.

## Evidence

**Element level, two real pipelines** (`backend/tests/audio_bridge_test.rs`): a producer
delivers a 437 Hz tone in 20 ms buffers, stalls 300 ms, then delivers what it held at once
— what a seat's jitterbuffer does after a stall. Asserted and passing: one underrun, the
held audio kept (peak backlog ≥ 250 ms) and drained (floor back under 60 ms at a 40 ms
target), one rate-1.0 segment downstream with contiguous timestamps, output within 150 ms
of the wall clock over 7 s, pitch 437 Hz ± 2 Hz while draining, no clicks. At
`max_rate_change_percent` 0 the backlog stays ≥ 200 ms, which is today's behaviour.

Each guard was checked by breaking what it guards: unity at exactly 1.0 puts clicks at the
underrun and the resume; no restamp element makes downstream see 23 segments; not releasing
the writer role makes a second program-flow start fail. Removing the anti-click fades does
**not** fail it — `scaletempo`'s overlap search lands the cut on a zero crossing for this
signal — so the fades rest on their unit test.

**Speech through the real elements**, 400 ms stall, 40 ms target, from the block's own
statistics:

| rate bound | peak backlog | back inside target + 20 ms | drained | backlog at the end |
|---|---|---|---|---|
| 5 % | 384 ms | 8.0 s after the stall | 360 ms | 40 ms |
| 10 % | 401 ms | 4.5 s after the stall | 370 ms | 40 ms |
| 0 % | 400 ms | never | 0 ms | 400 ms |

The listener gets a ~360 ms gap, faded both ends, then speech 5 % (or 10 %) fast at
unchanged pitch until the delay is gone. @wagenet listened to the recordings: the 5 % catch-up
passed unnoticed, 10 % was at most very slightly noticeable, and chose 10 % as the default
— this audio never airs, so half the recovery time is worth a little perceptibility. The
property text says where the edge is. The recordings are not in the
repo.

**WHIP rig**, two seats publishing Opus+H.264 into a program flow, impairment applied to
each publisher's SRTP; one conversation return per arm off the same network run, so every
arm sees the same impairment. `iaN` is today's `interaudiosink`/`interaudiosrc` at
`buffer-time` N; `brN` is this bridge at target N. **The rig arms below ran at a 5 % rate bound, before the
default became 10 %**: the rate bound moves only the "delay home" column, not dropout or
hole length, which follow the target. At 10 % that column roughly halves — the one
quiet-box trial taken at the new default brought a 300 ms stall home in 3.2 s at target 40
and 1.8 s at target 150, against 5.5 s and 3.3 s here, with dropout unchanged (1.45 % and
0.89 %). Delay is the return's tick
delay, dropout the holes in a continuous tone. Median [min–max] over valid trials; a trial
counts only when the box was quiet enough (program control and RFC jitter in range).

Clean link (4 trials). Same dropout everywhere, so this is purely what each costs in delay:

| arm | delay ms | dropout % |
|---|---|---|
| program | 510 [507–520] | 0.11 |
| ia450 | 637 [630–652] | 0.10 |
| ia120 | 311 [299–316] | 0.10 |
| **br40** | **206 [200–217]** | 0.11 |
| br150 | 318 [316–323] | 0.11 |
| br300 | 468 [463–476] | 0.11 |

One link stall, jb 400. "delay peak" is the return's delay just after the stall, "home" how
long until it is back at its own baseline:

| stall | arm | dropout % | longest hole ms | delay ms | delay peak ms | home s |
|---|---|---|---|---|---|---|
| 150 ms | ia450 | 0.09 | 0 | 628 | 628 | — |
| | br40 | 0.69 | 125 | 230 | 320 | 1.8 |
| | br150 | 0.11 | 0 | 337 | 363 | 1.8 |
| 300 ms | ia450 | 0.10 | 0 | 627 | 627 | — |
| | br40 | 1.50 | 280 | 223 | 468 | 5.5 |
| | br150 | 0.96 | 175 | 334 | 464 | 3.3 |
| | br300 | 0.12 | 0 | 480 | 515 | 2.0 |
| 600 ms | program | 5.94 | 490 | 507 | 527 | 2.0 |
| | ia450 | 1.65 | 200 | 621 | 640 | — |
| | br40 | 5.01 | 595 | 236 | 739 | 11.4 |
| | br300 | 2.75 | 330 | 485 | 746 | 6.5 |

A stall shorter than the target passes with no hole and the delay is back within ~2 s. A
longer one leaves a hole of stall minus target, and the delay walks home at the rate bound
(5 %: ~11 s for 600 ms). A 600 ms stall is past what the program path absorbs at all —
5.94 % there — so no setting is good; `ia450` is the mildest and pays 400 ms of delay all
show for it.

Impaired links, where the interesting comparison is at equal delay. Gilbert–Elliott burst
loss, ~13 % (3 trials), and a wifi-like link — 120 ms jitter, 200–400 ms stalls every 3–6 s,
0.4 % loss (4 of 7 trials quiet enough):

| link | arm | dropout % | holes ≥100 ms | delay ms |
|---|---|---|---|---|
| burst loss | program | 5.03 | 2 | 493 |
| | ia120 | 24.13 | 16 | 327 |
| | ia450 | 5.11 | 2 | 581 |
| | br150 | 8.45 | 9 | 547 |
| | br300 | 5.05 | 3 | 570 |
| wifi-like | program | 5.04 | 2 | — |
| | ia120 | 11.50 | 8 | — |
| | ia450 | 1.31 | 1 | 716 |
| | br150 | 4.89 | 6 | — |
| | br300 | 2.67 | 2 | 700 |

A fixed small buffer is the worst of both on a bad link — `ia120` takes 24 % where the
bridge at a comparable setting takes 8.5 % — because the bridge lets the backlog grow and
drains it later where `interaudiosrc` flushes. It also means the bridge gives up its low
delay on a bad link by design: `br150` measured 547 ms there, not the 318 ms it holds on a
clean one. Continuity still favours the big fixed buffer (`ia450`), at 400 ms of delay
every second of the show.

**Not measured:** whether the catch-up is audible on speech (recordings exist, judgement
for a listener), contributor video, and anything about the bed.

## Tests

`cargo test` (40 binaries, 858 passed, 0 failed; the 2 ignored are pre-existing opt-ins),
`cargo clippy --all-targets -- -D warnings`, `cargo fmt`. `openapi_test` passes unchanged —
the statistics travel through the existing generic `Statistic` type. The new tests need
`scaletempo` from `gst-plugins-good`, already in the CI package list, and fail rather than
skip without it.

Opened by Claude on behalf of @wagenet.


srperens:

